PR #41, opened 2021-09-02
Remove MavenRecipeTest in favor of importing from rewrite-maven

aegershman:
I don't think this is used anymore? PR'ing the removal of `MavenRecipeTest` to be sure 👍


tkvangorder (approved):
LGTM

PR #46, opened 2021-09-10
Add NoGuavaCreateTempDir recipe for RSPEC-4738

pway99:


src/main/java/org/openrewrite/java/migrate/guava/NoGuavaCreateTempDir.java
```diff
@@ -0,0 +1,140 @@
...
+import org.openrewrite.java.search.UsesType;
+import org.openrewrite.java.tree.*;
+
+import java.util.Collections;
+import java.util.Set;
+
+public class NoGuavaCreateTempDir extends Recipe {
+    @Override
+    public String getDisplayName() {
+        return "Use java.io.Files.createTempDirectory instead of Guava";
```
jkschneider, line 34: Please use ticks to surround the method pattern. No need for it to be fully qualified. So `Files#createTempDirectory()`

src/main/java/org/openrewrite/java/migrate/guava/NoGuavaCreateTempDir.java
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    @Override
+    public String getDescription() {
+        return "Replaces Guava `com.google.common.io.Files.createTempDir()` with `java.io.Files.createTempDirectory(..)`.  Note, java.io.Files.createTempDirectory() throws an IOException, if necessary a private method will be added to the class for invoking java.io.Files.createTempDirectory() and rethrowing the IOException as a new RuntimeException.";
```
jkschneider, line 39: Anywhere you refer to a type, surround with ``

src/test/kotlin/org/openrewrite/java/migrate/guava/NoGuavaCreateTempDirTest.kt
```diff
@@ -0,0 +1,235 @@
...
+        """,
+        after = """
+            import java.io.File;
+            import java.io.IOException;
+            import java.nio.file.Files;
+            
+            class A {
+                File gDir = createTempDirectory();
+            
+                private static File createTempDirectory() {
```
jkschneider, line 80: I really don't like generating boilerplate methods like this in recipes. Maybe we limit the places where we apply the recipe to scopes that already return or otherwise catch an `IOException`.

jkschneider (changes requested):

PR #78, opened 2022-04-01
Add UpgradeWro4jMavenPluginVersion

### LINKED ISSUE #75: Provide recipe to upgrade `wro4j` to a version that is compatible with Java 11
Need to look at the release notes for `wro4j` to determine which version should be used with Java 11.

See: 

https://stackoverflow.com/a/62362986
https://blog.jdriven.com/2022/03/major-migrations-made-easy-with-openrewrite/

timtebeek:
Closes #75


timtebeek:
Wasn't sure about:
- naming of the recipe
- if we should also include gradle, or future proof naming/description
- whether to include a test

While I'm not sure how much Wro4j is still used out in the wild, this should at the very least make it easier to migrate the Spring PetClinic 1.5.x as an example.

jkschneider:
> if we should also include gradle

Gradle support is going to keep improving so we can come back to that.

> or future proof naming/description

I think it's ok. The recipe could switch over multiple versions if that time ever comes without changing the name.

Issue #66, opened 2022-03-27
Override the Maven property java.version for projects using spring-boot-starter-parent


timtebeek:
As you're probably well aware, Maven projects using `spring-boot-starter-parent` can override the `java.version` property to set the `maven.compiler.source` and `maven.compiler.targer` versions. These are [defined in the parent pom.xml through their build.gradle file](https://github.com/spring-projects/spring-boot/blob/2.6.x/spring-boot-project/spring-boot-starters/spring-boot-starter-parent/build.gradle#L22).

It would make sense (to me at least) to define this property when needed, and override the value for such Spring Boot projects when moving from Java 8 to 11. That way users have one less manual step to perform when migrating in bulk from 8 to 11.

I imagine the implementation would not need much Spring specifics:
- you'll want to find spring-boot-starter-parent, anywhere up the hierarchy,
- detect the presence of the property value, also up the hierarchy,
- if the nearest value is below 11, define the property with the new value.

timtebeek:
A naive approach wouldn't even need to know about the `spring-boot-starter-parent` and can just override the `java.version` property through `org.openrewrite.maven.ChangePropertyValue`, not adding it if not present already.

tkvangorder:
@timtebeek , thanks for the suggestion. We initially decided against setting this property because those risk-averse enterprises might choose to first upgrade their dependencies (in preparation for the move to 11) but still want to continue to compile with Java 8.

I think now with the Java releases coming fast and furious, it would be a good time to do this by default (and it helps that we now have a maven model that would allow us to do this accurately)

tkvangorder:
We might need to write an imperative recipe for this, so we can walk up the parent hierarchy and place the property (if it does not exist) at the root project.

timtebeek:
This was briefly discussed over Slack; here's a naive approach that merely bumps the property version if it's already defined: https://github.com/openrewrite/rewrite-migrate-java/compare/main...timtebeek:SpringBootJavaVersion11 Could be an intermediate step while we wait for an imperative recipe. Unlikely to hurt if the property is already there.

PR #82, opened 2022-04-04
Change Maven property java.version to 11

timtebeek:
Partial fix for #66


timtebeek:
@tkvangorder opened a PR as suggested! :)

src/test/kotlin/org/openrewrite/java/migrate/SpringBootJavaVersion11Test.kt
```diff
@@ -0,0 +1,59 @@
...
+
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+import org.openrewrite.Recipe
+import org.openrewrite.config.Environment
+import org.openrewrite.maven.MavenRecipeTest
+
+class SpringBootJavaVersion11Test : MavenRecipeTest {
+    override val recipe: Recipe = Environment.builder()
+        .scanRuntimeClasspath("org.openrewrite.java.migrate.spring")
```
timtebeek, line 26:
```suggestion
        .scanRuntimeClasspath("org.openrewrite.java.migrate.spring")
```

src/main/resources/META-INF/rewrite/java.version-11.yml
```diff
@@ -0,0 +1,25 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+---
+type: specs.openrewrite.org/v1beta/recipe
+name: org.openrewrite.java.migrate.JavaVersion11
+displayName: Change Maven property java.version to 11
+description: Change the Maven compiler source and target versions to 11, via Spring Boot property java.version.
```
timtebeek, line 20: Was doubting to fully remove any mention of Spring here, as I'm not sure if the property is used as a convention in any other framework. So without the mention of Spring, we would also have to remove the maven compile source & target reference, as there's not guarantee it does anything at all if Spring is not present.

timtebeek: Either way, welcome to change this as you see fit; I'll be asleep for the next few hours :)

tkvangorder: 🤔 Yeah, you are right, I guess this property is more a "spring" thing.

I think it might be safer to add all three property changes (and only make the change if the property already exists) 

- `java.version`
- `maven.compiler.source`
- `maven.compiler.target`

That will take care of the three most common use cases. 

Then beyond this, Once we have added plugin/pluginManagement to the semantic model for maven, this can be converted to an imperative recipe, so we can make decisions like "if the property is not defined, add it to the top-level pom"

timtebeek: Pushed a change to now also change the values of the maven compiler properties if present.

I'd be inclined to rename the recipe, displayName & description to reflect that we are changing the maven compiler source & target version, rather than any Spring/Java specific property. That's essentially what we aim to do here, with the Spring specifics just being an implementation detail for some projects.

Welcome to weigh in or make any changes to wrap this up. :)

timtebeek:
Ok all done; sadly just too late for the patch release, but still good to see this make it in. Would drop another manual step from my blogpost. :)

tkvangorder:
Good news, since I am the one doing the point release....I will just release this project again.

PR #89, opened 2022-04-11
Added UseEnumSetOf recipe.

### LINKED ISSUE #88: RSPEC-1641: Sets with elements that are enum values should be replaced with EnumSet
> When all the elements in a Set are values from the same enum, the Set can be replaced with an EnumSet, which can be much more efficient than other sets because the underlying data structure is a simple bitmap.

Noncompliant Code Example

```java

public class MyClass {

  public enum COLOR {
    RED, GREEN, BLUE, ORANGE;
  }

  public void doSomething() {
    Set<COLOR> warm = new HashSet<COLOR>();
    warm.add(COLOR.RED);
    warm.add(COLOR.ORANGE);
  }

public void doSomething2() {
    Set<COLOR> warm = Set.of(COLOR.RED,COLOR.ORANGE);
}
```

Compliant Solution

```java
public class MyClass {

  public enum COLOR {
    RED, GREEN, BLUE, ORANGE;
  }

  public void doSomething() {
    Set<COLOR> warm = EnumSet.of(COLOR.RED, COLOR.ORANGE);
  }
}
```

See : https://rules.sonarsource.com/java/RSPEC-1641

traceyyoshima:
Changes:

- Added `UseEnumSetOf` recipe.
- Added declarative recipe `java-util-apis`.
- Added `java-util-apis` recipe to java 8 to 11.
- Moved `util` based recipes from `lang` to new package `util`.

fixes #88

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/util/MigrateCollectionsSingletonList.java
+++ b/src/main/java/org/openrewrite/java/migrate/util/MigrateCollectionsSingletonList.java
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.openrewrite.java.migrate.lang;
+package org.openrewrite.java.migrate.util;
 
 import org.openrewrite.ExecutionContext;
 import org.openrewrite.Recipe;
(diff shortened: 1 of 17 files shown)


tkvangorder (approved):
LGTM

PR #98, opened 2022-04-25
Remove Cobertura Maven Plugin, as it is not compatible with Java 11.

timtebeek:
While JaCoCo is a likely replacement, actually plugging that in would
likely also require integration with whatever the build was using to
visualize the results. It is unlikely there is a direct replacement
option there.

Removing Cobertura at least ensures the build is not left in a broken
state, even though Coverage information may then be missing. Any
coverage visualization is likely to surface that for users.


tkvangorder (approved):
This sounds reasonable to me, thanks for the contribution!

Issue #77, opened 2022-04-01
Java 15+ recipe to adopt non-static String formatted(Object... args)


timtebeek:
As of [Java 15 JEP 378: Text Blocks](https://openjdk.java.net/jeps/378) there's a new [`String formatted(Object... args)`](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/String.html#formatted(java.lang.Object...)) that's equivalent to `String.format(this, args)`. It reads quite nicely even when not used with text blocks.

I have no idea if you're already taking in suggestions beyond Java 11, but figured this could be a nice addition.

tkvangorder:
Support for later versions of Java is high on our priority list. This is a good suggestion.

timtebeek:
Perfect; cross referencing [Java 17 parser support](https://github.com/openrewrite/rewrite/issues/1501) here. Perhaps a new label could help group the issues here, as I imagine there's further quality of life suggestions that aren't actionable yet, such as text blocks, instance of test & assignment, and switch expressions.

Jonpez2:
I’d be really interested to learn more about your plans for later versions of Java. Is there anywhere I could read about it please?

Thank you!

PR #119, opened 2022-08-21
Add org.openrewrite.java.migrate.lang.StringFormatted to Java 17 recipes

timtebeek:
Fixes https://github.com/openrewrite/rewrite-migrate-java/issues/77

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,7 @@ build/
 .idea/
 .boot-releases
 out/
+.classpath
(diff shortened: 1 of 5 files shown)


timtebeek:
Some minor improvements that might be worthwhile still:
- we likely need not wrap method invocations as the first argument either
- should we set the type on the newly returned expression?

sambsnyd (approved):

PR #121, opened 2022-08-23
No need to wrap String.formatted for local variable or field

timtebeek:
Sorry for the drip feed pull requests lately; you always uncover more cases when you apply them in bulk. :)


jkschneider:
This wrapper thing feels like something we could promote into JavaTemplate.

timtebeek:
That does indeed sound like something that could be handled generically.
Calling a method on the result of an expression should wrap that expression if needed.

For now I'd thought to wrap defensively here, but would be great if that could be removed in the future.
Ideally the tests here will just break once it's added to rewrite-core, signalling it's time for removal here.

PR #124, opened 2022-09-04
Replace lombok.val with final var on projects using Java 11 or higher

timtebeek:
Figured this would be a fun one, and learning exercise. It replaces `lombok.val` with `final var`, as there's not a lot of value in retaining `lombok.val` nowadays. As a bonus it makes projects more compatible with OpenRewrite, at least until there's integrated support.

Only thing that I'm not yet sure how to resolve is the tests containing infix whitespace & comment. Some guidance there would be appreciated once again. My attempts at using `maybeAutoFormat` failed up to now.

Fixes https://github.com/projectlombok/lombok/issues/3207 
Fixes https://github.com/openrewrite/rewrite/issues/1297

TODO
- [ ] Reference from META-INF yaml
- [ ] Either fix or ignore infix whitespace & comments

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/lombok/LombokValToFinalVar.java
+++ b/src/main/java/org/openrewrite/java/migrate/lombok/LombokValToFinalVar.java
@@ -1,0 +1,83 @@
+/*
+ * Copyright 2022 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.java.migrate.lombok;
+
+import org.openrewrite.Applicability;
+import org.openrewrite.ExecutionContext;
+import org.openrewrite.Recipe;
+import org.openrewrite.TreeVisitor;
+import org.openrewrite.java.JavaIsoVisitor;
+import org.openrewrite.java.JavaParser;
+import org.openrewrite.java.JavaTemplate;
+import org.openrewrite.java.search.UsesJavaVersion;
+import org.openrewrite.java.search.UsesType;
+import org.openrewrite.java.tree.J;
+
+import java.time.Duration;
+
+public class LombokValToFinalVar extends Recipe {
+
+    private static final String LOMBOK_VAL = "lombok.val";
+
+    @Override
+    public String getDisplayName() {
+        return "Replace `lombok.val` with `final var`";
+    }
+
+    @Override
+    public String getDescription() {
+        return "Replace `lombok.val` with `final var` on projects using Java 11 or higher.";
+    }
+
+    @Override
+    protected TreeVisitor<?, ExecutionContext> getSingleSourceApplicableTest() {
+        return Applicability.and(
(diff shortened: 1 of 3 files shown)


timtebeek:
Wasn't entirely sure whether this recipe would fit in best here or in openrewrite/rewrite. My guess was here as it's most applicable for users upgrading between Java versions.

src/test/kotlin/org/openrewrite/java/migrate/lombok/LombokValToFinalVarTest.kt
```diff
@@ -0,0 +1,146 @@
...
+                        "foo";
+                    }
+                }
+                """,
+                """
+                class A {
+                    void bar() {
+                        final var foo = 
+                        // Infix
+                        "foo";
```
timtebeek, line 94: One could argue that both `retainInfixComment` & `retainWhitespace` should be rare, and therefore not worth supporting. That would be fine by me, but thought to make that decision explicit here, to make clear it's not an oversight if we drop it (for) now.

pway99:
Thanks Tim!

The recipe should also account for the case where `lombok.val` types are used in for-each statements  
```java
List<String> lst = new ArrayList<>();
for (val s : lst) {}
```

If you would like, I've got your PR loaded up and am happy to help with these changes :)

src/test/kotlin/org/openrewrite/java/migrate/lombok/LombokValToFinalVarTest.kt
pway99: My thought is that even though `NamedVariable` comments are rare they exist for a reason and they should be accounted for.  In this case the prefix from the original variable can just be copied to the transformed variable.

check this out:
```java
J.VariableDeclarations.NamedVariable nv = mv.getVariables().get(0);
return varDecls.withVariables(ListUtils.map(varDecls.getVariables(),
    varDecl -> varDecl.withInitializer(varDecl.getInitializer().withPrefix(nv.getInitializer().getPrefix()))));
```

src/main/java/org/openrewrite/java/migrate/lombok/LombokValToFinalVar.java
```diff
@@ -0,0 +1,83 @@
...
+        return new LombokValToFinalVarVisitor();
+    }
+
+    private static class LombokValToFinalVarVisitor extends JavaIsoVisitor<ExecutionContext> {
+
+        private static final String FINAL_VAR_ANY = "final var #{} = #{any()};";
+
+        @Override
+        public J.VariableDeclarations visitVariableDeclarations(J.VariableDeclarations mv, ExecutionContext p) {
+            if (mv.getTypeAsFullyQualified().isAssignableTo(LOMBOK_VAL)) {
```
pway99, line 63: TypeUtils is null-safe and would be a better option for checking type
```suggestion
            if (TypeUtils.isOfClassType(mv.getType(), LOMBOK_VAL)) {
```

timtebeek:
> If you would like, I've got your PR loaded up and am happy to help with these changes :)

Yes please! Just arrived in Stavanger, so won't get to it myself otherwise for another week. Feel free to change it however you feel is best!

pway99:
Well, It turns out the changes were not that simple after all...  I discovered a `Javatemplate  BlockStamentGenerator` bug which was preventing for-each loop transformations.  I'll add some changes to your PR and merge them in after [PR](https://github.com/openrewrite/rewrite/pull/2200) is merged.

pway99:
Hi Tim, I've managed to get this merged in and will test it on the [SaaS](https://public.moderne.io) if everything looks good I will add a reference to the appropriate `META-INF yaml` file.  Thanks again for your help on this one :)

timtebeek:
Thanks for the fixes & merge! The bug in JavaTemplateBlockStatementGenerator would have surely tripped me up, so I'm glad you were able to dive into this one. Curious to hear what effects you see on SaaS! :)

PR #127, opened 2022-10-12
JDeprScan release 17

timtebeek:
Figured for Java 17 we'd want bump the release to Java 17.
Would `AddPlugin` also change the configuration if already present?
Noticed there's no tests for this recipe yet; could be added if needed.


tkvangorder (approved):
LGTM

PR #128, opened 2022-10-12
Add recipe to upgrade to Java 17, which also upgrades from 8 to 11 first

timtebeek:
Based on how Spring Boot upgrades cascade down to earlier upgrades first. Expect the list of recipes to grow, which poorly fit into the JavaVersion17 recipe, which (now) only sets compiler source & target version again.


timtebeek:
The application here is to easily migrate an application to Java 17 regardless of the current Java version, without having to explicitly migrate to intermediate versions. Bonus that it shortens my demo. ;)

pway99 (approved):
Thanks @timtebeek!

PR #130, opened 2022-10-12
Replace !Optional.isPresent() with Optional.isEmpty() on Java 11+

timtebeek:
Really quick one, so might be missing some edge cases, but seemed simple enough, and improvement on Java 11 and above.


timtebeek:
While it's unlikely to be used much, we might want to add the inverse here as well; `!Optional.isEmpty() -> Optional.isPresent()`. What are your thoughts on that?

pway99:
> while it's unlikely to be used much, we might want to add the inverse here as well; !Optional.isEmpty() -> Optional.isPresent(). What are your thoughts on that?

I agree its probably not that common, although I would not be surprised if it's not all that uncommon ;)

pway99 (approved):
Looks good @timtebeek.
going to merge this in; adding support for transforming `!Optional.isEmpty()' to `Optional.isPresent()` can be done with a follow up pr.

PR #131, opened 2022-10-13
Verify String.formatted() works on text blocks in Java 17

timtebeek:
Had some issues escaping with Kotlin; not sure if this even works yet, but clearest with an example.
Looking forward to that Kotlin parser to convert the tests back to Java. :)


pway99:
Thanks @timtebeek

timtebeek:
You're welcome! Those last odd cases cover edge cases I saw fail in Wicket, but have not been able to reproduce here. Was part of migrating Wicket, but that had some parser issues as well.

PR #144, opened 2022-11-04
XML Bind and similar deps should look for ..* pattern 

BoykoAlex:
FQNs starting with `jakarta.xml.bind` should be accepted rather than jakarta.xml.bind<any char>. Looks like the pattern to be used in `UsesType` is incorrectly specified as FQNs such as: `javax.xml.bind.annotation.XmlElement` don't result in adding jakarta XML Bind dependency.


tkvangorder (approved):
TY!

PR #152, opened 2022-12-09
Java 11 recipe no longer adds JDeprScan plugin

nmck257:
By my estimation, the Java 11 recipe has outgrown this. A client could plausibly run this recipe and successfully migrate (replacing most relevant deprecations), and at that point, the additional plugin is just noise, which they'll likely want to remove later.

And for a user who does want to use jdeprscan, the syntax even without an explicitly-configured plugin is pretty concise:
```shell
mvn compile jdeprscan:jdeprscan -Drelease=11
```


timtebeek:
This also "resolves" the issue that the release property is not upgraded when going to Java 17.

PR #161, opened 2023-01-05
Recipe for migration JakartaEE Security is wrong

### LINKED ISSUE #160: Recipe for migration JakartaEE Security is wrong
Packages in our codebase were changed from `javax.security.*` to `jakarta.security.*` even though any dependency of `javax.security.enterprise` is included in the project.

The security packages `javax.security.*` are included in the JDK itself, and for JDK 11, it still contains the prefix `javax.*`, so probably these packages should not be updated in the codebase.

If we specify the required packages more, we'll avoid unnecessary collisions.

Issue to OpenRewrite docs: https://github.com/openrewrite/rewrite-docs/issues/111

See: https://github.com/quarkusio/quarkus/issues/30198
https://github.com/keycloak/keycloak/issues/16247

mabartos:
Fixes #160

For more info, see the attached issue.

### DIFF
--- a/src/main/resources/META-INF/rewrite/jakarta-ee-9.yml
+++ b/src/main/resources/META-INF/rewrite/jakarta-ee-9.yml
@@ -572,21 +572,21 @@ recipeList:
 ---
 type: specs.openrewrite.org/v1beta/recipe
 name: org.openrewrite.java.migrate.jakarta.JavaxSecurityToJakartaSecurity
-displayName: Migrate deprecated `javax.security` packages to `jakarta.security`
+displayName: Migrate deprecated `javax.security.enterprise` packages to `jakarta.security.enterprise`
 description: Java EE has been rebranded to Jakarta EE, necessitating a package relocation.
 recipeList:
   - org.openrewrite.maven.AddDependency:
       groupId: jakarta.security.enterprise
       artifactId: jakarta.security.enterprise-api
       version: 2.x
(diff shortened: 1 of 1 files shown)


src/main/resources/META-INF/rewrite/jakarta-ee-9.yml
```diff
@@ -572,21 +572,21 @@ recipeList:
...
-displayName: Migrate deprecated `javax.security` packages to `jakarta.security`
+displayName: Migrate deprecated `javax.security.enterprise` packages to `jakarta.security.enterprise`
 description: Java EE has been rebranded to Jakarta EE, necessitating a package relocation.
 recipeList:
   - org.openrewrite.maven.AddDependency:
       groupId: jakarta.security.enterprise
       artifactId: jakarta.security.enterprise-api
       version: 2.x
-      onlyIfUsing: javax.security.*
+      onlyIfUsing: javax.security.enterprise.+
```
mabartos, line 582: I'm not sure if use here '+' or '*'.

jkschneider: It's a '*', will change that and merge.

mabartos:
@jkschneider Thanks!

jkschneider:
Thank you @mabartos. Also thanks for pointing out the `.+` usage as there were several others written that way too.

PR #174, opened 2023-01-24
Bump GitHub Actions Java version

timtebeek:
Following on from https://github.com/openrewrite/rewrite-github-actions/pull/24

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -96,7 +96,7 @@ dependencies {
     implementation(platform("org.openrewrite:rewrite-bom:${rewriteVersion}"))
     implementation("org.openrewrite:rewrite-java")
     implementation("org.openrewrite:rewrite-maven")
-    implementation("org.openrewrite:rewrite-github-actions")
+    implementation("org.openrewrite.recipe:rewrite-github-actions:$rewriteVersion")
 
     runtimeOnly("org.openrewrite:rewrite-java-8")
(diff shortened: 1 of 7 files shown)


build.gradle.kts
```diff
@@ -96,6 +96,8 @@ dependencies {
     implementation(platform("org.openrewrite:rewrite-bom:${rewriteVersion}"))
     implementation("org.openrewrite:rewrite-java")
     implementation("org.openrewrite:rewrite-maven")
+    implementation("org.openrewrite:rewrite-github-actions")
```
sambsnyd, line 99:
```suggestion
    implementation("org.openrewrite.recipe:rewrite-github-actions:$rewriteVersion")
```

github-actions dose not have its version managed by rewrite-bom (that only covers stuff from openrewrite/rewrite) and its groupId is "org.openrewrite.recipe"

sambsnyd (changes requested):

build.gradle.kts
timtebeek: I was just looking at https://github.com/openrewrite/rewrite-recipe-bom/blob/main/build.gradle.kts#L36 indeed

sambsnyd:
It is not immediately obvious to me what is wrong, but there are some test failures complaining that recipe cannot be found on the classpath: https://ge.openrewrite.org/s/35nsrbt6uex5c/tests/overview?outcome=failed

PR #179, opened 2023-01-31
Add `InstanceOfPatternMatch` recipe to `java-version-17.yml`

knutwannheden:
Related: openrewrite/rewrite#2690

### DIFF
--- a/src/main/resources/META-INF/rewrite/java-version-17.yml
+++ b/src/main/resources/META-INF/rewrite/java-version-17.yml
@@ -45,3 +45,4 @@ recipeList:
       version: 17
   - org.openrewrite.java.migrate.maven.UseMavenCompilerPluginReleaseConfiguration:
       releaseVersion: 17
+  - org.openrewrite.java.InstanceOfPatternMatch


knutwannheden:
@timtebeek Is this the right place to add the recipe?

knutwannheden:
@timtebeek Also, I don't know if this can be integrated before a new release has been made. Is the CI using the released version or SNAPSHOT builds?

src/main/resources/META-INF/rewrite/java-version-17.yml
```diff
@@ -32,6 +32,7 @@ recipeList:
   - org.openrewrite.java.migrate.lang.StringFormatted
   - org.openrewrite.java.migrate.lombok.UpdateLombokToJava17
   - org.openrewrite.github.SetupJavaUpgradeJavaVersion
+  - org.openrewrite.java.InstanceOfPatternMatch
```
knutwannheden, line 35: @timtebeek Thanks. Somehow I grabbed the wrong recipe.

timtebeek: No worries there; it's easily overlooked; as a side effect we tested the CI and integration prior to a release of openrewrite/rewrite, so we're all clear here.

Would you want to [wait for a response on the potential cleanup package move](https://github.com/openrewrite/rewrite/pull/2690#issuecomment-1410362831) before we merge?

knutwannheden: Yes. Let's wait for that. Also, I am unsure if the new recipe is already available in the releases we use in the CI builds.

timtebeek:
I'm not sure of your intentions with the `System.out.println` in https://github.com/openrewrite/rewrite-migrate-java/pull/179/commits/507f67dd928a54888f73de5f07ce07bf1b887690;
but I've approved the run just this once; if that's what hacks us well done. ;)

knutwannheden:
> I'm not sure of your intentions with the `System.out.println` in [507f67d](https://github.com/openrewrite/rewrite-migrate-java/commit/507f67dd928a54888f73de5f07ce07bf1b887690);
> but I've approved the run just this once; if that's what hacks us well done. ;)

I just added a test to make sure that the new recipe is also being used. If the `System.out.println()` call in the test (which of course won't get executed) bothers you, we can of course change it 😄

knutwannheden:
Looks like the recipe wasn't applied properly. Makes me wonder why the test didn't fail before I modified it. Shouldn't it complain if it can't find the recipe? I could debug that.

timtebeek:
It should; surprised it didn't. Thanks for troubleshooting either way; it's good to uncover these cross-repository issues this way. Likely wouldn't have registered if it was indeed in the same repository.

knutwannheden:
> It should; surprised it didn't. Thanks for troubleshooting either way; it's good to uncover these cross-repository issues this way. Likely wouldn't have registered if it was indeed in the same repository.

Actually, my test is wrong. I will fix it.

timtebeek (approved):

src/main/resources/META-INF/rewrite/java-version-17.yml
timtebeek: Great; approved already; will monitor and merge, unless someone else beats me to it. Thanks again!

knutwannheden:
> Actually, my test is wrong. I will fix it.

The problem with the test was that the `instanceof` and type cast expressions were method invocations, which we can't know if they have side effects. So theoretically it could be important that the method gets called twice. That is why the recipe doesn't get applied and I had to extract the expression into a variable first. I had myself forgotten about this safety guard.

I will add an issue for support of [IntelliJ's `@Contract` annotation](https://www.jetbrains.com/help/idea/contract-annotations.html), which could be used for more precise side-effects calculations. Why `Map#get(Object)` isn't annotated as `@Contract(pure=true)` (see https://github.com/JetBrains/intellij-community/blob/master/java/jdkAnnotations/java/util/annotations.xml#L3249-L3254) I am unsure about, so that wouldn't have made a difference in my example.

PR #200, opened 2023-03-24
Update algorithm of `UseTextBlocks` to be more reliable

kunli2:
Update the algorithm of `UseTextBlocks` to be more reliable.
1. Fix bug found on SAAS. added unit test for it. 
2. Update the algorithm to adopt all the cases.

### DIFF
--- a/src/test/java/org/openrewrite/java/migrate/lang/UseTextBlocksTest.java
+++ b/src/test/java/org/openrewrite/java/migrate/lang/UseTextBlocksTest.java
@@ -48,7 +48,64 @@ class UseTextBlocksTest implements RewriteTest {
                     String query = \"""
                             SELECT * FROM
                             my_table
-                            WHERE something = 1;\""";
+                            WHERE something = 1;\\
+                            \""";
+                }
+                """
+            ),
+            17
+          )
+        );
+    }
+
+    @Test
+    void preserveTrailingWhiteSpaces() {
+        rewriteRun(
+          //language=java
+          version(
+            java(
+              """
+                class Test {
+                        String query = "SELECT * FROM    \\n" +
+                            "my_table    \\n" +
+                            "WHERE something = 1; ";
+                }
+                """,
+              """
+                class Test {
+                        String query = \"""
+                            SELECT * FROM   \\s
+                            my_table   \\s
+                            WHERE something = 1; \\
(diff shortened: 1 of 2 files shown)


src/test/java/org/openrewrite/java/migrate/lang/UseTextBlocksTest.java
```diff
@@ -393,4 +392,78 @@ public static void main(String[] args) {
...
+                        log(\"""
+                           \s
+                            =========================================================
+                                                                                    \\s
+                                      Welcome to Spring Integration!                \\s
+                                                                                    \\s
+                                For more information please visit:                  \\s
+                                https://www.springsource.org/spring-integration     \\s
+                                                                                    \\s
+                            =========================================================\\
```
knutwannheden, line 459: @kunli2 Just out of interest: What is this last `\\` used for?

kunli2: Yeah, good question, it's to ensure that the closing delimiter is in a new line to manage indentation.
And this should be the best practice via https://docs.oracle.com/en/java/javase/20/text-blocks/index.html

src/test/java/org/openrewrite/java/migrate/lang/UseTextBlocksTest.java
```diff
@@ -393,4 +392,78 @@ public static void main(String[] args) {
...
+                            + "\\n=========================================================");
+                    }
+                    void log(String s) {}
+                }
+                """,
+              """
+                class A {
+                    void welcome() {
+                        log(\"""
+                           \s
```
knutwannheden, line 451: This line looks wrong to me. This will print an `s`. Also note that in the result after the recipe every line has a leading space, whereas in the input there is no leading space.

kunli2: I think this is correct, `\s` here is used to explicitly make the expected result string preserve leading white spaces.

kunli2: the input has not a leading space, but a leading new line, and that new line is converted to indentation (those white spaces)  + newline.

PR #201, opened 2023-03-25
Optimize `StringFormatted` for performance

knutwannheden:
The `StringFormatted` recipe appears to have very bad performance. This one is actually quite easy to implement "manually" without the convenience of `JavaTemplate`, as it basically only shuffles the arguments around.

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/lang/StringFormatted.java
+++ b/src/main/java/org/openrewrite/java/migrate/lang/StringFormatted.java
@@ -15,20 +15,17 @@
  */
 package org.openrewrite.java.migrate.lang;
 
-import org.openrewrite.Applicability;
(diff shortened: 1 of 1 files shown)


knutwannheden:
@sambsnyd @timtebeek Maybe you can also check and see if you agree this recipe has bad performance. Also, it remains to be seen if this fix actually helps (possibly it is the `maybeAutoFormat()` which is the baddy).

timtebeek:
Seems reasonable to me; thanks for taking the time to revise this one. Can't say much as to the performance aspect; I very much agree with you that it's a likely improvement. Suggest we merge this as it is, and look at the daily recipe runs if the data tables show any change in performance there.

timtebeek (approved):

Issue #197, opened 2023-03-23
Convert Guava Optional to JDK


timtebeek:
We already have quite some recipes to convert Guava classes to standard JDK equivalents:
https://github.com/openrewrite/rewrite-migrate-java/blob/main/src/main/resources/META-INF/rewrite/no-guava.yml
Seems like Guava's Optional could be a nice addition to the list of supported migrations.

## Guava
https://guava.dev/releases/19.0/api/docs/com/google/common/base/Optional.html

Modifier and Type | Method and Description
-- | --
static <T> Optional<T> | absent()Returns an Optional instance with no contained reference.
abstract Set<T> | asSet()Returns an immutable singleton Set whose only element is the contained instance if it is present; an empty immutable Set otherwise.
static <T> Optional<T> | fromNullable(T nullableReference)If nullableReference is non-null, returns an Optional instance containing that reference; otherwise returns absent().
abstract T | get()Returns the contained instance, which must be present.
abstract boolean | isPresent()Returns true if this holder contains a (non-null) instance.
static <T> Optional<T> | of(T reference)Returns an Optional instance containing the given non-null reference.
abstract Optional<T> | or(Optional<? extends T> secondChoice)Returns this Optional if it has a value present; secondChoice otherwise.
abstract T | or(Supplier<? extends T> supplier)Returns the contained instance if it is present; supplier.get() otherwise.
abstract T | or(T defaultValue)Returns the contained instance if it is present; defaultValue otherwise.
abstract T | orNull()Returns the contained instance if it is present; null otherwise.
static <T> Iterable<T> | presentInstances(Iterable<? extends Optional<? extends T>> optionals)Returns the value of each present instance from the supplied optionals, in order, skipping over occurrences of absent().
abstract <V> Optional<V> | transform(Function<? super T,V> function)If the instance is present, it is transformed with the given Function; otherwise, absent() is returned.

## JDK
https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html

Modifier and Type | Method and Description
-- | --
static <T> Optional<T> | empty()Returns an empty Optional instance.
Optional<T> | filter(Predicate<? super T> predicate)If a value is present, and the value matches the given predicate, return an Optional describing the value, otherwise return an empty Optional.
<U> Optional<U> | flatMap(Function<? super T,Optional<U>> mapper)If a value is present, apply the provided Optional-bearing mapping function to it, return that result, otherwise return an empty Optional.
[...]

PR #202, opened 2023-03-25
Replace Guava Optional with JDK Optional

timtebeek:
For #197 

Thought to try my hand at a first attempt; This should cover the most common cases.
Uncovered cases are added as tests annotated with `@ExpectedToFail`.

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/guava/PreferJavaUtilOptionalOrElseNull.java
+++ b/src/main/java/org/openrewrite/java/migrate/guava/PreferJavaUtilOptionalOrElseNull.java
@@ -1,0 +1,89 @@
+/*
+ * Copyright 2023 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.java.migrate.guava;
+
+import org.openrewrite.ExecutionContext;
+import org.openrewrite.Recipe;
+import org.openrewrite.java.JavaIsoVisitor;
+import org.openrewrite.java.JavaTemplate;
+import org.openrewrite.java.MethodMatcher;
+import org.openrewrite.java.search.UsesMethod;
+import org.openrewrite.java.search.UsesType;
+import org.openrewrite.java.tree.J;
+
+import java.time.Duration;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
+
+public class PreferJavaUtilOptionalOrElseNull extends Recipe {
+    @Override
+    public String getDisplayName() {
+        return "Prefer `java.util.Optional#orElse(null)` over `com.google.common.base.Optional#orNull()`";
+    }
+
+    @Override
+    public String getDescription() {
+        return "Replaces `com.google.common.base.Optional#orNull()` with `java.util.Optional#orElse(null)`.";
+    }
+
+    @Override
+    public Duration getEstimatedEffortPerOccurrence() {
+        return Duration.ofMinutes(5);
+    }
+
+    @Override
+    public Set<String> getTags() {
+        return new HashSet<>(Arrays.asList("RSPEC-4738", "guava"));
+    }
+
+    @Override
+    protected UsesType<ExecutionContext> getApplicableTest() {
+        return new UsesType<>("com.google.common.base.Optional");
+    }
+
+    @Override
+    protected UsesMethod<ExecutionContext> getSingleSourceApplicableTest() {
+        return new UsesMethod<>("com.google.common.base.Optional orNull()");
+    }
+
+    @Override
+    protected JavaIsoVisitor<ExecutionContext> getVisitor() {
+        return new PreferJavaUtilOptionalOrElseNullVisitor();
+    }
+
+    private static class PreferJavaUtilOptionalOrElseNullVisitor extends JavaIsoVisitor<ExecutionContext> {
+        private static final MethodMatcher guavaCreateTempDirMatcher = new MethodMatcher("com.google.common.base.Optional orNull()");
+
+        @Override
+        public J.CompilationUnit visitCompilationUnit(J.CompilationUnit cu, ExecutionContext executionContext) {
+            J.CompilationUnit c = super.visitCompilationUnit(cu, executionContext);
+            maybeAddImport("java.util.Optional");
+            maybeRemoveImport("com.google.common.base.Optional");
+            return c;
+        }
+
+        @Override
+        public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, ExecutionContext executionContext) {
+            J.MethodInvocation mi = super.visitMethodInvocation(method, executionContext);
+            if (guavaCreateTempDirMatcher.matches(mi)) {
+                return mi.withName(mi.getName().withSimpleName("orElse"))
+                        .withTemplate(JavaTemplate.builder(this::getCursor, "null").build(), mi.getCoordinates().replaceArguments());
+            }
+            return mi;
+        }
+    }
+}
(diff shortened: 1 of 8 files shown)


src/test/java/org/openrewrite/java/migrate/guava/PreferJavaUtilOptionalTest.java
```diff
@@ -0,0 +1,309 @@
...
+                  Optional<String> foo(Optional<String> firstChoice, Optional<String> secondChoice) {
+                      return firstChoice.or(secondChoice);
+                  }
+              }
+              """, """
+              import java.util.Optional;
+
+              class A {
+                  Optional<String> foo(Optional<String> firstChoice, Optional<String> secondChoice) {
+                      return a.isPresent()? firstChoice : secondChoice;
```
murdos, line 173: There's probably a typo: 
```suggestion
                      return firstChoice.isPresent()? firstChoice : secondChoice;
```
However I think this would be way more idiomatic that way (requires java >= 9):
```suggestion
                      return firstChoice.or(() -> secondChoice);
```

timtebeek: Thanks for spotting the issue and the improvement suggestion; the [docs specifically mentioned the predicate replacement](https://guava.dev/releases/31.1-jre/api/docs/com/google/common/base/Optional.html#or(com.google.common.base.Optional)). Now I'm trying to think if there's any side effects that might occur with `.or(() -> ...)`, but can't yet think of any. Would you agree there?

knutwannheden: Using `or()` is certainly better, because the ternary operator would either require introducing a variable or evaluate the first optional expression twice, unless it is a plain variable reference. I think there is also a convoluted alternative for Java 8.

timtebeek: Does remind me we should add a restriction to only run these on Java 9+...

src/main/java/org/openrewrite/java/migrate/guava/PreferJavaUtilOptionalOrElseNull.java
```diff
@@ -0,0 +1,89 @@
...
+        return new UsesMethod<>("com.google.common.base.Optional orNull()");
+    }
+
+    @Override
+    protected JavaIsoVisitor<ExecutionContext> getVisitor() {
+        return new PreferJavaUtilOptionalOrElseNullVisitor();
+    }
+
+    private static class PreferJavaUtilOptionalOrElseNullVisitor extends JavaIsoVisitor<ExecutionContext> {
+        private static final MethodMatcher guavaCreateTempDirMatcher = new MethodMatcher("com.google.common.base.Optional orNull()");
```
knutwannheden, line 69: Detail: Appears to be a copy-paste name here.

src/main/resources/META-INF/rewrite/no-guava.yml
```diff
@@ -70,6 +71,34 @@ recipeList:
       oldFullyQualifiedTypeName: com.google.common.base.Function
       newFullyQualifiedTypeName: java.util.function.Function
 
+---
+type: specs.openrewrite.org/v1beta/recipe
+name: org.openrewrite.java.migrate.guava.PreferJavaUtilOptional
+displayName: Prefer `java.util.Optional`
+description: Prefer `java.util.Optional` instead of using `com.google.common.base.Optional`.
+tags:
+  - guava, RSPEC-4738
```
knutwannheden, line 80: Multiple tags should be on separate lines. I can see that this was just copy-pasted, so we might as well correct the others.

src/main/java/org/openrewrite/java/migrate/guava/PreferJavaUtilOptionalOrElseNull.java
```diff
@@ -0,0 +1,89 @@
...
+            maybeRemoveImport("com.google.common.base.Optional");
+            return c;
+        }
+
+        @Override
+        public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, ExecutionContext executionContext) {
+            J.MethodInvocation mi = super.visitMethodInvocation(method, executionContext);
+            if (guavaCreateTempDirMatcher.matches(mi)) {
+                return mi.withName(mi.getName().withSimpleName("orElse"))
+                        .withTemplate(JavaTemplate.builder(this::getCursor, "null").build(), mi.getCoordinates().replaceArguments());
```
knutwannheden, line 84: I suspect this is actually the problem of the `ChangeType` recipe (I will investigate this further, when I find time), but as you can see in the following screenshot from the debugger, the type attribution of the `optional` parameter refers to the correct new type, but as its owner still refers to the old `JavaType.Method` object: 
image

src/test/java/org/openrewrite/java/migrate/guava/PreferJavaUtilOptionalTest.java
```diff
@@ -0,0 +1,309 @@
...
+                  String foo(Optional<String> optional) {
+                      return optional.orNull();
+                  }
+              }
+              """, """
+              import java.util.Optional;
+
+              class A {
+                  String foo(Optional<String> optional) {
+                      return optional.orElse(null);
```
knutwannheden, line 81: It looks like the type attribution is slightly messed up by the recipe. In the following screenshot you see the method type of the method invocation of the `orElse()` method. I was slightly surprised by the `String` return type, but I think this is normal for OpenRewrite as it uses `JavaType.Method` both for method declarations and method invocations and they are slightly different in this respect. What is clearly missing, however, is the parameter type.
image

knutwannheden (commented):
There appear to be some minor errors in the type attribution. I think we can however fix that in a follow-up commit.

timtebeek:
After the addition of the applicability test in 98739f2 the tests now fail with:
> PreferJavaUtilOptionalTest > orToOrElse() FAILED
    java.lang.AssertionError: [Recipe validation must have no failures] 
    Expecting empty but was: [Invalid{property='org.openrewrite.java.search.UsesJavaVersion', value='{majorVersionMin=9}', message='Recipe class org.openrewrite.java.search.UsesJavaVersion cannot be found'}]

Don't immediately know why; but working on other things at the moment.

timtebeek:
Fun to do some recipe development again; Are you OK to pickup the identified type issues separately @knutwannheden ?

src/main/resources/META-INF/rewrite/no-guava.yml
```diff
@@ -64,19 +65,53 @@ name: org.openrewrite.java.migrate.guava.PreferJavaUtilFunction
...
+  - org.openrewrite.java.migrate.guava.NoGuavaOptionalToJavaUtil
+  - org.openrewrite.java.ChangeMethodName:
+      methodPattern: com.google.common.base.Optional absent()
+      newMethodName: empty
+  - org.openrewrite.java.ChangeMethodName:
+      methodPattern: com.google.common.base.Optional fromNullable(..)
+      newMethodName: ofNullable
+  - org.openrewrite.java.ChangeMethodName:
+      methodPattern: com.google.common.base.Optional or(com.google.common.base.Supplier)
+      newMethodName: orElseGet
```
knutwannheden, line 96: This recipe works nicely if the supplier is specified as a lambda but fails when the supplier is represented by a variable / parameter of type `com.google.common.base.Supplier`. I think we can keep this as a known limitation for now (we could just add a test case for it so we don't forget).

knutwannheden: I just noticed that we also have recipes to migrate from Guava Supplier and Function to the corresponding JDK types. Should we add a test for this?

knutwannheden: I just tested this myself (by replacing the recipe used by `PreferJavaUtilOptionalTest` with `org.openrewrite.java.migrate.guava.NoGuava`) and it indeed works 😃  But the `transformToMap()` test failed. This test however works if we in `no-guava.yml` move the `PreferJavaUtilFunction` recipe below the `PreferJavaUtilOptional` recipe.

timtebeek: Thanks! Hadn't thought about those interacting indeed; now tested and fixed in 06a5a8d.

As for testing Supplier and Function: those are implemented as calls to ChangeType only, so I don't think we have to test those extensively here. The fix would likely be there for these functional interfaces.
https://github.com/openrewrite/rewrite-migrate-java/blob/9a74c66bf6698dd437dedb3fcd0f74f07fe9bc03/src/main/resources/META-INF/rewrite/no-guava.yml#L85-L95

timtebeek: Also added a test case `orSupplierToOrElseGetWithSupplierArgument` in bb48f4c to match the test case you mentioned (I think).

knutwannheden (approved):
👍

PR #209, opened 2023-04-16
Replace various StringUtils isEmpty and isNotEmpty with JDK calls

timtebeek:
As requested for Apache Maven 
https://issues.apache.org/jira/browse/MNG-6829
https://issues.apache.org/jira/browse/MNG-6825

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -24,6 +24,10 @@ dependencies {
     runtimeOnly("org.openrewrite:rewrite-java-11")
     runtimeOnly("org.openrewrite:rewrite-java-17")
 
+    implementation("org.apache.commons:commons-lang3:3.+")
+    implementation("org.codehaus.plexus:plexus-utils:3.+")
+    implementation("org.apache.maven.shared:maven-shared-utils:3.+")
+
     testImplementation("org.junit.jupiter:junit-jupiter-api:latest.release")
     testImplementation("org.junit.jupiter:junit-jupiter-params:latest.release")
     testImplementation("org.junit-pioneer:junit-pioneer:2.0.0")
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -24,10 +24,6 @@ dependencies {
     runtimeOnly("org.openrewrite:rewrite-java-11")
     runtimeOnly("org.openrewrite:rewrite-java-17")
 
-    implementation("org.apache.commons:commons-lang3:3.+")
-    implementation("org.codehaus.plexus:plexus-utils:3.+")
-    implementation("org.apache.maven.shared:maven-shared-utils:3.+")
-
     testImplementation("org.junit.jupiter:junit-jupiter-api:latest.release")
     testImplementation("org.junit.jupiter:junit-jupiter-params:latest.release")
     testImplementation("org.junit-pioneer:junit-pioneer:2.0.0")
@@ -41,6 +37,9 @@ dependencies {
     testImplementation("com.google.guava:guava:29.0-jre")
 
     testImplementation("commons-codec:commons-codec:1.+")
+    testImplementation("org.apache.commons:commons-lang3:3.+")
+    testImplementation("org.codehaus.plexus:plexus-utils:3.+")
+    testImplementation("org.apache.maven.shared:maven-shared-utils:3.+")
 
     testRuntimeOnly("commons-io:commons-io:2.+")
     testRuntimeOnly("com.fasterxml.jackson.datatype:jackson-datatype-jsr353:latest.release")
--- a/src/main/java/org/openrewrite/java/migrate/apache/commons/lang/IsNotEmptyToJdk.java
+++ b/src/main/java/org/openrewrite/java/migrate/apache/commons/lang/IsNotEmptyToJdk.java
@@ -45,6 +45,14 @@ public class IsNotEmptyToJdk extends Recipe {
             private final MethodMatcher mavenSharedIsEmptyMatcher = new MethodMatcher(org.apache.maven.shared.utils.StringUtils.class.getName() + " isEmpty(..)");
             private final MethodMatcher mavenSharedIsNotEmptyMatcher = new MethodMatcher(org.apache.maven.shared.utils.StringUtils.class.getName() + " isNotEmpty(..)");
 
+            private final JavaTemplate commonsIsEmptyTemplate = JavaTemplate.compile(this, "CommonsIsEmpty", (String s) -> s == null || s.isEmpty()).build();
+            private final JavaTemplate commonsIsNotEmptyTemplate = JavaTemplate.compile(this, "CommonsIsNotEmpty", (String s) -> s != null && !s.isEmpty()).build();
+            private final JavaTemplate plexusIsEmptyTemplate = JavaTemplate.compile(this, "PlexusIsEmpty", (String s) -> s == null || s.isEmpty()).build();
+            private final JavaTemplate plexusIsNotEmptyTemplate = JavaTemplate.compile(this, "PlexusIsNotEmpty", (String s) -> s != null && !s.isEmpty()).build();
+            private final JavaTemplate mavenSharedIsEmptyTemplate = JavaTemplate.compile(this, "MavenSharedIsEmpty", (String s) -> s == null || s.isEmpty()).build();
+            private final JavaTemplate mavenSharedIsNotEmptyTemplate = JavaTemplate.compile(this, "MavenSharedIsNotEmpty", (String s) -> s != null && !s.isEmpty()).build();
+
+
             @Override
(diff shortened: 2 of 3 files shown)


src/main/java/org/openrewrite/java/migrate/apache/commons/lang/IsNotEmptyToJdk.java
```diff
@@ -0,0 +1,90 @@
...
+            private final MethodMatcher mavenSharedIsNotEmptyMatcher = new MethodMatcher(org.apache.maven.shared.utils.StringUtils.class.getName() + " isNotEmpty(..)");
+
+            private final JavaTemplate commonsIsEmptyTemplate = JavaTemplate.compile(this, "CommonsIsEmpty", (String s) -> (s == null || s.isEmpty())).build();
+            private final JavaTemplate commonsIsNotEmptyTemplate = JavaTemplate.compile(this, "CommonsIsNotEmpty", (String s) -> (s != null && !s.isEmpty())).build();
+            private final JavaTemplate plexusIsEmptyTemplate = JavaTemplate.compile(this, "PlexusIsEmpty", (String s) -> (s == null || s.isEmpty())).build();
+            private final JavaTemplate plexusIsNotEmptyTemplate = JavaTemplate.compile(this, "PlexusIsNotEmpty", (String s) -> (s != null && !s.isEmpty())).build();
+            private final JavaTemplate mavenSharedIsEmptyTemplate = JavaTemplate.compile(this, "MavenSharedIsEmpty", (String s) -> (s == null || s.isEmpty())).build();
+            private final JavaTemplate mavenSharedIsNotEmptyTemplate = JavaTemplate.compile(this, "MavenSharedIsNotEmpty", (String s) -> (s != null && !s.isEmpty())).build();
```
timtebeek, line 54: As also indicated below: the `name` can only be a literal; otherwise we could have reduced the duplication here.

src/main/java/org/openrewrite/java/migrate/apache/commons/lang/IsNotEmptyToJdk.java
```diff
@@ -0,0 +1,90 @@
...
+    @Override
+    protected TreeVisitor<?, ExecutionContext> getVisitor() {
+        return new JavaVisitor<ExecutionContext>() {
+
+            private final MethodMatcher commonsIsEmptyMatcher = new MethodMatcher(org.apache.commons.lang3.StringUtils.class.getName() + " isEmpty(..)");
+            private final MethodMatcher commonsIsNotEmptyMatcher = new MethodMatcher(org.apache.commons.lang3.StringUtils.class.getName() + " isNotEmpty(..)");
+            private final MethodMatcher plexusIsEmptyMatcher = new MethodMatcher(org.codehaus.plexus.util.StringUtils.class.getName() + " isEmpty(..)");
+            private final MethodMatcher plexusIsNotEmptyMatcher = new MethodMatcher(org.codehaus.plexus.util.StringUtils.class.getName() + " isNotEmpty(..)");
+            private final MethodMatcher mavenSharedIsEmptyMatcher = new MethodMatcher(org.apache.maven.shared.utils.StringUtils.class.getName() + " isEmpty(..)");
+            private final MethodMatcher mavenSharedIsNotEmptyMatcher = new MethodMatcher(org.apache.maven.shared.utils.StringUtils.class.getName() + " isNotEmpty(..)");
```
timtebeek, line 47: Using method matchers rather than template matchers, as I'm not sure the overlapping StringUtils classes would be separated correctly.

src/test/java/org/openrewrite/java/migrate/apache/commons/lang/IsNotEmptyToJdkTest.java
```diff
@@ -0,0 +1,124 @@
...
+    @ParameterizedTest
+    @CsvSource(delimiter = '#', textBlock = """
+      org.apache.commons.lang3.StringUtils # !StringUtils.isEmpty(first) # !(first == null || first.isEmpty())
+      org.apache.commons.lang3.StringUtils # !StringUtils.isNotEmpty(first) # !(first != null && !first.isEmpty())
+      org.apache.commons.lang3.StringUtils # !(StringUtils.isEmpty(first)) # !(first == null || first.isEmpty())
+      org.apache.commons.lang3.StringUtils # !(StringUtils.isNotEmpty(first)) # !(first != null && !first.isEmpty())
+      org.apache.maven.shared.utils.StringUtils # !StringUtils.isEmpty(first) # !(first == null || first.isEmpty())
+      org.apache.maven.shared.utils.StringUtils # !StringUtils.isNotEmpty(first) # !(first != null && !first.isEmpty())
+      org.codehaus.plexus.util.StringUtils # !StringUtils.isEmpty(first) # !(first == null || first.isEmpty())
+      org.codehaus.plexus.util.StringUtils # !StringUtils.isNotEmpty(first) # !(first != null && !first.isEmpty())
```
timtebeek, line 74: Would have been nicer if we could drop the negation; but `visitUnary` only happens after `visitMethodInvocation`, so if we want to invert the method calls we would need another recipe to run first.

Alternatively we don't yet have a recipe to push down the negation it seems; otherwise that could have been a doAfterVisit.

src/test/java/org/openrewrite/java/migrate/apache/commons/lang/IsNotEmptyToJdkTest.java
```diff
@@ -0,0 +1,124 @@
...
+      org.apache.commons.lang3.StringUtils # StringUtils.isNotEmpty(first + second)
+      org.apache.maven.shared.utils.StringUtils # StringUtils.isEmpty(foo())
+      org.codehaus.plexus.util.StringUtils # StringUtils.isEmpty(foo())
+      """)
+    void retain(String classname, String beforeLine) {
+        // language=java
+        rewriteRun(
+          java(
+            """
+              %s;
```
timtebeek, line 112:
```suggestion
              import %s;
```

src/main/java/org/openrewrite/java/migrate/apache/commons/lang/IsNotEmptyToJdk.java
knutwannheden: As the templates look identical I am unsure why we can't just use the same two templates everywhere.

src/main/java/org/openrewrite/java/migrate/apache/commons/lang/IsNotEmptyToJdk.java
knutwannheden: Yes, makes sense. This is a case where I am unsure if using a template to match actually even helps.

src/test/java/org/openrewrite/java/migrate/apache/commons/lang/IsNotEmptyToJdkTest.java
timtebeek: I'd also tried to determine from the method invocation if that's preceded by an exclamation mark, such that we can return the alternative replacement template, but was not able to do so; Any suggestions welcome!

src/main/java/org/openrewrite/java/migrate/apache/commons/lang/IsNotEmptyToJdk.java
timtebeek: You're right; had the same insight the moment I stepped away from the laptop; should do that more often! ;)

src/test/java/org/openrewrite/java/migrate/apache/commons/lang/IsNotEmptyToJdkTest.java
knutwannheden: I think we should do it the way you first thought: Implement a dedicated visitor which is applied to the unary expression.

Possibly we can come up with some additional patterns we can integrate into `SimplifyBooleanExpression`. E.g. if a binary contains at least one negated unary or another binary with a relational operator. I am just uncertain, because which version is simpler is often debatable and situation dependent.

build.gradle.kts
```diff
@@ -24,6 +24,10 @@ dependencies {
     runtimeOnly("org.openrewrite:rewrite-java-11")
     runtimeOnly("org.openrewrite:rewrite-java-17")
 
+    implementation("org.apache.commons:commons-lang3:3.+")
+    implementation("org.codehaus.plexus:plexus-utils:3.+")
+    implementation("org.apache.maven.shared:maven-shared-utils:3.+")
```
knutwannheden, line 29: Looks like we could change the scope of these to `testImplementation()` or similar, right?

knutwannheden (approved):
Unless you also want to take inspiration from `SimplifyBooleanExpression` to implement the unary negation, I think this PR looks good.

PR #211, opened 2023-04-18
Update UpgradeJavaVersion, add new recipe to update java version in the `build.gradle` file

kunli2:
For UpgradeJavaVersion
1. Added a recipe to upgrade Java version `java.toolchain.languageVersion` in `build.gradle(.kts)`
2. Refactor `org.openrewrite.java.migrate.JavaVersion17` to 3 recipes to have a dedicated function, 
  - MarkJavaVersion 
  - UpgradeJavaVersionMaven
  - UpgradeJavaVersionGradle


src/main/java/org/openrewrite/java/migrate/UpgradeJavaVersionGradle.java
```diff
@@ -0,0 +1,77 @@
...
+        return new GroovyIsoVisitor<ExecutionContext>() {
+            MethodMatcher javaLanguageVersionMatcher = new MethodMatcher("org.gradle.jvm.toolchain.JavaLanguageVersion of(int)");
+
+            @Override
+            public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method,
+                                                            ExecutionContext executionContext) {
+                method = super.visitMethodInvocation(method, executionContext);
+                if (javaLanguageVersionMatcher.matches(method)) {
+                    List<Expression> args = method.getArguments();
+                    J.Literal versionNumber = (J.Literal) args.get(0);
```
sambsnyd, line 67: It will not always be the case that this is a literal, this unchecked cast is not safe. Rather than throwing an exception you should return early and use `Markup.warn()` to notify that the recipe wanted to make a change and could not.

src/main/java/org/openrewrite/java/migrate/UpgradeJavaVersionGradle.java
```diff
@@ -0,0 +1,77 @@
...
+import org.openrewrite.java.MethodMatcher;
+import org.openrewrite.java.tree.Expression;
+import org.openrewrite.java.tree.J;
+
+import java.util.Collections;
+import java.util.List;
+
+@Value
+@EqualsAndHashCode(callSuper = true)
+public class UpgradeJavaVersionGradle extends Recipe {
```
sambsnyd, line 33: Feels like this could still be a single recipe which handles both Maven and Gradle updates and marker updates. Then yaml files could remain unchanged.

kunli2: Changed to be 1 recipe to do all the upgrade work.

PR #215, opened 2023-04-26
Fix escaping bugs in UseTextBlocks

aksh1618:
I recently happened upon this amazing project, kudos to all involved, I absolutely love the idea and implementation! I decided to use it for one of my projects, and it worked very well except for some minor cases, for which I decided to submit fixes:
- Backslashes are not escaped properly, for example in case of a double escaped quote
- Triple quotes are not escaped properly
- Added a test for the special case of ending quote as well for completion

Thanks again for making this wonderful project, let me know if any further changes required here!

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/lang/UseTextBlocks.java
+++ b/src/main/java/org/openrewrite/java/migrate/lang/UseTextBlocks.java
@@ -142,6 +142,7 @@ public class UseTextBlocks extends Recipe {
 
                 boolean isEndsWithNewLine = content.endsWith("\n");
                 content = content.replace("\\", "\\\\"); // Escape backslashes
+                content = content.replace("\"\"\"", "\"\"\\\""); // Escape triple quotes
                 content = content.replace(" \n", "\\s\n");
                 content = content.replace("\n", "\n" + indentation);
                 content = content.replace(passPhrase, "\\\n" + indentation);
--- a/src/main/java/org/openrewrite/java/migrate/lang/UseTextBlocks.java
+++ b/src/main/java/org/openrewrite/java/migrate/lang/UseTextBlocks.java
@@ -141,6 +141,7 @@ public class UseTextBlocks extends Recipe {
                 String indentation = getIndents(concatenation.toString(), useTab, tabSize);
 
                 boolean isEndsWithNewLine = content.endsWith("\n");
+                content = content.replace("\\", "\\\\"); // Escape backslashes
                 content = content.replace(" \n", "\\s\n");
                 content = content.replace("\n", "\n" + indentation);
                 content = content.replace(passPhrase, "\\\n" + indentation);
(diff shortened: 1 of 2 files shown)


src/main/java/org/openrewrite/java/migrate/lang/UseTextBlocks.java
```diff
@@ -141,14 +141,27 @@ private J.Literal toTextBlock(J.Binary binary, String content, List<J.Literal> s
...
                 boolean isEndsWithNewLine = content.endsWith("\n");
+
+                // references:
+                //  - https://docs.oracle.com/en/java/javase/14/docs/specs/text-blocks-jls.html
+                //  - https://javaalmanac.io/features/textblocks/
+
+                // escape backslashes
+                content = content.replace("\\", "\\\\");
+                // escape triple quotes
+                content = content.replace("\"\"\"", "\"\"\\\"");
```
kunli2, line 152: do you know why the triple quotes are transformed to `\"\"\\\"`  but not `\\\"\"\"`, the difference is the 1st or 3rd quote is escaped.
I found for an 8 quotes case: 
```
        String s1 = "\"\"\"\"\"\"\"\"";
        String s2 = """
                                \"""\"""\""\
                                """;
        String s3 = """
                                \"""\"""\""\
                                """;
```
s1 and s2 are equivalent, however, s3 doesn't compile unless change it to 
```
        String s3 = """
                                \"""\"""\""\
                                """;
```
which makes me think to escape the 3rd quote instead of 1st is like magic.
Do you have any insight of this?

aksh1618: @kunli2 I went for escaping the third quote instead of the first to optimize for the number of escapes needed in case of n number of continuous quotes (For example: 2 escapes instead of 3 in case of 8 quotes). Another thing I noticed was that IDEA by default escapes the third quote as well, for instance if you copy-paste a series of quotes inside a text block.

In your examples, I can't see any difference between s2 & s3 (both), did GitHub formatting probably make them look the same? Could you share it in some other form such as a gist so that I can try to understand I better, maybe then I could answer it better?.

kunli2: hi @aksh1618 , yes, I also found IDE by default escapes the 3rd quote.
what I noticed is below, the first statement in the code doesn't compile (`\"""\"""""` doesn't compile but `\"""\"""\""\` does). I guess that is the reason why IDEA  escapes the 3rd quote instead of 1st.

```
String s3 = """
                                \"""\"""""\
                                """;
String s3 = """
                                \"""\"""\""\
                                """;
```

aksh1618: @kunli2 thanks for clarifying, it's a very good example! So to summarize, in this case if we're replacing programmatically successively, escaping the first one would not work as it leads to code that doesn't compile due to 4 quotes left together. So now it makes sense why they went with the third one as well instead of having to write a more complex replacement algorithm for handling leftover quotes :+1:

kunli2: @aksh1618 yeah, it makes sense! thanks for fixing and contribution again!

Issue #217, opened 2022-04-16
Java 10+ Local-Variable Type Inference should be used `RSPEC-6212`


yeikel:
This is controversial and the Sonar team had to remove it from the default "Sonar Way" profile

https://rules.sonarsource.com/java/tag/java10/RSPEC-6212

timtebeek:
Thank you for your offer to help out on this one! I think you have a decent list of restrictions in place already to limit applicability. I'd indeed say not to use this for primitives, and even then it's still not always a clear improvement, which makes it a tricky recipe to recommend.

Might be good to take into account [the upcoming changes for OpenRewrite 8.0](https://github.com/openrewrite/rewrite-migrate-java/pull/216), which will have change the applicability tests for instance.

Maybe we can start with a suite of unit tests and do a round of feedback on that? Then we can more clearly decide which cases are worth covering, before we look into the finer details of the recipe implementation. Feel free to to open a draft pull request with what cases you would (not) want to cover as part of this work!

yeikel:
> I'd like to work on this one and would like to clarify a few things.
> 
> To be sure, do we want to implement [RSPEC-6212](https://rules.sonarsource.com/java/tag/java10/RSPEC-6212) 100% Sonar like? I would argue against it because it also suggests using `var` for primitives, which is not recommended, as you can see in Project Ambers [Local Variable Type Inference - G7](https://openjdk.org/projects/amber/guides/lvti-style-guide#G7).

`boolean` seems to be an exception for this rule, right? (so not all primitives are excluded)

MBoegers:
> boolean seems to be an exception

Good point, not all usages of primitives are bad.
- boolean, char and String are okay
- float, double and long are okay, iff initialized with correct typed `var aLong = 1.0L; var aFloat = 1.0F; var aDouble = 1.0; var otherDouble = 1.0D`
- short and byte are dangerous in most cases
- reuse for initialization is dangerous `static final float INITIAL = 3.0f; /*...*/ double temp = INITIAL;` is possible but with var inferred as float

So for primitives I suggest:
```
IF String, boolean, char, float with 'F', long with 'L', double THEN
   apply var transformation
IF shot, byte THEN
   do not apply var transformation
IF initalized by variable AND left hand side type matches THEN
    apply var transformation
```

MBoegers:
What about not correctly used generics?
[...]

PR #218, opened 2023-05-03
Java 10+ Local-Variable Type Inference

MBoegers:
Implement a recipe for local variable type inference as discussed here: https://github.com/openrewrite/rewrite-migrate-java/issues/217

This Draft contains the test suite, as tim suggested.

### DIFF
--- a/src/test/java/org/openrewrite/java/migrate/lang/VarUsageTest.java
+++ b/src/test/java/org/openrewrite/java/migrate/lang/VarUsageTest.java
@@ -1,0 +1,225 @@
+/*
+ * Copyright 2021 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.java.migrate.lang;
+
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
+import org.openrewrite.test.RecipeSpec;
+import org.openrewrite.test.RewriteTest;
+import org.openrewrite.test.TypeValidation;
+
+import static org.openrewrite.java.Assertions.java;
+import static org.openrewrite.java.Assertions.version;
+
+class VarUsageTest implements RewriteTest {
+    public void defaults(RecipeSpec spec) {
+        spec.recipe(new VarKeyword())
+          .typeValidationOptions(new TypeValidation().methodInvocations(false));
+    }
+
+    @Test
+    void donotTransformNull() {
+        //language=java
+        rewriteRun(
+          version(
+            java("""
+              package com.example.app;
+                        
+              class A {
+                void m() {
+                  String str = null;
+                }
+              }
+              """),
+            10
+          )
+        );
+    }
+
+    @Nested
+    class TransformObject {
+        @Test
+        void transformObjectInMethodBody() {
+            //language=java
+            rewriteRun(
+              version(
+                java("""
+                  package com.example.app;
+                                    
+                  class A {
+                    void m() {
+                        Object o = new Object();
+                    }
+                  }
+                  """, """
+                  package com.example.app;
+                                    
+                  class A {
+                    void m() {
+                      var o = new Object();
+                    }
+                  }
+                  """),
+                10
+              )
+            );
+        }
+
+        @Test
+        void donotTransformObjectInParameter() {
+            //language=java
+            rewriteRun(
+              version(
+                java("""
+                  package com.example.app;
+                                    
+                  class A {
+                    Object m(Object o) {
+                        return o;
+                    }
+                  }
+                  """),
+                10
+              )
+            );
+        }
+
+        @Test
+        void donotTransformObjectInClass() {
+            //language=java
+            rewriteRun(
+              version(
+                java("""
+                  package com.example.app;
+                                    
+                  class A {
+                    Object o;
+                    Object m() {
+                        return o;
+                    }
+                  }
+                  """),
+                10
+              )
+            );
+        }
+    }
+
+    @Nested
+    class Generics {
+        @Test
+        void transformFullGenerics() {
+            //language=java
+            rewriteRun(
+              version(
+                java("""
+                  package com.example.app;
+                                    
+                  class A {
+                    void m() {
+                        List<String> strs = new ArrayList<String>();
+                    }
+                  }
+                  """, """
+                  package com.example.app;
+                                    
+                  class A {
+                    void m() {
+                      var str = new ArrayList<String>();
+                    }
+                  }
+                  """),
+                10
+              )
+            );
+        }
+
+        @Test
+        void transformGenericsWithDiamond() {
+            //language=java
+            rewriteRun(
+              version(
+                java("""
+                  package com.example.app;
+                                    
+                  class A {
+                    void m() {
+                        List<String> strs = new ArrayList<>();
+                    }
+                  }
+                  """, """
+                  package com.example.app;
+                    
+                  class A {
+                    void m() {
+                      var str = new ArrayList<String>();
+                    }
+                  }
+                  """),
+                10
+              )
+            );
+        }
+
+        @Test
+        void transformGenericsOnFactory() {
+            //language=java
+            rewriteRun(
+              version(
+                java(
+                  """
+                    package com.example.app;
+                    
+                    import java.util.ArrayList;class A {
+                      void m() {
+                          List<String> strs = List.of("one","two");
+                      }
+                    }
+                    """, """
+                    package com.example.app;
+                    
+                    class A {
+                      void m() {
+                        var str = List.of("one","two");
+                      }
+                    }
+                    """),
+                10
+              )
+            );
+        }
+
+        @Test
+        void donotTransformGenericsOnEmptyFactory() {
+            //language=java
+            rewriteRun(
+              version(
+                java(
+                  """
+                      package com.example.app;
+                      
+                      class A {
+                        void m() {
+                          List<String> strs = List.of();
+                        }
+                      }
+                    """),
+                10
+              )
+            );
+        }
+    }
+}
--- a/src/main/java/org/openrewrite/java/migrate/lang/VarKeyword.java
+++ b/src/main/java/org/openrewrite/java/migrate/lang/VarKeyword.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright 2021 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.java.migrate.lang;
+
+import lombok.EqualsAndHashCode;
+import lombok.Value;
+import org.openrewrite.ExecutionContext;
+import org.openrewrite.Recipe;
+import org.openrewrite.TreeVisitor;
+import org.openrewrite.internal.lang.Nullable;
+import org.openrewrite.java.JavaVisitor;
+import org.openrewrite.java.search.HasJavaVersion;
+
+import java.time.Duration;
+
+@Value
+@EqualsAndHashCode(callSuper = false)
+public class VarKeyword extends Recipe {
+    public String getDisplayName() {
+        return "Use local variable type-inference (var) where possible";
+    }
+
+    @Override
+    public String getDescription() {
+        return "Local variable type-inference reduce the noise produces by repeating the type definitions in Java 10 or higher.";
+    }
+
+    @Override
+    public @Nullable Duration getEstimatedEffortPerOccurrence() {
+        return Duration.ofMinutes(1);
+    }
+
+    @Override
+    protected TreeVisitor<?, ExecutionContext> getSingleSourceApplicableTest() {
+        return new HasJavaVersion("11", true).getVisitor();
+    }
+
+    @Override
+    protected TreeVisitor<?, ExecutionContext> getVisitor() {
+        return new JavaVisitor<ExecutionContext>() {
+        };
+    }
+}


src/test/java/org/openrewrite/java/migrate/lang/VarUsageTest.java
```diff
@@ -0,0 +1,225 @@
...
+        @Test
+        void transformGenericsOnFactory() {
+            //language=java
+            rewriteRun(
+              version(
+                java(
+                  """
+                    package com.example.app;
+                    
+                    import java.util.ArrayList;class A {
```
timtebeek, line 186:
```suggestion
                    import java.util.ArrayList;
                    class A {
```

src/test/java/org/openrewrite/java/migrate/lang/VarUsageTest.java
```diff
@@ -0,0 +1,225 @@
...
+import org.openrewrite.test.RewriteTest;
+import org.openrewrite.test.TypeValidation;
+
+import static org.openrewrite.java.Assertions.java;
+import static org.openrewrite.java.Assertions.version;
+
+class VarUsageTest implements RewriteTest {
+    public void defaults(RecipeSpec spec) {
+        spec.recipe(new VarKeyword())
+          .typeValidationOptions(new TypeValidation().methodInvocations(false));
```
timtebeek, line 30: If we add the missing imports it should not be necessary to ignore type validation.
```suggestion
        spec.recipe(new VarKeyword());
```

timtebeek:
Might want to explicitly state what we want to do with primitives through tests; that could help indicate what we end up doing is intentional; great start!

Perhaps as a suggestion: right now your tests are a mix of should & should not apply recipe. Could be helpful to separate those into nested classes each such that you can drop the prefix.

src/test/java/org/openrewrite/java/migrate/lang/VarUsageTest.java
MBoegers: TBH I simply copied from your org.openrewrite.java.migrate.lang.StringFormattedTest :D 
Is there any reference or documentation when to use further customizations or do I missed downloading Sources with Javadoc?

src/test/java/org/openrewrite/java/migrate/lang/VarUsageTest.java
MBoegers: Oh Oo thought I catched all of them..

src/test/java/org/openrewrite/java/migrate/lang/VarUsageTest.java
timtebeek: We have some details about testing recipes on these URLs; not specific to this particular case though:
- https://docs.openrewrite.org/authoring-recipes/recipe-testing#recipespec
If you feel you're missing some details feel free to let us know how we can improve! :)

MBoegers:
Yes agree primitives are missing, see my comment [here](https://github.com/openrewrite/rewrite-migrate-java/issues/217#issuecomment-1533218927).

You suggest grouping the test by applicability instead, currently they are grouped as in the issue.
I can live with both ;)

MBoegers:
Ops.. I forgot to push the nested test version. Currently, GH seems to have issues with PRs.
I think I encountered a problem with the parsing of `var`
The following snippet fails with  _Syntax error at line 1:20 extraneous input '<EOF>' expecting {')', FullyQualifiedName, Number, Identifier}._ in rewrite-migrate-java.

```java
@Test
void varUsage() {
    //language=java
    rewriteRun(
      version(
        java("""
      package com.example.app;
                
      class A {
        void m() {
          var str1 = "Hello World!";
        }
      }
      """),
        10
      )
    );
}
```

I found a very similar but passing test in `org.openrewrite.java.tree.VariableDeclarationsTest#typeOnVarKeyword`.

MBoegers:
GH and I had issues the last days.. But now I updated the Test as we discussed

timtebeek:
Thanks a lot @MBoegers ; I'm asking my colleagues to have a look! :)

MBoegers:
@timtebeek cheers! I have not yet implemented generics, because I felt it really hard to detect them securely. I'd like to get a review for the _simple_ parts and aim for generics a bit later ;)

src/main/java/org/openrewrite/java/migrate/lang/UseVarKeyword.java
```diff
@@ -0,0 +1,162 @@
...
+            private final JavaType.Primitive SHORT_TYPE = JavaType.Primitive.Short;
+            private final JavaType.Primitive BYTE_TYPE = JavaType.Primitive.Byte;
+            private final JavaTemplate template = JavaTemplate.builder(this::getCursor, "var #{} = #{any()}")
+                    .javaParser(JavaParser.fromJavaVersion()).build();
+
+            @Override
+            public J.VariableDeclarations visitVariableDeclarations(J.VariableDeclarations multiVariable, ExecutionContext executionContext) {
+                J.VariableDeclarations vd = (J.VariableDeclarations) super.visitVariableDeclarations(multiVariable, executionContext);
+
+                boolean isOutsideMethode = !determineIfIsInsideMethode(this.getCursor());
```
knutwannheden, line 75: In a few places we can replace `methode` with `method` (also `Methode` with `Methode`).

src/main/java/org/openrewrite/java/migrate/lang/UseVarKeyword.java
```diff
@@ -0,0 +1,162 @@
...
+            private boolean determineIfMethodeParameter(@NotNull J.VariableDeclarations vd, @NotNull Cursor cursor) {
+                J.MethodDeclaration methodDeclaration = cursor.firstEnclosing(J.MethodDeclaration.class);
+                return nonNull(methodDeclaration) && methodDeclaration.getParameters().contains(vd);
+            }
+
+            /**
+             * Determines if a cursor is contained inside a Methode declaration without an intermediate Class declaration
+             * @param cursor value to determine
+             */
+            private boolean determineIfIsInsideMethode(@NotNull Cursor cursor) {
```
knutwannheden, line 109: I noticed that this excludes static and non-static initializers as in:
```java
class Foo {
    {
        Object o = new Object();
    }
    static {
        Object o = new Object();
    }
}
```
But feel free to supply a follow-up PR for that.

src/main/java/org/openrewrite/java/migrate/lang/UseVarKeyword.java
```diff
@@ -0,0 +1,162 @@
...
+                boolean isByteVariable = typeExpression instanceof J.Primitive && BYTE_TYPE.equals(typeExpression.getType());
+                boolean isShortVariable = typeExpression instanceof J.Primitive && SHORT_TYPE.equals(typeExpression.getType());
+                if (isByteVariable || isShortVariable) return vd;
+
+                boolean definesNoVariable = vd.getVariables().isEmpty();
+                boolean isCompoundDefinition = vd.getVariables().size() < 1;
+                boolean isPureAssigment = isNull(vd.getTypeExpression());
+                if (definesNoVariable || isCompoundDefinition || isPureAssigment) return vd;
+
+                Expression initializer = vd.getVariables().get(0).getInitializer();
```
knutwannheden, line 89: The initializer should be unwrapped, if it is inside parentheses, otherwise you end up with `var` when you have `Object o = (null);`, which wouldn't compile. You can call `Expression#unwrap()` for this. But even better is probably if you call `getType()` on the expression and check against `JavaType.Primitive.Null`.

src/main/java/org/openrewrite/java/migrate/lang/UseVarKeyword.java
```diff
@@ -0,0 +1,162 @@
...
+                boolean isMethodeParameter = determineIfMethodeParameter(vd, this.getCursor());
+                if (isOutsideMethode || isMethodeParameter) return vd;
+
+                TypeTree typeExpression = vd.getTypeExpression();
+                boolean isByteVariable = typeExpression instanceof J.Primitive && BYTE_TYPE.equals(typeExpression.getType());
+                boolean isShortVariable = typeExpression instanceof J.Primitive && SHORT_TYPE.equals(typeExpression.getType());
+                if (isByteVariable || isShortVariable) return vd;
+
+                boolean definesNoVariable = vd.getVariables().isEmpty();
+                boolean isCompoundDefinition = vd.getVariables().size() < 1;
```
knutwannheden, line 85:
```suggestion
                boolean isCompoundDefinition = vd.getVariables().size() > 1;
```

Also, I suggest to add a corresponding test.

knutwannheden (commented):
I think the PR looks really good! Some small comments:
- I found a problem with compound definitions, which currently would do the wrong thing.
- The null-type checking can also be improved
- I also noticed that it would currently not apply to code inside static and non-static initializers
- It might be cool to also support the variables of enhanced for-loops (e.g. `for (var s : List.of("a", "b"))`)

src/main/java/org/openrewrite/java/migrate/lang/UseVarKeyword.java
MBoegers: 🫣

src/main/java/org/openrewrite/java/migrate/lang/UseVarKeyword.java
MBoegers: Oh yes! I searched for `Primitive.Null` but somehow without success. Cheers

src/main/java/org/openrewrite/java/migrate/lang/UseVarKeyword.java
MBoegers: Yes totally! And as well, for your suggestion about `for(var s : List.of("Hello", "world", "!")) { /* ...*/ }` 
I've checked the JLS, and it seems like I misused the term "compound" see [JLS SE 8: 16.26.2](https://docs.oracle.com/javase/specs/jls/se8/html/jls-15.html#jls-15.26.2) combine it to `boolean definedSingleVariable = vd.getVariables().size() == 1;`. I think it is better in terms of semantic and naming ;)

src/main/java/org/openrewrite/java/migrate/lang/UseVarKeyword.java
MBoegers: Thanks for the review. 

>     * I found a problem with compound definitions, which currently would do the wrong thing.
_compound_ is the wrong term, multiple variable definition would be bette and should be covered by `UseVarKeywordTest#GeneralNotApplicable#simpleAssigment`.
According to [Java SE 8 Language Spec 15.26.2|https://docs.oracle.com/javase/specs/jls/se8/html/jls-15.html#jls-15.26.2] compound assignments are reassignments and therefore not a target for local variable type inference.

>     * I also noticed that it would currently not apply to code inside static and non-static initializers
>     * It might be cool to also support the variables of enhanced for-loops (e.g. `for (var s : List.of("a", "b"))`)
Yes they are not supported, yet. 
As I think about cool stuff, not covered stuff:
* initialization with ternary operator
* initialization via method invocations

Would I be an acceptable way to implement tests as marker and postpone the implementation to a followup?

knutwannheden:
@MBoegers Thanks for looking into this! Regarding your question:
> Would I be an acceptable way to implement tests as marker and postpone the implementation to a followup?

Definitely! The recipe doesn't need to be perfect from day one. We do however have the [do no harm guideline](https://docs.openrewrite.org/authoring-recipes/recipe-conventions-and-best-practices#do-no-harm) which boils down to that a recipe should rather not make any change than a wrong change. This can in itself already by quite a challenge (and we certainly still have some recipes which violate this rule), but we should make a best effort to not produce any wrong code. So if we can fix the few cases I think I found, I think we can go ahead and integrate this recipe and do improvements in follow-up commits.

MBoegers:
I've updated the PR, as you may see I excluded some features for now. They should be a new issue, otherwise this PRs lifetime explodes :D

What is possible by now:
- direct assignments and reassignments for primitive and boxed primitive
- inside static and instance initializer blocks as well as methods
- skipping of var usage, fields definitions, method parameters, multiple variable declarations and null assignments 

The leftovers:
- Generics and Method invocation in general, I think we should spend some more time defining what should work for generics and especially what *not*
- Ternary operator, similar to generics, this needs some more spec work
- Enhanced for loop, is tightly tied to generics
- Splitting the recipe into small recipes, handling each sub feature. Like Methods, Primitives, Ternary and the _easy_ direct assignments

Should I write some issues like I have them in mind for further refinement?

MBoegers:
I've extracted to open work for ternary operator support to #237 and support for method invocation to #236.

src/main/java/org/openrewrite/java/migrate/lang/var/UseVarForObject.java
```diff
@@ -0,0 +1,171 @@
...
+    public String getDescription() {
+        //language=markdown
+        return "Try to apply local variable type inference `var` to variables containing Objects where possible." +
+               "This recipe will not touch variable declaration with genrics or initializer containing ternary operators.";
+    }
+
+    @Override
+    public List<RecipeExample> getExamples() {
+        RecipeExample recipeExample = new RecipeExample();
+        List<RecipeExample.Source> examples = new ArrayList<>();
```
timtebeek, line 54: I think we typically do this by annotating representative tests with `@DocumentExample`.

timtebeek: I suggest we follow that approach and remove this method override; unless there's a convincing reason.

src/main/java/org/openrewrite/java/migrate/lang/var/UseVarForObject.java
```diff
@@ -0,0 +1,171 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+@Value
+@EqualsAndHashCode(callSuper = false)
+public class UseVarForObject extends Recipe {
+    @Override
+    public String getDisplayName() {
+        return "UseVarForObjects";
```
knutwannheden, line 38:
```suggestion
        return "Use `var` for reference-typed variables";
```

src/main/java/org/openrewrite/java/migrate/lang/var/UseVarForPrimitive.java
```diff
@@ -0,0 +1,182 @@
...
+import static java.lang.String.format;
+import static java.util.Objects.isNull;
+import static java.util.Objects.nonNull;
+
+@Value
+@EqualsAndHashCode(callSuper = false)
+public class UseVarForPrimitive extends Recipe {
+    @Override
+    public String getDisplayName() {
+        return "UseVarForPrimitives";
```
knutwannheden, line 41:
```suggestion
        return "Use `var` for primitive-typed variables";
```

src/main/java/org/openrewrite/java/migrate/lang/var/UseVarForPrimitive.java
```diff
@@ -0,0 +1,182 @@
...
+
+    @Override
+    public String getDescription() {
+        //language=markdown
+        return "Try to apply local variable type inference `var` to primitiv variables where possible." +
+               "This recipe will not touch variable declaration with initializer containing ternary operators.";
+    }
+
+    @Override
+    public List<RecipeExample> getExamples() {
```
knutwannheden: You are typically not expected to override this method. Instead declare a corresponding test and annotate it with `@DocumentExample`.

src/main/java/org/openrewrite/java/migrate/lang/var/UseVarForObject.java
```diff
@@ -0,0 +1,171 @@
...
+        List<RecipeExample> exampleList = new ArrayList<>();
+        exampleList.add(recipeExample);
+        return exampleList;
+    }
+
+    @NotNull
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return Preconditions.check(
+                new HasJavaVersion("10", true),
```
knutwannheden, line 53: We typically use `UsesJavaVersion` instead. Not sure why we have both actually...

src/main/java/org/openrewrite/java/migrate/lang/var/UseVarForObject.java
MBoegers: Removed the `Recipe#getExamples` overwrite and add `@DocumentExample` to

knutwannheden (commented):
Looks very good. It will be interesting to see what turns up when we run this recipe in the SaaS.

src/main/java/org/openrewrite/java/migrate/lang/var/DeclarationCheck.java
```diff
@@ -0,0 +1,179 @@
...
+     * @param vd     variable definition at question
+     * @return true if var is applicable in general
+     */
+    public static boolean isVarApplicable(@NotNull Cursor cursor, @NotNull J.VariableDeclarations vd) {
+        boolean isInsideMethod = DeclarationCheck.isInsideMethod(cursor);
+        boolean isMethodParameter = DeclarationCheck.isMethodParameter(vd, cursor);
+        boolean isInsideInitializer = DeclarationCheck.isInsideInitializer(cursor, 0);
+        boolean isSingleDefinition = DeclarationCheck.isInitializedSingleDefinition(vd);
+
+        return (isInsideMethod || isInsideInitializer) && !isMethodParameter && isSingleDefinition;
```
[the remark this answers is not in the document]
MBoegers: Reorder regarding hot paths. I also improved null handling

knutwannheden:
@MBoegers Looks like you just committed some changes. Some of my comments may not apply anymore...

src/main/java/org/openrewrite/java/migrate/lang/var/DeclarationCheck.java
```diff
@@ -0,0 +1,179 @@
...
+        Expression initializer = vd.getVariables().get(0).getInitializer().unwrap();
+        return initializer instanceof J.Ternary;
+    }
+
+    /**
+     * Determines if a cursor is contained inside a Method declaration without an intermediate Class declaration
+     *
+     * @param cursor value to determine
+     */
+    private static boolean isInsideMethod(@NotNull Cursor cursor) {
```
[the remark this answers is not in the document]
MBoegers: No I was not. It is easier and I came up with a missing test for `ClassDeclaration` inside `MethodDeclaration`

src/main/java/org/openrewrite/java/migrate/lang/var/UseVarForPrimitive.java
MBoegers: Tim also mentioned this, done ;)

src/main/java/org/openrewrite/java/migrate/lang/var/package-info.java
```diff
@@ -0,0 +1,21 @@
+/*
+ * Copyright 2021 the original author or authors.
```
timtebeek, line 2: Looks like you're [not caught up with main](https://github.com/openrewrite/rewrite-migrate-java/commit/2f240932eae2acb366aab61fb89850412fbc23e2) yet.
```suggestion
 * Copyright 2023 the original author or authors.
```

MBoegers: 🫣 yes in very header..

knutwannheden (approved):
Very nice! @timtebeek Anything else you want to add or should we merge it?

knutwannheden:
@MBoegers Thanks again for your contribution!

MBoegers:
Thanks to you both for your help

timtebeek:
Yes thanks a lot! I know it was a long one to get through, with a lot of discovery along the way, but really appreciate the effort!

knutwannheden:
@MBoegers FYI: I integrated two small fixes for the recipe: https://github.com/openrewrite/rewrite-migrate-java/commit/ff4d3dcc5ada9afea668bf11f4575d1f5556bfc3 and https://github.com/openrewrite/rewrite-migrate-java/commit/688ce4768cbd1cb4d85a8218cf3d365c10818f12

PR #219, opened 2023-05-12
Use rewrite-gradle UpdateJavaCompatibility recipe directly

shanman190:
This depends upon https://github.com/openrewrite/rewrite/pull/3238


shanman190:
This needs a bit more work as it presently doesn't correctly handle JVM toolchains for either Gradle or Maven.

In particular, each subproject could be on a different Java version. With the current implementation, it would find the first Java version marker and use it as a basis for the entire repository. This would be correct only if the first marker was less than the desired version. However, if it's equal or newer than the desired version and other projects exist later that are on older Java versions this would behave erroneously.

I think to correct the implementation, we will need to look at all `JavaVersion` markers and if there exists _any_ less than the desired version, we would then execute the migration recipe.

timtebeek (approved):

PR #220, opened 2023-05-12
Format text blocks SQL and JSON (WIP)

MahatmaFatalError:
This is a PoC to get feedback from the maintainers if such a behavior is desirable. 
IMHO it would be useful.

If I get positive feedback, I would try to clean up the code and properly test it.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -25,6 +25,9 @@ dependencies {
     implementation("org.openrewrite:rewrite-maven")
     implementation("org.openrewrite:rewrite-gradle")
     implementation("org.openrewrite.recipe:rewrite-github-actions:$rewriteVersion")
+    implementation("com.github.vertical-blank:sql-formatter:2.0.3")
+    implementation("com.github.jsqlparser:jsqlparser:4.6")
+    implementation("org.json:json:20230227")
 
     runtimeOnly("org.openrewrite:rewrite-java-8")
     runtimeOnly("org.openrewrite:rewrite-java-11")
--- a/src/main/java/org/openrewrite/java/migrate/lang/UseTextBlocks.java
+++ b/src/main/java/org/openrewrite/java/migrate/lang/UseTextBlocks.java
@@ -27,6 +27,15 @@ import org.openrewrite.java.style.TabsAndIndentsStyle;
 import org.openrewrite.java.tree.*;
 import org.openrewrite.marker.Markers;
 
+import com.github.vertical_blank.sqlformatter.SqlFormatter;
+import com.github.vertical_blank.sqlformatter.languages.Dialect;
+import net.sf.jsqlparser.JSQLParserException;
+import net.sf.jsqlparser.parser.CCJSqlParserUtil;
+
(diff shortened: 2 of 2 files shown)


timtebeek:
Oh wow; I'd missed this earlier, but great initiative! I do think this could be a nice addition; I'm not too sure about the inclusion within the existing `UseTextBlocks` recipe though. We try to keep our recipes focused and small, such that they are easier to develop, understand and debug, and don't have any unexpected effects when run. It also helps to limit the amount of configuration options, and how those combine to affect code changes. I propose that instead we create separate recipes for formatting text blocks containing SQL and JSON.

For SQL that might slot in nicely into this repository: [openrewrite/rewrite-sql](https://github.com/openrewrite/rewrite-sql), such that you can also [reuse the SQL detector we have there](https://github.com/openrewrite/rewrite-sql/blob/main/src/main/java/org/openrewrite/sql/SqlDetector.java#L47). In terms of testing we'd have to be sure we cover both the syntax for prepared statements, as well as (insecure) `String.format`, and possibly in the future JEP 430 String Templates.

Then once we have the recipe for SQL ironed out we add a similar one for JSON in [rewrite-json/src/main/java/org/openrewrite/json](https://github.com/openrewrite/rewrite/tree/main/rewrite-json/src/main/java/org/openrewrite/json).

Would you be open to that?

MahatmaFatalError:
Thanks for the great feedback @timtebeek, that is exactly what I was hoping for 👍

I didn't know about [openrewrite/rewrite-sql](https://github.com/openrewrite/rewrite-sql) (and actually I cannot find it either on https://docs.openrewrite.org/) but definitely makes sense what you proposed.

So to outline the steps for building a Recipe in openrewrite/rewrite-sql, implementing `getVisitor()`:

1. Finding text blocks (likely copy and adaption of `isRegularStringLiteral(Expression expr)`)
2. Check if content is SQL (using `SqlDetector`)
3. Format the content (using https://github.com/vertical-blank/sql-formatter)
4. Write back to the text block

Does that make sense?

BTW, I am not very familiar with the package structure, so if you could provide a unit test with an empty class-under-test it would give me guidance and I could fill it with functionality. Otherwise, I will come up with my own proposal.

PS: For the time being, I would leave that PR here open just as a reference.

kunli2:
Thanks for the contribution! I think it’s a nice addition! and also agreed that it’s better to have a separate recipe for this rather than add it to the useTextBlock recipe.

MahatmaFatalError:
> Welcome to open up a draft pull request early where you allow us to push changes as well; that should ensure we can help you if it were needed.

It seems the only way to push is via a fork, but then you can't edit.
image

timtebeek:
> > Welcome to open up a draft pull request early where you allow us to push changes as well; that should ensure we can help you if it were needed.
> 
> It seems the only way to push is via a fork, but then you can't edit. image

When you open a pull request from a fork there is an option to allow maintainers to commit to your fork branch. So you can create a fork first, on that create a branch, and then you'll see prompts to create a pull request after the first commit. Does that help?

PR #223, opened 2023-05-23
Feature/isolate upgrades

nmck257:
implementation of https://github.com/openrewrite/rewrite-spring/pull/351

sibling PR: https://github.com/openrewrite/rewrite-spring/pull/361


timtebeek:
We don't yet support the `hidden` tag in rewrite-docs, and given the feedback received on https://github.com/openrewrite/rewrite-spring/pull/361 I'm not sure whether we'd want to continue with this approach here or there.

Personally I do like that this changeset follows a pattern that we already had for `Java8toJava11`, and seems reasonable to continue with less chance of confusion around duplicated recipes.

Given the feedback on rewrite-spring earlier, I don't feel like merging this here should be my decision alone. @jkschneider OK to weigh in here as well?

nmck257:
Good points. I've flipped this and https://github.com/openrewrite/rewrite-spring/pull/361 to Draft, pending further discussion with Jonathan (and potentially dropping those `hidden` tags).

I do agree, though, that the change is a little more subtle on this repo than over in `rewrite-spring` -- maybe the changes on `rewrite-spring` as-written are more palatable if we change the naming convention from `UpgradeSpringBoot_2_7_Only` to `UpgradeSpringBoot_2_6_To_2_7`, like we have here? Either way, I'm still interested in feedback on the alternative design I put in the comments over there; it has other neat qualities.

@timtebeek @jkschneider

nmck257:
Withdrawing this PR; it'll stay in the history if we ever need it again

PR #232, opened 2023-06-08
Prefer Objects.requireNonNull and StandardCharsets over Guava

timtebeek:
## What's changed?
Added recipes for these two transformations
com.google.common.base.Preconditions.checkNotNull -> Objects.requireNonNull
com.google.common.base.Charsets.UTF_8 -> java.nio.charset.StandardCharsets.UTF_8

## What's your motivation?
As discovered on https://github.com/wiremock/wiremock/issues/2111#issuecomment-1581474504


knutwannheden (commented):
Not quite sure about my comments. But maybe we can add a test to verify.

timtebeek:
> Not quite sure about my comments. But maybe we can add a test to verify.

Thanks! Fixed now.. Also reduced the scope as we can't exactly match one of the replacements, and it's not worth adding a dedicated recipe just yet.

knutwannheden (approved):
👍

PR #233, opened 2023-06-09
Remove finalizer from classes in util.zip 

satvika-eda:
## What's changed?
finalize method has been removed from the ZipFile, Deflater and Inflalter of the util package for which the needed automation has been done as part of the recipe created

## What's your motivation?

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/util/RemoveFinalizerFromZip.java
+++ b/src/main/java/org/openrewrite/java/migrate/util/RemoveFinalizerFromZip.java
@@ -19,13 +19,16 @@ package org.openrewrite.java.migrate.util;
 import lombok.EqualsAndHashCode;
 import lombok.Value;
 import org.openrewrite.ExecutionContext;
+import org.openrewrite.Preconditions;
 import org.openrewrite.Recipe;
+import org.openrewrite.TreeVisitor;
 import org.openrewrite.internal.ListUtils;
 import org.openrewrite.java.JavaIsoVisitor;
+import org.openrewrite.java.search.UsesJavaVersion;
+import org.openrewrite.java.search.UsesType;
 import org.openrewrite.java.tree.J;
 
 import java.util.Objects;
-import java.util.concurrent.atomic.AtomicBoolean;
 
 @Value
 @EqualsAndHashCode(callSuper = true)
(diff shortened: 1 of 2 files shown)


timtebeek:
Thanks a lot for this recipe @satvika-eda ! I'm adding a link here for context: https://bugs.openjdk.org/browse/JDK-8212129
Reviewing now.

timtebeek:
I've gone ahead and added some tests to show some of the potential issues I saw with the current implementation; such as also matching finalize methods called on other classes, and removing the select, which could change side effects. Would you want to work through those issues yourself? 

High over I'd say the recipe is not following the typical pattern I'd expect of recipes; right now you're visiting a  class declaration, and through that looking through the body, and from there methods and statements. That's rather narrow in terms of what it will match. You're also matching classes and methods based on their name, rather than the types that we set on tree elements, and the matchers we provide.

I'd instead roughly expect a pattern where you first use a Precondition argument to check that any file visited extends Deflater/Inflater/ZipFile, and then as a second argument to Precondition add a visitor that overrides visitMethodInvocation, applies a MethodMatcher for `java.util.Object finalize()`, and if the `methodInvocation.getSelect()` matches the class only then return null to remove the method invocation.

I hope all of that is somewhat clear; we really appreciate you providing the initial implementation here, and hope you'll consider making these changes if you feel comfortable doing so.

satvika-eda:
I've worked on fixing the recipe as per your comments. Please review the PR and provide your inputs. I also made sure all the test cases passed. For "removeCallsToSelfFinalize" test, as finalize() would be called from Object class, before and after would be same as that is not a self finalize as per the recipe.

timtebeek:
I've made quite some changes in https://github.com/openrewrite/rewrite-migrate-java/pull/233/commits/c6613d11b0eeb289d637793e0b0c7a70413fec33 to have the recipe better fit in with our expected use of OpenRewrite, making use of the type system for comparisons where we can. We now also cover [this case](https://github.com/openrewrite/rewrite-migrate-java/pull/233/commits/2397bf7fd44e2a25dce186a4cbf6385ef85187a8), by looking at the surrounding class when finalize is called without a selector Hope you agree with the changes, and thank you for your contribution!

timtebeek (approved):

PR #244, opened 2023-06-27
Remove `Runtime.runFinalizersOnExit()` in upgrade to Java 11

satvika-eda:
## What's changed?
Runtime.runFinalizersOnExit() has been removed and replaced in latest JDK versions.

## What's your motivation?

### Checklist


timtebeek:
For context: https://bugs.openjdk.org/browse/JDK-8198250

Some of the same review comments from #246 and #247 apply here, such as adding tags, markdown, removing markers. Could you apply those same changes here?

Thanks again; really neat to see so many pull requests opened at once!

src/test/java/org/openrewrite/java/migrate/lang/ReplaceRuntimeFinalizerTest.java
```diff
@@ -0,0 +1,91 @@
...
+                 class FooBar {
+                      public void test(){
+                          Runtime.runFinalizersOnExit(true);
+                      }
+                  }
+                """,
+              """
+                 class FooBar {
+                      public void test(){
+                          Runtime.getRuntime().addShutdownHook(new Thread());
```
timtebeek, line 47: I think I lack some background on why we inject `addShutdownHook(new Thread())` here; do you have a source to explain why that's needed? Seems strange to have an empty Thread run yet still provide value.

src/main/java/org/openrewrite/java/migrate/lang/ReplaceRuntimeFinalizer.java
```diff
@@ -0,0 +1,87 @@
...
+import java.util.Collections;
+import java.util.Objects;
+import java.util.Set;
+
+@Value
+@EqualsAndHashCode(callSuper = true)
+@Getter
+public class ReplaceRuntimeFinalizer extends Recipe {
+
+    private static final String METHOD_PATTERN = "runFinalizersOnExit";
```
knutwannheden, line 42: It would be more idiomatic to add two `MethodMatcher`s for `java.lang.Runtime runFinalizersOnExit(boolean)` and `java.lang.System runFinalizersOnExit(boolean)`. That has the advantage that they will also match if these methods were imported statically.

src/test/java/org/openrewrite/java/migrate/lang/ReplaceRuntimeFinalizerTest.java
knutwannheden: I would also like some context here. Can't the method call simply be removed?

satvika-eda: Made the changes to recipe to exit the JVM as per https://bugs.openjdk.org/browse/JDK-8198250. Let me know your suggestions.

joanvr (commented):
Looking at the referenced documents, I do not understand why we are replacing the method calls for this replacements, instead of just removing them... Maybe you could provide us more context?

timtebeek:
Thanks again @satvika-eda ! I've taken the liberty to use your PR to clean up some duplication we had around removing existing method calls. All of that's now reusable from properties, which should help if there's future such cases.

PR #246, opened 2023-06-27
remove thread destroy method recipe

satvika-eda:
## What's changed?
To remove Thread.destroy deprecated method

## What's your motivation?

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/lang/RemoveThreadDestroyMethod.java
+++ b/src/main/java/org/openrewrite/java/migrate/lang/RemoveThreadDestroyMethod.java
@@ -27,7 +27,9 @@ import org.openrewrite.java.search.UsesType;
 import org.openrewrite.java.tree.J;
 import org.openrewrite.java.tree.TypeUtils;
 
+import java.util.Collections;
 import java.util.Objects;
+import java.util.Set;
 
 @Value
 @EqualsAndHashCode(callSuper = true)
@@ -37,12 +39,17 @@ public class RemoveThreadDestroyMethod extends Recipe {
 
     @Override
     public String getDisplayName() {
-        return "Remove deprecated Thread.destroy() statement";
+        return "Remove deprecated `Thread.destroy()`";
     }
 
     @Override
     public String getDescription() {
-        return "Remove deprecated invocations of Thread.destroy() which have no alternatives needed.";
+        return "Remove deprecated invocations of `Thread.destroy()` which have no alternatives needed.";
+    }
+
+    @Override
+    public Set<String> getTags() {
+        return Collections.singleton("JDK-8204260");
     }
 
     @Override
@@ -53,7 +60,11 @@ public class RemoveThreadDestroyMethod extends Recipe {
                     @Override
                     public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, ExecutionContext ctx) {
                         J.MethodInvocation mi = super.visitMethodInvocation(method, ctx);
-                        return (Objects.nonNull(mi.getSelect()) && TypeUtils.isAssignableTo(JAVA_LANG_THREAD, mi.getSelect().getType()) && mi.getSimpleName().equals("destroy")) ? null : mi;
+                        if (Objects.nonNull(mi.getSelect())
+                                && TypeUtils.isAssignableTo(JAVA_LANG_THREAD, mi.getSelect().getType())
+                                && mi.getSimpleName().equals("destroy"))
+                            return null;
+                        return mi;
                     }
                 });
     }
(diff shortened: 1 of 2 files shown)


timtebeek:
For context: https://bugs.openjdk.org/browse/JDK-8204260

src/main/java/org/openrewrite/java/migrate/lang/RemoveThreadDestroyMethod.java
```diff
@@ -0,0 +1,61 @@
...
+
+@Value
+@EqualsAndHashCode(callSuper = true)
+public class RemoveThreadDestroyMethod extends Recipe {
+
+    public static final String JAVA_LANG_THREAD = "java.lang.Thread";
+
+    @Override
+    public String getDisplayName() {
+        return "Remove deprecated Thread.destroy() statement";
```
timtebeek, line 40: You can use Markdown around code snippets here and in the description such that it'll look slightly better in the documentation.
```suggestion
        return "Remove deprecated `Thread.destroy()`";
```

src/main/java/org/openrewrite/java/migrate/lang/RemoveThreadDestroyMethod.java
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    @Override
+    public String getDescription() {
+        return "Remove deprecated invocations of Thread.destroy() which have no alternatives needed.";
```
timtebeek, line 45:
```suggestion
        return "Remove deprecated invocations of `Thread.destroy()` which have no alternatives needed.";
```

src/test/java/org/openrewrite/java/migrate/lang/RemoveThreadDestroyMethodTest.java
```diff
@@ -0,0 +1,102 @@
...
+import org.openrewrite.test.RewriteTest;
+
+import static org.openrewrite.java.Assertions.java;
+import static org.openrewrite.java.Assertions.javaVersion;
+
+class RemoveThreadDestroyMethodTest implements RewriteTest {
+
+    @Override
+    public void defaults(RecipeSpec spec) {
+        spec.recipe(new RemoveThreadDestroyMethod()).allSources(s -> s.markers(javaVersion(8)));
```
timtebeek, line 30: I don't there's anything specific around the Java version of the sources here right? Maybe best to then not explicitly set any markers, as it could cause confusion
```suggestion
        spec.recipe(new RemoveThreadDestroyMethod());
```

timtebeek (commented):
Thanks a lot for implementing this one! Some of the same concerns regarding the use of the type system as in #247 apply here, but I don't think there's anything much we can do about that. Some smaller suggestions that when applied can lead to a merge.

src/main/java/org/openrewrite/java/migrate/lang/RemoveThreadDestroyMethod.java
```diff
@@ -0,0 +1,61 @@
...
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+
+        return Preconditions.check(new UsesType<>(JAVA_LANG_THREAD, false),
+                new JavaIsoVisitor<ExecutionContext>() {
+                    @Override
+                    public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, ExecutionContext ctx) {
+                        J.MethodInvocation mi = super.visitMethodInvocation(method, ctx);
+                        return (Objects.nonNull(mi.getSelect()) && TypeUtils.isAssignableTo(JAVA_LANG_THREAD, mi.getSelect().getType()) && mi.getSimpleName().equals("destroy")) ? null : mi;
```
timtebeek, line 56: Suggest to add some newlines here for each of the conditions that we check; and split the ternary into an if statement such that it's clearer that we only make a change (`return null`) when those conditions match.
```suggestion
                        J.MethodInvocation mi = super.visitMethodInvocation(method, ctx);
                        if (Objects.nonNull(mi.getSelect())
                            && TypeUtils.isAssignableTo(JAVA_LANG_THREAD, mi.getSelect().getType())
                            && mi.getSimpleName().equals("destroy")) {
                            return null;
                        }
                        return mi;
```

src/test/java/org/openrewrite/java/migrate/lang/RemoveThreadDestroyMethodTest.java
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    @Test
+    void noChangeWithoutThreadDestroy() {
+        //language=java
+        rewriteRun(
+          java(
+            """
+              import java.lang.*;
+
```
timtebeek, line 89:
```suggestion
```

satvika-eda:
Thank you for your detailed inputs. I've made the necessary changes. Please review them.

knutwannheden:
@satvika-eda Thanks for the contribution!

PR #247, opened 2023-06-27
Removing deprecated Runtime trace methods in move to Java 17

satvika-eda:
## What's changed?
Remove deprecated invocations of Runtime.traceInstructions() and Runtime.traceMethodCalls() which have no alternatives needed.

## What's your motivation?

### Checklist


timtebeek:
For context: this was removed in Java 13
https://bugs.openjdk.org/browse/JDK-8225330
As per https://javaalmanac.io/jdk/17/apidiff/11/

src/test/java/org/openrewrite/java/migrate/lang/RemoveDeprecatedRuntimeTest.java
```diff
@@ -0,0 +1,112 @@
...
+        //language=java
+        rewriteRun(
+          version(
+            java(
+              """
+                import java.lang.*;
+                                
+                class FooBar{
+                    public void test() {
+                        Runtime.traceInstructions();
```
timtebeek, line 44: The methods you're replacing aren't static on Runtime; they require an instance, which I've passed into the method.
```suggestion
                    void test(Runtime r) {
                        r.traceInstructions();
```

src/test/java/org/openrewrite/java/migrate/lang/RemoveDeprecatedRuntimeTest.java
```diff
@@ -0,0 +1,112 @@
...
+                    }
+                }
+                 """,
+              """
+                import java.lang.*;
+                                
+                class FooBar{
+                    public void test() {
```
timtebeek, line 52:
```suggestion
                    void test(Runtime r) {
```

src/test/java/org/openrewrite/java/migrate/lang/RemoveDeprecatedRuntimeTest.java
```diff
@@ -0,0 +1,112 @@
...
+        //language=java
+        rewriteRun(
+          version(
+            java(
+              """
+                import java.lang.*;
+                                
+                class FooBar{
+                    public void test() {
+                        Runtime.traceMethodCalls();
```
timtebeek, line 74:
```suggestion
                    void test(Runtime r) {
                        r.traceMethodCalls();
```

src/test/java/org/openrewrite/java/migrate/lang/RemoveDeprecatedRuntimeTest.java
```diff
@@ -0,0 +1,112 @@
...
+                    }
+                }
+                 """,
+              """
+                import java.lang.*;
+                                
+                class FooBar{
+                    public void test() {
```
timtebeek, line 82:
```suggestion
                    void test(Runtime r) {
```

src/test/java/org/openrewrite/java/migrate/lang/RemoveDeprecatedRuntimeTest.java
```diff
@@ -0,0 +1,112 @@
...
+        //language=java
+        rewriteRun(
+          version(
+            java(
+              """
+                import java.lang.*;
+                                
+                class FooBar{
+                    public void test() {
+                        Runtime.getRuntime();
```
timtebeek, line 103: Probably best to also call an instance method here.
```suggestion
                    void test(Runtime r) {
                        r.gc();
```

src/main/java/org/openrewrite/java/migrate/lang/RemoveDeprecatedRuntime.java
```diff
@@ -0,0 +1,63 @@
...
+
+    @Override
+    public String getDisplayName() {
+        return "Remove deprecated statements from Runtime module";
+    }
+
+    @Override
+    public String getDescription() {
+        return "Remove deprecated invocations of Runtime.traceInstructions() and Runtime.traceMethodCalls() which have no alternatives needed.";
+    }
```
timtebeek, line 44:
```suggestion
    }

    @Override
    public Set<String> getTags() {
        return Collections.singleton("JDK-8225330");
    }
```

src/main/java/org/openrewrite/java/migrate/lang/RemoveDeprecatedRuntime.java
```diff
@@ -0,0 +1,63 @@
...
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+
+
+        return Preconditions.check(new UsesType<>(JAVA_LANG_RUNTIME, false),
```
timtebeek, line 50: I had hoped to instead here use:
```java
        return Preconditions.check(Preconditions.or(new UsesMethod<>(TRACE_INSTRUCTIONS), new UsesMethod<>(TRACE_METHOD_CALLS)),
```
But that trips up the tests since those use Java 17 by default, which lacks the corresponding methods, and forcing `Java8Parser` or `Java11Parser` there so far gives the same result, as well as trying to override the Runtime class in a `java(...)` SourceSpec block. Any suggestions @knutwannheden ?

timtebeek: With the following fields at the top:
```java
    private static final MethodMatcher TRACE_INSTRUCTIONS = new MethodMatcher("java.lang.Runtime traceInstructions(boolean)");
    private static final MethodMatcher TRACE_METHOD_CALLS = new MethodMatcher("java.lang.Runtime traceMethodCalls(boolean)");
```

src/main/java/org/openrewrite/java/migrate/lang/RemoveDeprecatedRuntime.java
```diff
@@ -0,0 +1,63 @@
...
+                new JavaIsoVisitor<ExecutionContext>() {
+                    @Override
+                    public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, ExecutionContext ctx) {
+                        J.MethodInvocation mi = super.visitMethodInvocation(method, ctx);
+                        return (Objects.nonNull(mi.getSelect()) && TypeUtils.isAssignableTo(JAVA_LANG_RUNTIME, mi.getSelect().getType())
+                                && (TRACE_INSTRUCTIONS.equals(mi.getSimpleName()) || TRACE_METHOD_CALLS.equals(mi.getSimpleName()))
+                                ? null : mi);
+
```
timtebeek, line 58: Ideally this would just be 
```java
                        if (TRACE_INSTRUCTIONS.matches(mi) || TRACE_METHOD_CALLS.matches(mi)) {
                            return null;
                        }
                        return mi;
```

timtebeek (commented):
Some suggestions already, after checking out your branch locally and running through the changes.

In general I can say the code snippets in the tests were not yet valid for any version of Java; you can fix that through my code suggestions that you can apply all at once from this page: https://github.com/openrewrite/rewrite-migrate-java/pull/247/files

The recipe does not make the best use of the type system, possibly due to limitations with replacing methods formerly in the JDK. Not sure if there's a better way, as I struggled with forcing the test and replacements to use older versions of Java, and still did not get the method type to show. We can keep things as they are for now, as there's unlikely to arise any conflicts given that these methods are already on `java.lang.Runtime`.

src/main/java/org/openrewrite/java/migrate/lang/RemoveDeprecatedRuntime.java
timtebeek: If that approach fails I'd prefer to use an if statement over the ternary expression here; as it's slightly less obvious what exactly we're doing now.

src/main/java/org/openrewrite/java/migrate/lang/RemoveDeprecatedRuntime.java
knutwannheden: That indeed sounds tricky... It may be possible to set the `-release` compiler option in `ReloadableJava17Parser` and we may even want to add some of these options to the `JavaParser.Builder` API. I am not suggesting we do this in the scope of this PR, but this may allow us to test this properly in the future.

For now we may have to live with that we cannot properly test this.

PR #258, opened 2023-07-12
Fix/no guava immutable recipes

### LINKED ISSUE #256: Guava ImmutableMap recipe fails when variable type is ImmutableMap
I am using:

- OpenRewrite v8.1.8
- Maven plugin v5.2.6
- rewrite-migrate-java v2.0.6

I am using the Maven plugin and my project is a single module project.
```xml
<plugin>
    <version>5.2.6</version>
    <configuration>
        <checkstyleDetectionEnabled>false</checkstyleDetectionEnabled>
    </configuration>
    <dependencies>
        <dependency>
            <version>2.0.6</version>
        </dependency>
    </dependencies>
</plugin>
```

```java
package org.example;

import com.google.common.collect.ImmutableMap;

public class ImmutableCollectionDemo {
    private void poorVariableType() {
        ImmutableMap<String, String> map = ImmutableMap.of("alpha", "beta");
    }
}
```
This code is a little unusual give that the type of `map` is `ImmutableMap`, rather than `Map`.

```java
package org.example;

import java.util.Map;

public class ImmutableCollectionDemo {
    private void poorVariableType() {
        Map<String, String> map = Map.of("alpha", "beta");
    }
}
```
Both uses of `ImmutableMap` converted for consistency.
(Alternatively, we could play it safe and not change anything in this instance.)

## What did you see instead?
```java
package org.example;

import java.util.Map;

import com.google.common.collect.ImmutableMap;

public class ImmutableCollectionDemo {
    private void poorVariableType() {
        ImmutableMap<String, String> map = Map.of("alpha", "beta");
    }
}
```
Only the second use of `ImmutableMap` is converted leading to a type mis-match and compilation error.

The generated code does not compile due to the mismatch of types:
```
no instance(s) of type variable(s) K, V exist so that Map<K, V> conforms to ImmutableMap<String, String>
```

## Any other information

I tested the equivalent for `ImmutableList` and `ImmutableSet` and no change to the code was made by either recipe. So perhaps we should be consistent for `ImmutableMap`.

Yes, if possible. Not sure how useful I'll be as I'm not familiar with your code base, nor have I written my own recipe (yet).

joanvr:
## What's changed?
First, we removed a previous change that allowed Map.of to be assigned to ImmutableMap (this was done to allow nested maps), but was causing uncompilable code.

Added some tests to validate that we do not produce changes in those scenarios for all three collections (List, Set, Map).

Also, since all three recipes (List, Set, Map), have the same duplicated code that needs to be synched, and already had slightly different versions, I've decided to extract it in a parent class, to avoid future inconsistence problems like it was happening right now

## What's your motivation?
Fixes #256 

### Checklist
- [x] I've updated the documentation (if applicable)

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/guava/AbstractNoGuavaImmutableOf.java
+++ b/src/main/java/org/openrewrite/java/migrate/guava/AbstractNoGuavaImmutableOf.java
@@ -31,7 +31,7 @@ import java.time.Duration;
 import java.util.Objects;
 import java.util.stream.Collectors;
 
-public abstract class AbstractNoGuavaImmutableOf extends Recipe {
+abstract class AbstractNoGuavaImmutableOf extends Recipe {
 
     private final String guavaType;
     private final String javaType;
--- a/src/main/java/org/openrewrite/java/migrate/guava/AbstractNoGuavaImmutableOf.java
+++ b/src/main/java/org/openrewrite/java/migrate/guava/AbstractNoGuavaImmutableOf.java
@@ -33,23 +33,25 @@ import java.util.stream.Collectors;
 
 public abstract class AbstractNoGuavaImmutableOf extends Recipe {
 
-    private transient final MethodMatcher IMMUTABLE_MATCHER = new MethodMatcher(getGuavaType()+" of(..)");
+    private transient final MethodMatcher IMMUTABLE_MATCHER = new MethodMatcher(getGuavaType() + " of(..)");
 
     abstract String getGuavaType();
+
     abstract String getJavaType();
+
     private String getShortType(String fullyQualifiedType) {
         return fullyQualifiedType.substring(getJavaType().lastIndexOf(".") + 1);
     }
 
     @Override
     public String getDisplayName() {
-        return "Prefer `"+getShortType(getJavaType())+".of(..)` in Java 9 or higher";
+        return "Prefer `" + getShortType(getJavaType()) + ".of(..)` in Java 9 or higher";
     }
 
     @Override
     public String getDescription() {
         String name = getGuavaType().substring(getGuavaType().lastIndexOf("."));
-        return "Replaces `"+getShortType(getGuavaType())+".of(..)` if the returned type is immediately down-cast.";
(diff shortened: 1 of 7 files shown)


timtebeek (approved):
Looks good to me; thanks for fixing the issue and cleaning up the code at the same time.

src/main/java/org/openrewrite/java/migrate/guava/AbstractNoGuavaImmutableOf.java
```diff
@@ -0,0 +1,182 @@
...
+import org.openrewrite.java.search.UsesType;
+import org.openrewrite.java.tree.*;
+
+import java.time.Duration;
+import java.util.Objects;
+import java.util.stream.Collectors;
+
+public abstract class AbstractNoGuavaImmutableOf extends Recipe {
+
+    private transient final MethodMatcher IMMUTABLE_MATCHER = new MethodMatcher(getGuavaType() + " of(..)");
```
timtebeek, line 36: What's the background behind adding the `transient` here? Is it because of the call to `getGuavaType` when combined with `final`?

I'm [reading up a bit](https://stackoverflow.com/questions/2968876/final-transient-fields-and-serialization) and found some conflicting reports of using transient, final, and non-trivial objects together when serialization might be involved. 

If you think it'll work fine that fine by me; but wanted to ask both to learn and get a second opinion.

timtebeek: Maybe it would make sense to define a constructor in `AbstractNoGuavaImmutableOf` that takes in two arguments, and reduce `NoGuavaImmutableMapOf` to just calling that constructor with those two arguments?

timtebeek: pushed just now in https://github.com/openrewrite/rewrite-migrate-java/pull/258/commits/20564abd8aba709efd30827c7913abfc218e7b95 
hope you agree; otherwise open for a revert and further discussion

timtebeek (approved):

PR #263, opened 2023-07-19
Fix errors for Arrays and inception bounds

MBoegers:
## What's changed?
Fixed two additional edge cases in `UseVarForGenericsConstructors` found with run [moderne.io/results/CuzcK](https://app.moderne.io/results/CuzcK?filter=eyJpdGVtcyI6W3siZmllbGQiOiJzdGF0dXMiLCJvcGVyYXRvciI6Im5vdCIsImlkIjo0Nzc3MiwidmFsdWUiOiJGSU5JU0hFRCJ9XSwibG9naWNPcGVyYXRvciI6ImFuZCIsInF1aWNrRmlsdGVyVmFsdWVzIjpbXSwicXVpY2tGaWx0ZXJMb2dpY09wZXJhdG9yIjoiYW5kIn0%253D)

## What's your motivation?
Fix edgecases in UseVarForGenericsConstructors.

It breaks moderne.io runs, so no.

### Checklist
- [x] ~I've updated the documentation (if applicable)~

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/lang/var/UseVarForGenericMethodInvocations.java
+++ b/src/main/java/org/openrewrite/java/migrate/lang/var/UseVarForGenericMethodInvocations.java
@@ -17,7 +17,6 @@ package org.openrewrite.java.migrate.lang.var;
 
 import java.util.ArrayList;
 import java.util.List;
-import java.util.stream.Collectors;
 
 import org.openrewrite.*;
 import org.openrewrite.java.JavaIsoVisitor;
@@ -72,13 +71,22 @@ public class UseVarForGenericMethodInvocations extends Recipe {
 
             //if no type paramters are present and no arguments we assume the type is hard to determine a needs manual action
             boolean hasNoTypeParams = ((J.MethodInvocation) initializer).getTypeParameters() == null;
-            boolean argumentsEmpty = ((J.MethodInvocation) initializer).getArguments().stream().allMatch(p -> p instanceof J.Empty);
+            boolean argumentsEmpty = allArgumentsEmpty((J.MethodInvocation) initializer);
             if (hasNoTypeParams && argumentsEmpty) return vd;
 
             // mark imports for removal if unused
             if (vd.getType() instanceof JavaType.FullyQualified) maybeRemoveImport((JavaType.FullyQualified) vd.getType());
 
             return transformToVar(vd, new ArrayList<>(), new ArrayList<>());
+        }
+
+        private static boolean allArgumentsEmpty(J.MethodInvocation invocation){
+            for (Expression argument : invocation.getArguments()) {
+                if (!(argument instanceof J.Empty)) {
+                    return false;
+                }
+            }
+            return true;
         }
 
         private J.VariableDeclarations transformToVar(J.VariableDeclarations vd, List<JavaType> leftTypes, List<JavaType> rightTypes) {
@@ -88,10 +96,10 @@ public class UseVarForGenericMethodInvocations extends Recipe {
             // if left is defined but not right, copy types to initializer
             if(rightTypes.isEmpty() && !leftTypes.isEmpty()) {
                 // we need to switch type infos from left to right here
-                List<Expression> typeArgument = leftTypes.stream()
-                        .map(t ->
-                                new J.Identifier(Tree.randomId(), Space.EMPTY, Markers.EMPTY, ((JavaType.Class)t).getClassName(), t, null))
-                        .collect(Collectors.toList());
+                List<Expression> typeArgument = new ArrayList<>();
+                for (JavaType t : leftTypes) {
+                    typeArgument.add(new J.Identifier(Tree.randomId(), Space.EMPTY, Markers.EMPTY, ((JavaType.Class) t).getClassName(), t, null));
+                }
                 J.ParameterizedType typedInitializerClazz = ((J.ParameterizedType) ((J.NewClass) initializer).getClazz()).withTypeParameters(typeArgument);
                 initializer = ((J.NewClass) initializer).withClazz(typedInitializerClazz);
             }
--- a/src/main/java/org/openrewrite/java/migrate/lang/var/UseVarForGenericsConstructors.java
+++ b/src/main/java/org/openrewrite/java/migrate/lang/var/UseVarForGenericsConstructors.java
@@ -19,8 +19,10 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
 import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 
 import org.openrewrite.*;
+import org.openrewrite.internal.StringUtils;
 import org.openrewrite.internal.lang.Nullable;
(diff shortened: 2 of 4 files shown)


src/test/java/org/openrewrite/java/migrate/lang/var/UseVarForGenericsConstructorsTest.java
```diff
@@ -357,6 +357,63 @@ void m() {
             );
         }
 
+        //see https://app.moderne.io/results/CuzcK/details/eyJfX3R5cGVuYW1lIjoiR2l0SHViUmVwb3NpdG9yeSIsIm9yaWdpbiI6ImdpdGh1Yi5jb20iLCJwYXRoIjoic3ByaW5nLXByb2plY3RzL3NwcmluZy12YXVsdCIsImJyYW5jaCI6Im1haW4ifQ==?referrer=%2Fresults%2FCuzcK%3Ffilter%3DeyJpdGVtcyI6W3siZmllbGQiOiJzdGF0dXMiLCJvcGVyYXRvciI6Im5vdCIsImlkIjo0Nzc3MiwidmFsdWUiOiJGSU5JU0hFRCJ9XSwibG9naWNPcGVyYXRvciI6ImFuZCIsInF1aWNrRmlsdGVyVmFsdWVzIjpbXSwicXVpY2tGaWx0ZXJMb2dpY09wZXJhdG9yIjoiYW5kIn0%25253D
```
timtebeek, line 360: These URLs only last as long as the worker that holds them, so while I appreciate the content for review, it's likely to break soon, so let's remove it from the code.
```suggestion
```

src/main/java/org/openrewrite/java/migrate/lang/var/UseVarForGenericsConstructors.java
```diff
@@ -86,6 +86,18 @@ public J.VariableDeclarations visitVariableDeclarations(J.VariableDeclarations v
             return transformToVar(vd, leftTypes, rightTypes);
         }
 
+        private static boolean hasBounds(JavaType type) {
+            if (type instanceof JavaType.Parameterized) {
+                return ((JavaType.Parameterized) type).getTypeParameters().stream()
+                        .map(UseVarForGenericsConstructorsVisitor::hasBounds)
+                        .reduce(false, Boolean::logicalOr);
```
timtebeek, line 93: In terms of performance we have preference for simpler constructs without Streams, as there's less objects created while the recipe runs, which at scale really starts to matter. Could this be simplified into a for loop with an early return?

src/main/java/org/openrewrite/java/migrate/lang/var/UseVarForGenericsConstructors.java
```diff
@@ -169,10 +181,22 @@ private J.VariableDeclarations transformToVar(J.VariableDeclarations vd, List<Ja
...
+            }
+            if (type instanceof JavaType.Class) {
                 String className = ((JavaType.Class) type).getClassName();
                 return new J.Identifier(Tree.randomId(), Space.EMPTY, Markers.EMPTY, className, type, null);
             }
+            if (type instanceof JavaType.Array){
+                List<JRightPadded<Space>> dimensions = IntStream.of(StringUtils.countOccurrences(type.toString(), "[]"))
+                        .mapToObj(i -> Space.EMPTY)
+                        .map(JRightPadded::build)
+                        .collect(Collectors.toList());
```
timtebeek, line 196: Nice that you thought about multi dimensional arrays!

timtebeek (commented):
Love that you ran the recipe at scale through the platform to flush out any imperfections; really appreciated!

One small request, that I can also take on if you'd rather work on something else. :)

src/main/java/org/openrewrite/java/migrate/lang/var/UseVarForGenericsConstructors.java
MBoegers: does this preference also apply to snippets like this?
```java
((J.ParameterizedType) clazz).getTypeParameters().stream()
                                .map(Expression::getType)
                                .filter(Objects::nonNull)
                                .collect(Collectors.toList());
```

timtebeek: It depends a bit on where in the flow we find that snippet; I'd commented on this one in particular since it's in the path of determining whether or not to make changes, which will be called repeatedly over time, since we recommend people "continuously migrate", by running idempotent recipes periodically.

If a snippet is only in the path where we are making code changes, then that's assumed to run only once, and we're a little more tolerant to readability over performance. With that background I trust that you can use your own judgement as to which is the case when :)

src/main/java/org/openrewrite/java/migrate/lang/var/UseVarForGenericsConstructors.java
MBoegers: Thanks! But I consider counting "[]" as a hack. I wasn't able to find something like JavaType.Array#getDimensions. There is no way in the API, yet?

timtebeek: Not that I know; I see you've logged https://github.com/openrewrite/rewrite/issues/3425 ; seems like a good idea to me unless we both missed something.

src/main/java/org/openrewrite/java/migrate/lang/var/UseVarForGenericMethodInvocations.java
```diff
@@ -81,17 +80,26 @@ public J.VariableDeclarations visitVariableDeclarations(J.VariableDeclarations v
...
             if(rightTypes.isEmpty() && !leftTypes.isEmpty()) {
                 // we need to switch type infos from left to right here
-                List<Expression> typeArgument = leftTypes.stream()
-                        .map(t ->
-                                new J.Identifier(Tree.randomId(), Space.EMPTY, Markers.EMPTY, ((JavaType.Class)t).getClassName(), t, null))
-                        .collect(Collectors.toList());
+                List<Expression> typeArgument = new ArrayList<>();
+                for (JavaType t : leftTypes) {
+                    typeArgument.add(new J.Identifier(Tree.randomId(), Space.EMPTY, Markers.EMPTY, ((JavaType.Class) t).getClassName(), t, null));
+                }
```
timtebeek, line 102: I forgot to mention the use of `ListUtils`; that stops us creating a new list if it's not necessary, so a neat addition here (and conforms to other recipes).
```suggestion
                List<Expression> typeArgument = ListUtils.map(leftTypes, t -> {
                    typeArgument.add(new J.Identifier(Tree.randomId(), Space.EMPTY, Markers.EMPTY, ((JavaType.Class) t).getClassName(), t, null));
                });
```

timtebeek: Nevermind; read this wrong!

timtebeek (approved):

timtebeek:
Thanks again! Really appreciate how involved you are with seeing this one through.

PR #264, opened 2023-07-22
fix: added emptyList for annotations akin to the adaption made to cor…

SimonVerhoeven:
## What's changed?

Added the emptyList() to J.Field akin to the changes made to core after commit #df5478b8592d90bba65923878b08ae0a412fdbc9

## What's your motivation?

Working build

No

### Checklist
- [x] I've updated the documentation (if applicable)


timtebeek (approved):

timtebeek:
Much appreciated, thanks! Indeed had to restore & deprecate the old constructors, and now we have to go over all repositories to add those arguments, so glad to have some help with that!

PR #270, opened 2023-08-10
Add Apache Commons StringUtils refaster template replacements

timtebeek:


timtebeek:
I've had to put a symlink in `src/main`
```bash
cd src/main
ln -s ../../build/generated/sources/annotationProcessor/java/main/ generated
```

src/test/java/org/openrewrite/java/migrate/apache/commons/lang/ApacheCommonsStringUtilsTest.java
```diff
@@ -0,0 +1,56 @@
...
+          //language=java
+          java("""
+            import org.apache.commons.lang3.StringUtils;
+                          
+            class Foo {
+                String in = "foo";
+                String out = StringUtils.defaultString(in);
+            }
+            """, """
+            import java.util.Objects;
```
timtebeek, line 45: This import isn't added presently; Instead the old import is retained. Any quick thoughts there @knutwannheden ?

knutwannheden: Possibly we only handle imports for the template and not for the LST. I need to take a look at the code. Can you tell from the generated class?

timtebeek: I see import in the generated JavaTemplate before/after classes
```java
public class DefaultStringRecipe$1_after {
    public static JavaTemplate.Builder getTemplate(JavaVisitor<?> visitor) {
        return JavaTemplate
                .builder("Objects.toString(#{any(java.lang.String)})")
                .imports("java.util.Objects");
    }
}
```


I did not see imports added or removed in 
```java
return new JavaVisitor<ExecutionContext>() {
    final JavaTemplate before = JavaTemplate.compile(this, "before", (JavaTemplate.F1<?, ?>) (String s) -> defaultString(s)).build();
    final JavaTemplate after = JavaTemplate.compile(this, "after", (JavaTemplate.F1<?, ?>) (String s) -> Objects.toString(s)).build();

    @Override
    public J visitMethodInvocation(J.MethodInvocation elem, ExecutionContext ctx) {
        JavaTemplate.Matcher matcher;
        if ((matcher = before.matcher(getCursor())).find()) {
            return after.apply(getCursor(), elem.getCoordinates().replace(), matcher.parameter(0));
        }
        return super.visitMethodInvocation(elem, ctx);
    }

};
```


It's not pretty, but this gets the job done: https://github.com/openrewrite/rewrite-templating/pull/10

knutwannheden: I believe my plan was to use `ShortenFullyQualifiedTypeReferences` and another recipe to remove unused imports after applying the template. But I never got around to it.

AlekSimpson:
I've tried getting another recipe for `isBlank()` to work to understand how these refaster recipes work. Is the recipe file auto generated or are you supposed to create manually? For example, how is `DefaultStringRecipe` created?

timtebeek:
Continued in #271

PR #271, opened 2023-08-11
Add Apache Commons StringUtils refaster template replacements

### LINKED ISSUE #267: StringUtils.defaultString ->  Objects.toString
```java

    package org.apache.commons.lang3; 


    /**
     * Returns either the given String, or if the String is
     * {@code null}, {@code nullDefault}.
     *
     * <pre>
     * StringUtils.defaultString(null, "NULL")  = "NULL"
     * StringUtils.defaultString("", "NULL")    = ""
     * StringUtils.defaultString("bat", "NULL") = "bat"
     * </pre>
     *
     * @see Objects#toString(Object, String)
     * @see String#valueOf(Object)
     * @param str  the String to check, may be null
     * @param nullDefault  the default String to return
     *  if the input is {@code null}, may be null
     * @return the passed in String, or the default if it was {@code null}
     * @deprecated Use {@link Objects#toString(Object, String)}
     */
    @Deprecated
    public static String defaultString(final String str, final String nullDefault) {
        return Objects.toString(str, nullDefault);
    }



```

AlekSimpson:
## What's changed?
Refaster rule recipes to replace Apache Commons StringUtils methods with JDK internals where possible.

## What's your motivation?
Fixes #267

@timtebeek 

### Checklist


src/main/java/org/openrewrite/java/migrate/apache/commons/lang/ApacheCommonsStringUtils.java
```diff
@@ -0,0 +1,357 @@
...
+        @AfterTemplate
+        String after(String s) {
+            return Objects.toString(s);
+        }
+    }
+
+    public static class Equals {
+        @BeforeTemplate
+        boolean before(String s) {
+            return StringUtils.equals(s, "string");
```
knutwannheden, line 53: Here you should change the signature so that you have `(String s1, String s2)` and then use those. Otherwise the recipe will only ever do anything if you actually call the method with the `"string"` literal you specify here. I think this same comment also applies to some of your other Refaster templates.

Unfortunately our Refaster support doesn't yet allow checking for the reverse order.

src/main/java/org/openrewrite/java/migrate/apache/commons/lang/ApacheCommonsStringUtils.java
```diff
@@ -0,0 +1,357 @@
...
+        }
+
+        @AfterTemplate
+        boolean after(String s) {
+            return s.equals("string");
```
knutwannheden, line 58: Can you verify that this is corrects with respect to nullability? Otherwise I suggest to use `Objects.equal()` here.

src/main/java/org/openrewrite/java/migrate/apache/commons/lang/ApacheCommonsStringUtils.java
```diff
@@ -0,0 +1,357 @@
...
+        @BeforeTemplate
+        boolean before(String s) {
+            return StringUtils.isAlphanumericSpace("^[a-zA-Z0-9\\s]*$");
+        }
+    }
+
+    public static class Left {
+        @BeforeTemplate
+        String before(String s) {
+            return StringUtils.left(s, 5);
```
knutwannheden, line 208: Also here for `5` instead introduce a second `int l` param.

src/main/java/org/openrewrite/java/migrate/apache/commons/lang/ApacheCommonsStringUtils.java
```diff
@@ -0,0 +1,357 @@
...
+        }
+
+        @AfterTemplate
+        String after(String s) {
+            return s.substring(0, 5);
```
knutwannheden, line 213: Haven't checked if we need to deal with `null` here. Should be checked for all templates, I think.

knutwannheden (commented):
I have added a first set of review comments.

src/main/java/org/openrewrite/java/migrate/apache/commons/lang/ApacheCommonsStringUtils.java
```diff
@@ -0,0 +1,361 @@
...
+
+    public static class IsEmpty {
+        @BeforeTemplate
+        boolean before(String s) {
+            return StringUtils.isEmpty(s);
+        }
+
+        @AfterTemplate
+        boolean after(String s) {
+            return s.isEmpty();
```
timtebeek, line 70: I think this requires a null check too
```suggestion
            return s ==null || s.isEmpty();
```

AlekSimpson:
Ok the latest commit has all the changes made based off of the feedback given.

AlekSimpson:
I am still seeing errors when I run the tests. It tells me that `an enclosing instance that contains org.openrewrite.java.migrate.apache.commons.lang.ApacheCommonsStringUtilsRecipes.DefaultStringRecipe` is required. I'm not to sure what that error is saying to be honest.

knutwannheden:
> I am still seeing errors when I run the tests. It tells me that `an enclosing instance that contains org.openrewrite.java.migrate.apache.commons.lang.ApacheCommonsStringUtilsRecipes.DefaultStringRecipe` is required. I'm not to sure what that error is saying to be honest.

You probably need to declare the Refaster template classes as static.

AlekSimpson:
I have made them all static. I think the issue could be because I have the a local version of `rewrite-templating` to get some fixes earlier so I could make some progress while I waited for the fixes to make it to the main branch.

knutwannheden:
Meanwhile quite a few fixes have been integrated. I suggest you try with the latest integration build.

src/main/java/org/openrewrite/java/migrate/apache/commons/lang/ApacheCommonsStringUtils.java
```diff
@@ -0,0 +1,365 @@
...
+            return StringUtils.isEmpty(s);
+        }
+
+        @AfterTemplate
+        boolean after(String s) {
+            return s == null || s.isEmpty();
+        }
+    }
+
+    public static class Replace {
```
timtebeek, line 76: Let's sort all nested classes here, such that it's easier to find them. Your IDEA likely has a sort-members option.

src/main/java/org/openrewrite/java/migrate/apache/commons/lang/ApacheCommonsStringUtils.java
```diff
@@ -0,0 +1,365 @@
...
+    //    @BeforeTemplate
+    //    String before(String s, int l) {
+    //        return StringUtils.rightPad(s, l);
+    //    }
+
+    //    @AfterTemplate
+    //    String after(String s, int l) {
+    //        return String.format("%" + (-l) + "s", s);
+    //    }
+    //}
```
timtebeek, line 147: I think we can try these again after the recent fixes.

src/test/java/org/openrewrite/java/migrate/apache/commons/lang/ApacheCommonsStringUtilsTest.java
```diff
@@ -23,58 +24,273 @@
...
+                  String in = "foo";
+                  String[] out = in.split(" ");
+              }
+              """
+          )
+        );
+    }
+
+    @Test
+    @Disabled("Argument variant not covered yet")
```
timtebeek, line 134: I think the corresponding recipe was missing here.

knutwannheden:
I suggest we also try to declare them as private, so that there won't be quite as many user-visible tiny recipes in the catalog.

timtebeek:
Given the amount of repetition in the tests I propose we instead adopt a pattern where we test multiple recipes in one test:
```java
@Test
void uberTest() {
    rewriteRun(
      //language=java
      java(
        """
          import org.apache.commons.lang3.StringUtils;
          
          class Foo {
              String in = "foo";
              String defaultString = StringUtils.defaultString(in);
              boolean isEmpty = StringUtils.isEmpty(in);
              String[] split = StringUtils.split(in);
              //String[] split2 = StringUtils.split(in, ", ");
              boolean equals = StringUtils.equals(in, "string");
          }
          """,
        """
          import java.util.Objects;
                        
          class Foo {
              String in = "foo";
              String defaultString = Objects.toString(in);
              boolean isEmpty = in == null || in.isEmpty();
              String[] split = in.split(" ");
              //String[] split2 = StringUtils.split(in, ", ");
              boolean equals = Objects.equals(in, "string");
          }
          """
      )
    );
}
```
Given the low complexity I think that's good-enough for these types of recipes & tests, and helps reduce the amount of code we produce and maintain.

src/main/java/org/openrewrite/java/migrate/apache/commons/lang/ApacheCommonsStringUtils.java
AlekSimpson: Where can I find that option?

src/main/java/org/openrewrite/java/migrate/apache/commons/lang/ApacheCommonsStringUtils.java
AlekSimpson: Seems like there still is an issue with these recipes. The error I get is `error: unclosed string literal .builder("String.format(\"%\" #{any(int)} l + \"s#{any(java.lang.String)}", s)");`

AlekSimpson:
> I suggest we also try to declare them as private, so that there won't be quite as many user-visible tiny recipes in the catalog.

Do you mean the refaster class templates?

src/main/java/org/openrewrite/java/migrate/apache/commons/lang/ApacheCommonsStringUtils.java
AlekSimpson: Seems like the last '"' after `String)}` should also be escaped

AlekSimpson:
> Given the amount of repetition in the tests I propose we instead adopt a pattern where we test multiple recipes in one test:
> 
> ```java
> @Test
> void uberTest() {
>     rewriteRun(
>       //language=java
>       java(
>         """
>           import org.apache.commons.lang3.StringUtils;
>           
>           class Foo {
>               String in = "foo";
>               String defaultString = StringUtils.defaultString(in);
>               boolean isEmpty = StringUtils.isEmpty(in);
>               String[] split = StringUtils.split(in);
>               //String[] split2 = StringUtils.split(in, ", ");
>               boolean equals = StringUtils.equals(in, "string");
>           }
>           """,
>         """
>           import java.util.Objects;
>                         
>           class Foo {
>               String in = "foo";
>               String defaultString = Objects.toString(in);
>               boolean isEmpty = in == null || in.isEmpty();
>               String[] split = in.split(" ");
>               //String[] split2 = StringUtils.split(in, ", ");
>               boolean equals = Objects.equals(in, "string");
>           }
>           """
>       )
>     );
> }
> ```
> 
> Given the low complexity I think that's good-enough for these types of recipes & tests, and helps reduce the amount of code we produce and maintain.

Do we want to group them together in groups of 4 or 5?

knutwannheden:
> > I suggest we also try to declare them as private, so that there won't be quite as many user-visible tiny recipes in the catalog.
> 
> Do you mean the refaster class templates?

Yes. The nested Refaster template classes. But for testing purposes package private would be slightly better. I don't really know which visibility will cause a recipe to get listed in the catalog.

timtebeek:
> Do we want to group them together in groups of 4 or 5? Or were you imagining that they would all for the most part be in one test?

We can start with all in one; and if there's an interesting variants you want to test for more complicated recipes then I'd cover those in a separate test.

src/main/java/org/openrewrite/java/migrate/apache/commons/lang/ApacheCommonsStringUtils.java
timtebeek: Can get it to work here; the option appears to be named something like rearrange, but you might need to look at how to get that working, or just quickly drag and drop from the side bar.

src/main/java/org/openrewrite/java/migrate/apache/commons/lang/ApacheCommonsStringUtils.java
AlekSimpson: Well it happens every time I regenerate the recipes so I assume its something in rewrite-templating. This is on my desktop which didn't ever have the local rewrite-templating dependency so I don't think it could be an issue with that. If you uncomment the code and try to rebuild do you see the error?

timtebeek: Confirmed; we can ignore this one for now then, and focus on the others to get a first batch through.

src/test/java/org/openrewrite/java/migrate/apache/commons/lang/ApacheCommonsStringUtilsTest.java
```diff
@@ -45,6 +45,20 @@ void defaultString() {
...
+                  String out = StringUtils.chop(in);
+                  String out = StringUtils.replace(in, "o", "z");
+                  String out = StringUtils.strip(in);
+                  String out = StringUtils.join(in);
+                  String out = StringUtils.deleteWhitespace(in);
+                  String out = StringUtils.abbreviate(in, 5);
+                  String out = StringUtils.trimToEmpty(in);
+                  String out = StringUtils.substringAfter(in, ",");
+                  String out = StringUtils.right(in, 5);
+                  String out = StringUtils.mid(in, 2, 4);
```
timtebeek, line 61: This won't compile with the left side variables all named the same, right? Perhaps best to reassign per type, such that you have multiple statements `str = StringUtils.xyz(in, ...)` and `arr = StringUtils.split(in);`

AlekSimpson: Hm ya, I can fix this. Its odd though because I do remember being able to run the test and it passed fine. I wonder why this wasn't a bigger issue?

timtebeek: I think we don't validate that we can actually compile the output of the recipe, because that's not always the case when you migrate but need a different version of a dependency on the classpath suddenly. Hence why we need to make sure the code looks at least plausible, ideally compiles with a manual inspection.

AlekSimpson:
Ok I have finished adding all the refaster template classes for all `StringUtils` methods that could be replaced. Most of them pass, a couple don't work due to a bug with `rewrite-templating` and the `Chomp` recipe does not work either. `Chomp` could possibly be not working due to another `rewrite-templating` bug. 

I also found the class arrangement tabs in IDEA to sort the refaster template classes. I'm not sure what criteria to use to sort the class though. @timtebeek how did you want the classes to be sorted?

timtebeek:
Figured best to sort alphabetically, that way it's easiest to spot any that are missing.

src/main/java/org/openrewrite/java/migrate/apache/commons/lang/ApacheCommonsStringUtils.java
```diff
@@ -0,0 +1,607 @@
...
+
+    private static class Abbreviate {
+        @BeforeTemplate
+        String before(String s1, int width) {
+            return StringUtils.abbreviate(s1, width);
+        }
+
+        @AfterTemplate
+        String after(String s, int width) {
+            return s.substring(0, Math.min(s.length(), width));
```
knutwannheden, line 38: I believe there are many overloads of `abbreviate()`. This particular one adds a trailing ellipsis. So you probably want something like:
```suggestion
            return s == null || s.length() <= width ? s : s.substring(0, width - 3) + "...";
```

src/main/java/org/openrewrite/java/migrate/apache/commons/lang/ApacheCommonsStringUtils.java
```diff
@@ -0,0 +1,607 @@
...
+
+    private static class Capitalize {
+        @BeforeTemplate
+        String before(String s) {
+            return StringUtils.capitalize(s);
+        }
+
+        @AfterTemplate
+        String after(String s) {
+            return s.substring(0, 1).toUpperCase() + s.substring(1);
```
knutwannheden, line 50: I believe this is wrong with respect to nullability. I believe most of `StringUtils` methods accept `null` as input, so you should probably in most cases use the ternary operator to guard against that.

src/main/java/org/openrewrite/java/migrate/apache/commons/lang/ApacheCommonsStringUtils.java
```diff
@@ -0,0 +1,607 @@
...
+            return StringUtils.contains(s, search);
+        }
+
+        @AfterTemplate
+        boolean after(String s, String search) {
+            return s.contains(search);
+        }
+    }
+
+    private static class CountMatches {
```
knutwannheden, line 90: I think I would probably remove this rule. While it probably is possible to do something like this, we would have to properly escape the pattern (as the input to `countMatches()` isn't a regex) and also using regex has a much higher computational overhead.

We can however provide a rule for `countMatches(String, char)` by using something like `s.chars().filter(c -> c == pattern).count()` (or `s.codePoints().filter(c -> c == pattern).count()`).

src/main/java/org/openrewrite/java/migrate/apache/commons/lang/ApacheCommonsStringUtils.java
```diff
@@ -0,0 +1,607 @@
...
+
+    private static class EndsWithIgnoreCase {
+        @BeforeTemplate
+        boolean before(String s, String suffix) {
+            return StringUtils.endsWithIgnoreCase(s, suffix);
+        }
+
+        @AfterTemplate
+        boolean after(String s, String suffix) {
+            return s.regionMatches(true, 0, suffix, 0, suffix.length());
```
knutwannheden, line 134: This doesn't yet account for nulls, but should otherwise be correct:
```suggestion
            return s.regionMatches(true, s.length() - suffix.length(), suffix, 0, suffix.length());
```

src/main/java/org/openrewrite/java/migrate/apache/commons/lang/ApacheCommonsStringUtils.java
```diff
@@ -0,0 +1,607 @@
...
+        @BeforeTemplate
+        int before(String s, String search) {
+            return StringUtils.indexOfAny(s, search);
+        }
+
+        @AfterTemplate
+        int after(String s, String search) {
+            return IntStream.range(0, s.length())
+                    .filter(i -> search.indexOf(s.charAt(i)) >= 0)
+                    .findFirst()
```
knutwannheden, line 172:
```suggestion
                    .min()
```

src/main/java/org/openrewrite/java/migrate/apache/commons/lang/ApacheCommonsStringUtils.java
```diff
@@ -0,0 +1,607 @@
...
+    }
+
+    private static class IsAlphanumericSpace {
+        @BeforeTemplate
+        boolean before(String s) {
+            return StringUtils.isAlphanumericSpace(s);
+        }
+
+        boolean after(String s) {
+            return s.matches("^[a-zA-Z0-9\\s]*$");
```
knutwannheden, line 184: I suspect this isn't correct with respect to Unicode characters (i.e. non ASCII characters). I suggest to instead use a `Stream` based on `String#chars()` like you do in `IsAlpha`.

knutwannheden (commented):
Didn't check all rules, but added a few comments.

build.gradle.kts
```diff
@@ -51,6 +52,7 @@ dependencies {
     testImplementation("commons-codec:commons-codec:1.+")
     testImplementation("org.apache.commons:commons-lang3:3.+")
```
timtebeek, line 53: For dependencies pulled up to `implementation` we don't need to declare them again for tests.
```suggestion
```

src/test/java/org/openrewrite/java/migrate/apache/commons/lang/ApacheCommonsStringUtilsTest.java
```diff
@@ -0,0 +1,188 @@
...
+                      outString = StringUtils.stripStart(in, "chars");
+                      outString = StringUtils.uncapitalize(in);
+                      outString = StringUtils.capitalize(in);
+                      outString = StringUtils.removeEnd(in, "remove");
+                      int outInt = StringUtils.countMatches(in, "pattern");
+                      outString = StringUtils.trimToNull(in);
+                      outInt = StringUtils.indexOfAny(in, "search");
+                      outString = StringUtils.swapCase(in);
+                      outBool = StringUtils.isAlpha(in);
+                      outBool = StringUtils.isAlphaSpace(in);
```
timtebeek, line 72: I'd sort these based on the method called as well, such that it's again easier to spot what might not be covered by tests yet.

AlekSimpson:
@timtebeek Ok, I've commented out a bunch of the templates that I am not 100% sure are accurate replacements. I also added a test to explore the question of whether replacements should be surrounded in parentheses. 

I also think having all the tests grouped together into one test is a little tough to manage. Currently when I run the super test it throws a Java error but its not easy to tell which recipe(s) is/are causing it. Maybe its worth it to keep them in groups of 20? This its easier to track down the source of errors if they arise. 

I also wonder if it might be useful for large scale refaster recipe development to include an annotation that can disable specific templates. Maybe something like `@DisableTemplate`? This way refaster templates wouldn't have to be commented out like in this PR they could just be disable until ready. Its just a thought I had today that I thought was worth mentioning.

src/test/java/org/openrewrite/java/migrate/apache/commons/lang/ApacheCommonsStringUtilsTest.java
```diff
@@ -82,31 +97,46 @@ void test() {
...
+                      str = (int) (in == null ? 0 : in.chars().filter(c -> c == "pattern").count());
+                      str = in == null ? null : in.replaceAll("\\s+", "");
+                      boolean bool = in == null ? null : in.replaceAll("\\s+", "");
+                      bool = in.equalsIgnoreCase("other");
+                      bool = Objects.equals(in, "other");
+                      str = Objects.equals(in, "other");
+                      bool = in == null ? false : in.matches("^[a-zA-Z0-9\\s]*$");
+                      bool = in == null ? false : in.chars().allMatch(Character::isAlphabetic);
+                      bool = in == null ? false : in.matches("[a-zA-Z\\s]+");
+                      bool = in == null ? false : in.chars().allMatch(Character::isLetter);
```
timtebeek, line 114: On these four my IDE has some suggestions for simplification; might be an Ultimate feature; are you using that as a student?
image

src/test/java/org/openrewrite/java/migrate/apache/commons/lang/ApacheCommonsStringUtilsTest.java
```diff
@@ -82,31 +97,46 @@ void test() {
...
-                      outString = in == null || in.trim() == null ? null : in.trim();
-                      outInt = IntStream.range(0, in.length()).filter((i) -> "search".indexOf(in.charAt(i)) >= 0).findFirst().orElse(-1);
-                      outString = in.chars().map((c) -> Character.isUpperCase(c) ? Character.toLowerCase(c) : Character.toUpperCase(c)).collect(StringBuilder::new, StringBuilder::appendCodePoint, StringBuilder::append).toString();
-                      outBool = in.chars().allMatch(Character::isLetter);
-                      outBool = in.matches("[a-zA-Z\\s]+");
+                      
+                      String str = in == null || in.length() <= 10 ? in : in.substring(0, 10 - 3) + "...";
+                      str = in == null ? null : in.substring(0, 1).toUpperCase() + in.substring(1);
+                      str = in == null ? null : in.substring(0, in.length() - 1);
+                      str = in == null || "search" == null ? null : in.contains("search");
```
timtebeek, line 104: This one doesn't compile.

src/test/java/org/openrewrite/java/migrate/apache/commons/lang/ApacheCommonsStringUtilsTest.java
```diff
@@ -82,31 +97,46 @@ void test() {
...
+                      str = (int) (in == null ? 0 : in.chars().filter(c -> c == "pattern").count());
+                      str = in == null ? null : in.replaceAll("\\s+", "");
+                      boolean bool = in == null ? null : in.replaceAll("\\s+", "");
```
timtebeek, line 107: This one doesn't compile.

src/test/java/org/openrewrite/java/migrate/apache/commons/lang/ApacheCommonsStringUtilsTest.java
```diff
@@ -82,31 +97,46 @@ void test() {
...
+                      str = in == null ? null : (in.endsWith("remove") ? in.substring(0, in.length() - "remove".length()) : in);
+                      str = in == null ? null : new String(new char[3]).replace("\\0", in);
+                      str = in == null ? null : in.replaceFirst(Pattern.quote("search"), "replacement");
+                      str = in == null ? null : in.replaceAll("target", "replacement");
+                      str = in == null ? null : new StringBuilder(in).reverse().toString();
+                      str = in == null ? null : in.substring(in.length() - l, in.length() - 1);
+                      str = in == null ? null : in.split(", ");
+                      str = in == null ? null : (in.endsWith("suffix") ? in.substring(0, in.lastIndexOf("suffix")) : in);
+                      str = in == null ? null : (in.startsWith("chars") ? in.substring("chars".length()) : in);
+                      str = in == null || "prefix" == null ? null : in.startsWith("prefix");
```
timtebeek, line 130: This one doesn't compile.

src/test/java/org/openrewrite/java/migrate/apache/commons/lang/ApacheCommonsStringUtilsTest.java
```diff
@@ -82,31 +97,46 @@ void test() {
...
+                      bool = in == null ? false : in.chars().allMatch(Character::isLetter);
+                      bool = in == null || in.isEmpty();
+                      str = in == null ? null : String.join("", in);
+                      str = in == null ? null : in.substring(0, 4);
+                      str = in == null ? null : in.toLowerCase();
+                      str = in == null ? null : (3 + 3 < in.length() ? in.substring(3, 3 + 3) : in.substring(3, in.length() - 1));
+                      str = in == null ? null : in.substring(0, 3) + "overlay" + in.substring(3);
+                      str = in == null ? null : (in.endsWith("remove") ? in.substring(0, in.length() - "remove".length()) : in);
+                      str = in == null ? null : new String(new char[3]).replace("\\0", in);
+                      str = in == null ? null : in.replaceFirst(Pattern.quote("search"), "replacement");
```
timtebeek, line 123: Missing Pattern import

src/test/java/org/openrewrite/java/migrate/apache/commons/lang/ApacheCommonsStringUtilsTest.java
AlekSimpson: Oh no I'm using the community edition. I will look into downloading the ultimate edition today.

src/test/java/org/openrewrite/java/migrate/apache/commons/lang/ApacheCommonsStringUtilsTest.java
```diff
@@ -0,0 +1,266 @@
...
+              class Foo {
+                  void test(String s) {
+                      String test = StringUtils.strip(s).toString();
+                  }
+              }
+              """,
+            """
+              class Foo {
+                  void test(String s) {
+                      String test = (s == null ? null : in.trim()).toString();
```
timtebeek, line 234: @knutwannheden this test currently fails; it's similar to a case we've discussed before on whether the `JavaTemplate`
 should itself detect whether a statements needs to be wrapped in braces. In this synthetic example we're changing behavior: before it would have thrown a NPE when the input is `null`, after non-braced replacement it no longer would. We should not make such behavioral changes.
```suggestion
                      String test = s == null ? null : in.trim().toString();
```
https://ge.openrewrite.org/s/jwi2f75wlzjts/tests/task/:test/details/org.openrewrite.java.migrate.apache.commons.lang.ApacheCommonsStringUtilsTest/canCallMethodOnResult()?top-execution=1

src/test/java/org/openrewrite/java/migrate/apache/commons/lang/ApacheCommonsStringUtilsTest.java
```diff
@@ -0,0 +1,266 @@
...
+              class Foo {
+                  void test(String s) {
+                      String test = !StringUtils.equalsIgnoreCase(s, "other");
+                  }
+              }
+              """,
+            """
+              class Foo {
+                  void test(String s) {
+                      String test = !(s != null && !s.equalsIgnoreCase("other"));
```
timtebeek, line 259: Another variant @knutwannheden ; this one actually produces code that does not compile, but we should maintain the previous behavior.
```suggestion
                      String test = !(s != null && !s.equalsIgnoreCase("other"));
```
https://ge.openrewrite.org/s/jwi2f75wlzjts/tests/task/:test/details/org.openrewrite.java.migrate.apache.commons.lang.ApacheCommonsStringUtilsTest/invertedBooleanHandled()?top-execution=1

timtebeek:
We need a merge of this one to fix the parentheses cases:
- https://github.com/openrewrite/rewrite-templating/pull/22

src/test/java/org/openrewrite/java/migrate/apache/commons/lang/ApacheCommonsStringUtilsTest.java
timtebeek: Fixed in https://github.com/openrewrite/rewrite-templating/pull/22/

src/test/java/org/openrewrite/java/migrate/apache/commons/lang/ApacheCommonsStringUtilsTest.java
```diff
@@ -0,0 +1,287 @@
...
+    void inputMethodsNotCalledTwice() {
+        rewriteRun(
+          //language=java
+          java(
+            """
+              import org.apache.commons.lang3.StringUtils;
+                            
+              class Foo {
+                  void test(String s) {
+                      String test = StringUtils.strip(bar()).toString();
```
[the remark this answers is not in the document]
timtebeek: Logged as
- https://github.com/openrewrite/rewrite-templating/issues/27

build.gradle.kts
```diff
@@ -49,8 +50,8 @@ dependencies {
     testImplementation("com.google.guava:guava:29.0-jre")
 
     testImplementation("commons-codec:commons-codec:1.+")
-    testImplementation("org.apache.commons:commons-lang3:3.+")
     testImplementation("org.apache.maven.shared:maven-shared-utils:3.+")
+    testImplementation("org.codehaus.plexus:plexus-utils:3.+")
```
timtebeek, line 54:
```suggestion
```

PR #272, opened 2023-08-13
Plexus FileUtils refaster template recipes

timtebeek:
## What's changed?
Added recipes to convert from Plexus FileUtils to either JDK or Apache Commons IO FileUtils.

## What's your motivation?
Standardizing on a single implementation aids later migration to JDK internals, or at least reduces the number of dependencies.

Matchers fail to match 🤔 

@knutwannheden 

## Any additional context
https://issues.apache.org/jira/browse/MNG-6825
Assumes https://github.com/openrewrite/rewrite-templating/pull/17 gets merged

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -29,6 +29,9 @@ dependencies {
     implementation("org.openrewrite.recipe:rewrite-static-analysis:$rewriteVersion")
     implementation("org.openrewrite.gradle.tooling:model:$rewriteVersion")
 
+    implementation("commons-io:commons-io:2.+")
+    implementation("org.codehaus.plexus:plexus-utils:3.+")
+
     runtimeOnly("org.openrewrite:rewrite-java-8")
     runtimeOnly("org.openrewrite:rewrite-java-11")
     runtimeOnly("org.openrewrite:rewrite-java-17")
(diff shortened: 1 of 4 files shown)


knutwannheden (commented):
Since we now have multiple FileUtils classes with the same name, can you add a test and see what happens when the source also references a method which isn't migrated? I suspect we will get conflicting imports.

timtebeek:
@knutwannheden This PR is now at the stage where we insert FQN everywhere, when comparing expected vs actual.
Should we try again with a variant of https://github.com/openrewrite/rewrite-templating/pull/18 merged?

That would mean we have to update the expected values here, but overall I think should improve both the recipe development experience (able to use FQN) and resulting output (non-FQN). What are your thoughts at this stage?

Of course we can iterate again later, but it'd be nice to have a mostly functional set of recipes with some broad assumptions.

```diff
Unexpected result in "Test.java"
diff --git a/Test.java b/Test.java
index 1880345..310d3d3 100644
--- a/Test.java
+++ b/Test.java
@@ -1,13 +1,13 @@ 
 import java.io.File;
 import java.io.IOException;
-import org.apache.commons.io.FileUtils;
+import org.codehaus.plexus.util.FileUtils;
 class Test {
     void test() throws IOException {
-        FileUtils.deleteDirectory(new File("test"));
+        org.apache.commons.io.FileUtils.deleteDirectory(new File("test"));
         org.apache.commons.io.FileUtils.deleteDirectory(new File("test"));
 
         File file = new File("test");
-        FileUtils.deleteDirectory(file);
+        org.apache.commons.io.FileUtils.deleteDirectory(file);
         org.apache.commons.io.FileUtils.deleteDirectory(file);
 
         FileUtils.dirname("/foo/bar"); // Unused
```

knutwannheden:
Yes, let's do that. We won't be able to have full control anyway.

PR #274, opened 2023-08-18
Apache Commons File Utils Migrations

AlekSimpson:
## What's changed?
This PR adds two refaster templates to migrate `FileUtils` methods. 

## What's your motivation?
Moving onto `FileUtils` after [StringUtils](https://github.com/openrewrite/rewrite-migrate-java/pull/271#event-10131923309)

So currently it won't build because the `Write` template is a little off. I think the reason its failing with this one is that `FileUtils.write()` returns void while Java's write method `Files.write()` returns a Path. I don't think there are any other one line method Java provides to write to a file. Will we just not be able to do migrations for `FileUtils.write()` because of this? 

@timtebeek 

### Checklist


src/main/java/org/openrewrite/java/migrate/apache/commons/lang/ApacheCommonsFileUtils.java
```diff
@@ -0,0 +1,54 @@
...
+public class ApacheCommonsFileUtils {
+    private static class Write {
+        @BeforeTemplate
+        void before(File file, CharSequence data, Charset cs) throws Exception {
+            FileUtils.write(file, data, cs);
+        }
+
+        @AfterTemplate
+        void after(File file, CharSequence data, Charset cs) throws Exception {
+            Files.write(file.toPath(), Arrays.asList(data), cs);
```
timtebeek, line 38: @knutwannheden (no rush! just interesting)
This fails compile in the generated recipe due to
```
java: reference to compile is ambiguous
both method compile(org.openrewrite.java.JavaVisitor<?>,java.lang.String,org.openrewrite.java.JavaTemplate.P3<?,?,?>) in org.openrewrite.java.JavaTemplate
 and method compile(org.openrewrite.java.JavaVisitor<?>,java.lang.String,org.openrewrite.java.JavaTemplate.F3<?,?,?,?>) in org.openrewrite.java.JavaTemplate match
```

timtebeek: Any idea what might be going on here? The generated code is:
```java
final JavaTemplate before = JavaTemplate.compile(this, "before", (java.io.File file, CharSequence data, java.nio.charset.Charset cs) -> FileUtils.write(file, data, cs)).build();
final JavaTemplate after = JavaTemplate.compile(this, "after", (java.io.File file, CharSequence data, java.nio.charset.Charset cs) -> Files.write(file.toPath(), Arrays.asList(data), cs)).build();
```

src/test/java/org/openrewrite/java/migrate/apache/commons/io/ApacheCommonsFileUtilsTest.java
```diff
@@ -0,0 +1,142 @@
...
+                      f = FileUtils.getFile(s);
+                      f = FileUtils.getFile(s, s);
+                      f = FileUtils.toFile(url);
+
+                      str = FileUtils.byteCountToDisplaySize(l);
+                      FileUtils.cleanDirectory(fileA);
+                      bool = FileUtils.contentEquals(fileA, fileB);
+                      FileUtils.copyDirectory(fileA, fileB);
+                      FileUtils.copyFile(fileA, fileB);
+                      FileUtils.copyURLToFile(url, fileA);
```
timtebeek, line 73: I'm missing `copyFileToDirectory` here, among a few other that we might or might not migrate in the future. Could you add those still, keeping the sort order by method name? No need to add overloads just yet; we can do that when we choose to implement methods.

timtebeek (commented):
Let's only expand on the ubertest before we merge this first structure, and from there discuss what methods to pick up.

src/main/java/org/openrewrite/java/migrate/apache/commons/lang/ApacheCommonsFileUtils.java
timtebeek: We can work around this with a simple cast of course. ;/

knutwannheden: Haven't looked at it yet. Possibly we need to generate an explicit cast, because the return type of the called method is not void like that of the template method.

PR #278, opened 2023-08-22
Added ApacheCommonsFileUtils WriteStringToFile Refaster templates

AlekSimpson:
## What's changed?
This PR adds two new refaster templates for the `FileUtils` migration. 

## What's your motivation?
There are still many `FileUtils` methods that don't have refaster templates setup yet.

Both `CopyDirectory` and `WriteStringToFile` throw `java: reference to compile is ambiguous` errors. I think this could maybe be an issue with the templating engine but I'm not sure. 

@timtebeek 

### Checklist


timtebeek:
Sure; we'll try to work the Refaster template processor; I imagine you know what changes you'd need in the tests here?

AlekSimpson:
> Sure; we'll try to work the Refaster template processor; I imagine you know what changes you'd need in the tests here?

Ya currently I think we have both `copyDirectory` and `writeStringToFile` in the tests. When I am able to build these templates I'll check and make sure it makes sense with the tests.

src/main/java/org/openrewrite/java/migrate/apache/commons/io/ApacheCommonsFileUtils.java
```diff
@@ -47,5 +48,30 @@ File after(String name) {
...
 
+    private static class CopyDirectory {
+        @BeforeTemplate
+        void before(File a, File b) throws Exception {
+            FileUtils.copyDirectory(a, b);
+        }
+
+        @AfterTemplate
+        void after(File a, File b) throws Exception {
+            Files.copy(a.toPath(), b.toPath());
```
knutwannheden, line 59: I think the copy options `REPALCE_EXISTING` and probably also `COPY_ATTRIBUTES` should be specified to maintain a behavior as close as possible to that of `FileUtils`. However, in many cases users may not actually be really dependent on `COPY_ATTRIBUTES`, so I am unsure what to do about that one. Also, this is still just an approximation of what `FileUtils` does.

src/main/java/org/openrewrite/java/migrate/apache/commons/io/ApacheCommonsFileUtils.java
```diff
@@ -47,5 +48,30 @@ File after(String name) {
...
+    @SuppressWarnings("deprecation")
+    private static class WriteStringToFile {
+        @BeforeTemplate
+        void before(File a, String s) throws Exception {
+            FileUtils.writeStringToFile(a, s);
+        }
+
+        @AfterTemplate
+        void after(File a, String s) throws Exception {
+            Files.write(a.toPath(), s.getBytes());
```
knutwannheden, line 72: This looks good, AFAICT. We could also add additional recipes to cover some of the other signatures (like `writeStringToFile(File, String, boolean)`, `writeStringToFile(File, String, Charset)`, and `writeStringToFile(File, String, Charset, boolean)`). But it is probably worth doing a search in the SaaS to check how often these get used. Probably they could all be covered by the same recipe, just using different before templates (all with the same input signatures).

AlekSimpson: Sounds good, I'll check the SaaS and see how often these overloads get used. I wanted to make sure this first template was accurate before worrying about overloads.

timtebeek:
Found some commits I hadn't pushed earlier; not sure if our changes were merged correctly. 🤔

src/main/java/org/openrewrite/java/migrate/apache/commons/io/ApacheCommonsFileUtils.java
timtebeek: If this is not a direct replacement, maybe best to leave it out for now @AlekSimpson

src/main/java/org/openrewrite/java/migrate/apache/commons/io/ApacheCommonsFileUtils.java
timtebeek: Sounds like we could use a few more overloads here and merge those separate from the above `CopyDirectory`. Would you mind finishing this one @AlekSimpson ?

src/main/java/org/openrewrite/java/migrate/apache/commons/io/ApacheCommonsFileUtils.java
```diff
@@ -47,5 +48,30 @@ File after(String name) {
...
+        void before(File a, File b) throws Exception {
+            FileUtils.copyDirectory(a, b);
+        }
+
+        @AfterTemplate
+        void after(File a, File b) throws Exception {
+            Files.copy(a.toPath(), b.toPath());
+        }
+    }
+
```
timtebeek, line 62:
```suggestion
```

src/main/java/org/openrewrite/java/migrate/apache/commons/io/ApacheCommonsFileUtils.java
timtebeek: Would also be good to add a single document example test that shows which variants are replaced.

AlekSimpson:
@timtebeek I'm now remembering more context around this PR from when I took a break with it. It seems the replacement method for `writeStringToFile` returns a `Path` but `writeStringToFile` returns void. Was this something you and @knutwannheden were planning on fixing with the refaster engine or should I try and find a write method that also returns void? If I remember correctly though, `Files.write()` is the only method that the Java `Files` class provides for writing to files.

timtebeek:
ah yes; that's where we got stuck before; would it help to add explicit casts to the `after` template that matches the first `before` template type @knutwannheden ? I think we only replace like-for-like here right? Given that the context can differ..

knutwannheden:
I will try to take a look. Should be something we can fix.

knutwannheden:
Should now be fixed with https://github.com/openrewrite/rewrite-templating/commit/b472f3ad4723b36cfb9cfae8a265ab981f1a1b6d.

timtebeek (approved):
Let's do the overloads in a separate PR. Thanks for the help @knutwannheden

PR #285, opened 2023-09-06
Fix Failures in UseVar Recipe discovered by moderne.io

MBoegers:
## What's your motivation?
To fix failures occurring in moderne.io runs, examples:
* spring-cloud/spring-cloud-contract @ main: spring-cloud-contract-verifier/src/test/resources/contractsToCompile/contract_multipart.java java.lang.IllegalArgumentException: Unable to parse expression from JavaType Unknown
* spring-projects/spring-data-commons @ main: src/test/java/org/springframework/data/domain/ManagedTypesUnitTests.java: Expected a template that would generate exactly one statement to replace one statement, but generated 2. Template: var typesSupplier = __P__.<org.springframework.data.domain.ManagedTypesUnitTests.1>/*__p1__*/p()
* // spring-projects/spring-hateoas @ main src/test/java/org/springframework/hateoas/mediatype/html/HtmlInputTypeUnitTests.java: Expected a template that would generate exactly one statement to replace one statement, but generated 2. Template: var numbers = __P__.<java.util.stream.Stream<org.springframework.hateoas.mediatype.html.HtmlInputTypeUnitTests..>>/*__p1__*/p()

no, failures are not an option, here ;)

### Checklist


MBoegers:
I have problems to understand the failure 3.
The test [src/test/java/org/openrewrite/java/migrate/lang/UseVarKeywordTest.BugFixing#duplicateTemplate](src/test/java/org/openrewrite/java/migrate/lang/UseVarKeywordTest.java) reproduces failure 3.

The Problem seems to be that the used JavaTemplate `var #{} = #{any()}` in [UseVarForGenericMethodInvocations.UseVarForGenericsVisitor](https://github.com/MBoegers/rewrite-migrate-java/blob/bugfixing_moderneio_findings/src/main/java/org/openrewrite/java/migrate/lang/var/UseVarForGenericMethodInvocations.java#L53) which is applied in [UseVarForGenericMethodInvocations.UseVarForGenericsVisitor#transformToVarL109](https://github.com/MBoegers/rewrite-migrate-java/blob/bugfixing_moderneio_findings/src/main/java/org/openrewrite/java/migrate/lang/var/UseVarForGenericMethodInvocations.java#L109) produces two statements and I am unable to see why.

knutwannheden:
Possibly you need to invoke `contextSensitive()` on the template builder. This is necessary when to template refers to any elements from the context into which it will be inserted.

timtebeek:
Looks like we already support two of the three new tests covered here; Managed to get the third closer by adding the `.contextSensitive()` as suggested, and fixed some small test issues. Not yet looked at what's needed for a proper fix to that last failing test. Would you like to continue exploring that @MBoegers ?

MBoegers:
I also played around with it but have no clue yet how to fix... 
I would love to have them fixed, because breaking some code bases is not a great option, but TBH I have no time to look into this until 07.October. Feel free to step in 😅

PR #287, opened 2023-09-11
Add PlexusStringUtils refaster recipes

timtebeek:
## What's your motivation?
For https://issues.apache.org/jira/browse/MNG-6825

## TODO
- [ ] Verify null handling matches
- [ ] Consider additional recipes

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/apache/commons/lang/RepeatableArgumentMatcher.java
+++ b/src/main/java/org/openrewrite/java/migrate/apache/commons/lang/RepeatableArgumentMatcher.java
@@ -20,7 +20,7 @@ import org.openrewrite.java.tree.Expression;
 import org.openrewrite.java.tree.J;
 import org.openrewrite.java.tree.TypeUtils;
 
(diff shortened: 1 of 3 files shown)


sambsnyd (approved):

knutwannheden:
@timtebeek For the more complicated cases I wonder if we shouldn't instead provide a recipe which creates a "local" copy of the utils class with implementations as per the after templates. What do you think?

timtebeek:
> @timtebeek For the more complicated cases I wonder if we shouldn't instead provide a recipe which creates a "local" copy of the utils class with implementations as per the after templates. What do you think?

Hmm; I'd for now had postponed the more complicated cases, as I think that should be a broader discussion with the folks at Apache Maven. For now we're gradually reducing the dependency on those libraries, with a lot of work to be done to cover additional methods, and additional utility classes. It's not yet clear if a complete removal is possible, and whether some duplication would be tolerable to facilitate that. We'll have to learn more over time, and start a discussion once we've got a better understanding of what's left to do, and what that needs to complete.

PR #290, opened 2023-09-13
`javax.security.cert`, and `com.sun.net.ssl.internal.ssl.Provider` migration recipes

cjobinabo:
## What's changed?
This pull request includes recipes for the following migration issues:
Screenshot 2023-09-13 at 5 14 40 PM
Screenshot 2023-09-13 at 5 15 41 PM

The test for these recipes currently fail with the following exception:
```
io.github.classgraph.ClassGraphException: Uncaught exception during scan
	at io.github.classgraph.ClassGraph.scan(ClassGraph.java:1637)
	at io.github.classgraph.ClassGraph.scan(ClassGraph.java:1654)
	at io.github.classgraph.ClassGraph.scan(ClassGraph.java:1667)
	at org.openrewrite.config.ClasspathScanningLoader.scanClasses(ClasspathScanningLoader.java:143)
	at org.openrewrite.config.ClasspathScanningLoader.<init>(ClasspathScanningLoader.java:60)
	at org.openrewrite.config.Environment$Builder.scanRuntimeClasspath(Environment.java:224)
	at org.openrewrite.java.migrate.UpgradeToJava17Test.defaults(UpgradeToJava17Test.java:34)
	at org.openrewrite.test.RewriteTest.rewriteRun(RewriteTest.java:140)
	at org.openrewrite.test.RewriteTest.rewriteRun(RewriteTest.java:132)
	at org.openrewrite.test.RewriteTest.rewriteRun(RewriteTest.java:127)
	at org.openrewrite.java.migrate.UpgradeToJava17Test.upgradeFromJava8ToJava17(UpgradeToJava17Test.java:42)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
Caused by: java.lang.OutOfMemoryError: Java heap space
	at java.base/java.util.Arrays.copyOf(Arrays.java:3568)
	at nonapi.io.github.classgraph.fileslice.reader.ClassfileReader.readTo(ClassfileReader.java:206)
	at <unknown class>.<unknown method>(Unknown Source)
	at io.github.classgraph.Classfile.readMethods(Classfile.java:1482)
	at io.github.classgraph.Classfile.<init>(Classfile.java:2027)
	at io.github.classgraph.Scanner$ClassfileScannerWorkUnitProcessor.processWorkUnit(Scanner.java:725)
	at io.github.classgraph.Scanner$ClassfileScannerWorkUnitProcessor.processWorkUnit(Scanner.java:647)
	at nonapi.io.github.classgraph.concurrency.WorkQueue.runWorkLoop(WorkQueue.java:246)
	at nonapi.io.github.classgraph.concurrency.WorkQueue.runWorkQueue(WorkQueue.java:161)
	at io.github.classgraph.Scanner.processWorkUnits(Scanner.java:316)
	at io.github.classgraph.Scanner.performScan(Scanner.java:969)
	at io.github.classgraph.Scanner.openClasspathElementsThenScan(Scanner.java:1114)
	at io.github.classgraph.Scanner.call(Scanner.java:1148)
	at io.github.classgraph.Scanner.call(Scanner.java:82)
```

My IntelliJ IDE is currently set to use 12GBs of memory. I noticed that other existing tests on this repo also fail so I'm curious to see if the builds show the same failure. If they do I'm hoping to get some guidance on how to resolve these.

A test currently doesn't exist for the `org.openrewrite.java.migrate.RemovedLegacySunJSSEProviderName` recipe. The core code for that currently lives in the rewrite repo. I'm wondering if I can move org.openrewrite.java.ChangeStringLiteral from the rewrite-liberty repo to the core rewrite repo.

@timtebeek @joanvr 

### Checklist


src/main/resources/META-INF/rewrite/java-version-17.yml
```diff
@@ -47,3 +49,41 @@ recipeList:
...
+type: specs.openrewrite.org/v1beta/recipe
+name: org.openrewrite.java.migrate.DeprecatedLogRecordMethods
+displayName: Use getLongThreadID and setLongThreadID
+description: Avoid using the deprecated methods in java.util.logging.LogRecord
+tags:
+  - java17
+recipeList:
+  - org.openrewrite.java.ChangeMethodName:
+      methodPattern: java.util.logging.LogRecord getThreadID()
+      newMethodName: getLongThreadID
```
timtebeek, line 84: The old method returned an `int`, whereas the new method returns a `long`. If the return value is assigned to an `int` field or variable, or passed into a method that accepts an `int`, we might need to change the type to `long`, or the compiler won't allow the code change.

src/main/resources/META-INF/rewrite/java-version-17.yml
```diff
@@ -47,3 +49,41 @@ recipeList:
...
+type: specs.openrewrite.org/v1beta/recipe
+name: org.openrewrite.java.migrate.RemovedLegacySunJSSEProviderName
+displayName: Use "SunJSSE"
+description: The com.sun.net.ssl.internal.ssl.Provider provider name was removed
+tags:
+  - java17
+recipeList:
+  - org.openrewrite.java.ChangeStringLiteral:
+      oldFullyQualifiedTypeName: com\.sun\.net\.ssl\.internal\.ssl\.Provider
+      newFullyQualifiedTypeName: SunJSSE
```
timtebeek, line 73: We can pull `ChangeStringLiteral` here first, as it does not appear to be used in rewrite-liberty right now. We might want to rename that recipe to something that better reflects it's use; I'm thinking `ReplaceFirstStringLiteralMatch` (or similar), and reconsider if the arguments should really support a regular expression as input, should maybe use `Matcher.matches()` over `Matcher.find()`, and whether we indeed only want to replace the first occurrence. Things might be simpler if we start with just replacing string literals as strings anywhere in another String literal.

src/main/resources/META-INF/rewrite/java-version-17.yml
```diff
@@ -47,3 +49,41 @@ recipeList:
...
+  - java17
+recipeList:
+  - org.openrewrite.java.ChangeType:
+      oldFullyQualifiedTypeName: javax.security.cert
+      newFullyQualifiedTypeName: java.security.cert
+---
+type: specs.openrewrite.org/v1beta/recipe
+name: org.openrewrite.java.migrate.RemovedLegacySunJSSEProviderName
+displayName: Use "SunJSSE"
+description: The com.sun.net.ssl.internal.ssl.Provider provider name was removed
```
timtebeek, line 67: Descriptions can use backticks and other markdown symbols to enhance the readability in the docs; best to apply those around code references where we can.
```suggestion
description: The `com.sun.net.ssl.internal.ssl.Provider` provider name was removed
```

src/main/resources/META-INF/rewrite/java-version-17.yml
```diff
@@ -33,6 +33,8 @@ recipeList:
   - org.openrewrite.github.SetupJavaUpgradeJavaVersion
   - org.openrewrite.staticanalysis.InstanceOfPatternMatch
   - org.openrewrite.java.migrate.lang.UseTextBlocks
+  - org.openrewrite.java.migrate.DeprecatedLogRecordMethods
+  - org.openrewrite.java.migrate.DeprecatedJavaxSecurityCert
```
timtebeek, line 37: Any reason not to include `RemovedLegacySunJSSEProviderName` here?

timtebeek (commented):
Thanks for adding these recipes; I've added some comments where I think we can improve, but a good start overall. I think moving (and renaming) `ChangeStringLiteral` here would be a first step to getting the tests going.

src/main/resources/META-INF/rewrite/java-version-17.yml
cjobinabo: Yeah, it looks like the `rewrite-liberty` repo is no longer using `ChangeStringLiteral`. I can bring it over here with a new name. The regex capabilities would be necessary for a recipe we will write sometime next year. I can remove the capability for now until it is necessary.

src/main/resources/META-INF/rewrite/java-version-17.yml
cjobinabo: You're right. I overlooked the type of the value getting returned. Is it possible to account for that in the YAML, or would this need to change to a Java recipe?

timtebeek: That would have to be a custom recipe I'm afraid. Likely find any assignments to fields and variables first, as that's what I think is most likely, and methods we can tackle separately. Perhaps best to take that one on in a separate PR, such that we can merge the other two recipes here early.
`setLongThreadID` would have no such issues, as we can upcast from `int` to `long`.

src/main/resources/META-INF/rewrite/java-version-17.yml
cjobinabo: I'm actually not sure `ReplaceFirstStringLiteralMatch` is a better fit. For our use case the match would need to be all or nothing. For that reason I also think leaving it as a regular expression would be best to make it more flexible for other use cases. The recipe would need a way to cover cases such as the one defined here https://github.com/openrewrite/rewrite-liberty/blob/main/src/test/java/org/openrewrite/java/liberty/InvalidInitialContextTest.java. That is a rule we plan on adding.

timtebeek: I've gone and renamed the recipe in 9f02edb to align with others in other modules, with the goal being to have a uniform experience across recipes in terms of naming and arguments. hope you agree!

src/test/java/org/openrewrite/java/migrate/UpgradeToJava17Test.java
```diff
@@ -33,7 +33,7 @@ public void defaults(RecipeSpec spec) {
         spec.recipe(Environment.builder()
           .scanRuntimeClasspath("org.openrewrite.java.migrate")
           .build()
-          .activateRecipes("org.openrewrite.java.migrate.UpgradeToJava17")).allSources(s -> s.markers(javaVersion(17)));
+          .activateRecipes("org.openrewrite.java.migrate.UpgradeToJava17"));
```
cjobinabo, line 36: Thanks! I thought I removed this after testing. I guess not.

src/main/resources/META-INF/rewrite/java-version-17.yml
timtebeek: I've restored the `setThreadID`; don't think it's very common, but we can trivially cover that already.
https://github.com/openrewrite/rewrite-migrate-java/pull/290/commits/36f8490f0e827673fa5d8329f5012a106230e0de

timtebeek (approved):

timtebeek:
Thanks again @cjobinabo ; I'm merging these changes early as I think there's enough value in having these already. We can come back to that `int getThreadID()` to `long getThreadID()` later with a custom recipe that takes into account the type.

PR #293, opened 2023-09-15
Rewrite `Stream.collect(Collectors.toUnmodifiableList())` to `Stream.toList()`

Laurens-W:
## What's changed?

Continuation of https://github.com/openrewrite/rewrite-static-analysis/pull/167, which may be declined at this point
Adding a recipe that rewrites Java 11 `Stream.collect(Collectors.toUnmodifiableList())` to Java 16 `Stream.toList()` with an optional toggle to include `Stream.collect(Collectors.toList())` so no harm is done initially.

## What's your motivation?

I asked on slack if there was an opposing recipe to replacestreamtolistwithcollect, we came to the conclusion that it does not exist but does seem justifiable to have.

@timtebeek @joanvr 

### Checklist


src/test/java/org/openrewrite/java/migrate/util/ReplaceStreamCollectWithToListTest.java
```diff
@@ -0,0 +1,183 @@
...
+                          .toList();
+                  }
+              }
+              """
+          )
+        );
+    }
+
+    @Test
+    void reatinComment() {
```
timtebeek, line 150:
```suggestion
    void retainComment() {
```

timtebeek (approved):
Thanks a lot @Laurens-W , both for suggesting the recipe and implementing it. I've done a few polishing changes, but otherwise a direct success. :)

PR #294, opened 2023-09-15
Add refaster templates for writeStringToFile overloads

AlekSimpson:
## What's changed?
This PR adds refaster templates for writeStringToFile overloads.

## What's your motivation?
[This comment from an earlier PR](https://github.com/openrewrite/rewrite-migrate-java/pull/278#pullrequestreview-1627584184)

@timtebeek 

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/apache/commons/io/ApacheCommonsFileUtils.java
+++ b/src/main/java/org/openrewrite/java/migrate/apache/commons/io/ApacheCommonsFileUtils.java
@@ -74,8 +74,8 @@ public class ApacheCommonsFileUtils {
         }
 
         @AfterTemplate
-        void after(File a, String data) throws Exception {
-            Files.write(a.toPath(), data.getBytes(), StandardOpenOption.APPEND);
+        void after(File a, String data, boolean append) throws Exception {
+            Files.write(a.toPath(), data.getBytes(), append ? StandardOpenOption.APPEND : null);
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/openrewrite/java/migrate/apache/commons/io/ApacheCommonsFileUtils.java
```diff
@@ -61,5 +66,70 @@ void after(File a, String s) throws Exception {
...
+    // file, string, boolean
+    private static class WriteStringToFileOverloadOne {
+        @BeforeTemplate
+        void before(File a, String data, boolean append) throws Exception {
+            FileUtils.writeStringToFile(a, data, append);
+        }
+
+        @AfterTemplate
+        void after(File a, String data, boolean append) throws Exception {
+            Files.write(a.toPath(), data.getBytes(), append ? StandardOpenOption.APPEND : null);
```
timtebeek, line 78: It looks like `data` is nullable, and therefore we can't immediately call `.getBytes()` here or above. :/ Guess that was missed in the earlier review. Not sure if we should then continue with these recipes, as a ternary for `data` substantially hurts the readability.

timtebeek: It looks like `Files.writeString` would have been a better candidate; that said given the nullability issues I doubt if we should continue with these recipes. 🤔

PR #295, opened 2023-09-16
Completes LogRecord.getLongThreadID recipe

cjobinabo:
## What's changed?

The pull requests introduces a ChangeMethodInvocationReturnType recipe to handle the final part of the `org.openrewrite.java.migrate.DeprecatedLogRecordThreadID` recipe.
Screenshot 2023-09-13 at 5 16 35 PM

@timtebeek @joanvr 

### Checklist

### DIFF
--- a/src/test/java/org/openrewrite/java/migrate/UpgradeToJava17Test.java
+++ b/src/test/java/org/openrewrite/java/migrate/UpgradeToJava17Test.java
@@ -21,6 +21,7 @@ import org.openrewrite.config.Environment;
 import org.openrewrite.java.marker.JavaVersion;
 import org.openrewrite.test.RecipeSpec;
 import org.openrewrite.test.RewriteTest;
+import org.openrewrite.java.ChangeMethodName;
 
 import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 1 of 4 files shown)


src/test/java/org/openrewrite/java/migrate/UpgradeToJava17Test.java
```diff
@@ -21,6 +21,7 @@
 import org.openrewrite.java.marker.JavaVersion;
 import org.openrewrite.test.RecipeSpec;
 import org.openrewrite.test.RewriteTest;
+import org.openrewrite.java.ChangeMethodName;
```
timtebeek, line 24:
```suggestion
```

timtebeek (approved):
Looks good to me; of course there's cases we don't yet cover such as passing the return type directly into a method that only takes an int for instance, but I think those cases should be relatively rare.

cjobinabo:
Thanks Tim! Yeah, I spent some more time exploring how to handle other test cases the current approach wouldn't handle, but decided to move on for now to write other recipes. I may revisit this recipe once we have more headroom this quarter.

PR #298, opened 2023-09-19
Adding WsWsocServerContainerDeprecationRecipe

ranuradh:
## What's changed?
image

@timtebeek @joanvr

### Checklist


timtebeek (approved):
Thanks for adding this start to Jakarta EE 10 migrations @ranuradh !

I've moved the jar to `src/test/resources`, as declarative recipes can not refer to classpath resources. We'd expect folks running this recipe to have a working classpath containing the replaced `WsWsocServerContainer`, and if so there's no need to distribute that nested jar with rewrite-migrate-java.  Thought to point that out in case you have more details or a different viewpoint there.

PR #301, opened 2023-09-21
Add PreferJavaUtilObjectsRequireNonNullElse recipe for Java 11

josephw:
## What's changed?
This adds `PreferJavaUtilObjectsRequireNonNullElse`, a `ChangeMethodName`/`ChangeMethodTargetToStatic` combination to convert Guava's `MoreObjects.firstNonNull` to the Java 9 equivalent of `Objects.requireNonNullElse`.

## What's your motivation?
I'm aiming to extend the set of Guava calls that are replaced by direct equivalents, and get that included in a bulk recipe.

This method only shows up in Java 9. I've taken the approach of adding a new `NoGuavaJava11` recipe, in the style of the `UpgradeToJavaX` recipes, so this doesn't break Java 8 projects, or require users to nominate the individual recipe directly.

The method is generic; I found that it didn't match with `"firstNonNull(Object, Object)"`, so I'm using `".."`.

As I need this recipe, my alternative is to keep it in a local ruleset.

### Checklist
- [x] I've added unit tests to cover ~both~ positive ~and negative~ cases

### DIFF
--- a/src/main/resources/META-INF/rewrite/no-guava.yml
+++ b/src/main/resources/META-INF/rewrite/no-guava.yml
@@ -63,6 +63,20 @@ recipeList:
 
 ---
 type: specs.openrewrite.org/v1beta/recipe
+name: org.openrewrite.java.migrate.guava.NoGuavaJava11
+displayName: Prefer the Java 11 standard library instead of Guava
+description: >
+  Guava filled in important gaps in the Java standard library and still does. But at least some of Guava's API surface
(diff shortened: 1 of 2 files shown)


timtebeek (approved):
Looks good to me! Thanks for adding an immediate PR.

timtebeek:
You should be able to [use our snapshot versions](https://docs.openrewrite.org/reference/snapshot-instructions) to try this locally; I expect we'll follow up with a release in the next week or two. Thanks again!

PR #307, opened 2023-10-02
Add recipe descriptors to ApacheCommonsStringUtils

timtebeek:
## What's your motivation?
Better documentation once these show up in the docs & platform.

- I think in the past we used the before and after snippets to show document examples.
That might help to (also) do that to document what these recipes actually replace with.
- We could also take the detected method (or complete lambda) in the before template and use that to generate the description (only). That would need special handling for recipes with multiple before templates.


okundzich:
>I think in the past we used the before and after snippets to show document examples.
That might help to (also) do that to document what these recipes actually replace with.

That caused various issues and was removed. It was getting examples from tests I believe. Not sure if we can undertake it again anytime soon.

timtebeek:
Yes I wouldn't bring back the exact thing we had, but we still support that format if we write a different processor for before/after templates only. More relevant now that we will start to show the nested recipes.

PR #312, opened 2023-10-05
Add new JDBC method implementations introduced in Java 6 and 7

cjobinabo:
Screenshot 2023-10-05 at 5 03 57 PM

Screenshot 2023-10-05 at 5 04 53 PM

## What's changed?

This pull request adds a recipes which can be used to add missing method implementations. With this recipes I added 2 new migration recipes to test it.

@timtebeek @joanvr 
It would be nice to include the `// TODO Auto-generated method stub` comment in these methods. Suggestions on how to easily do this would be great if possible. 

I'll likely push more tests to this pr.

### Checklist


cjobinabo:
That's a good point about polymorphic methods. I'll update methodName to use Method patterns. At that point though, `fullyQualifiedClassName` is more redundant since it could be discerned from the method pattern. The fact that method pattern has the ability to support any class may be a reason to keep `fullyQualifiedClassName` since the class name is required for this recipe.

Regarding the auto-generated comment, I was hoping for an easy solution to add it to the recipe code outside of the YML. When I played around with adding it to the YML, but the formatting for the comment was always off. I'll hold off on that addition for now.

joanvr (approved):

src/main/resources/META-INF/rewrite/java-version-6.yml
```diff
@@ -0,0 +1,40 @@
...
+description: Check for new methods on Java SQL interfaces.
+recipeList:
+  - org.openrewrite.java.migrate.AddMissingMethodImplementation:
+      fullyQualifiedClassName: java.sql.Wrapper
+      methodPattern: "*..* isWrapperFor(..)"
+      methodTemplateString: "public boolean isWrapperFor(Class<?> iface) throws java.sql.SQLException { return false; }"
+  - org.openrewrite.java.migrate.AddMissingMethodImplementation:
+      fullyQualifiedClassName: java.sql.Wrapper
+      methodPattern: "*..* unwrap(..)"
+      methodTemplateString: "public <T> T unwrap(Class<T> iface) throws java.sql.SQLException { return null; }"
```
timtebeek, line 40: As far as I can tell this interface was added in 1.6, as were both methods contained therein? Then it does not make sense for anyone to have a class that implements that interface, yet lacks those methods, right? The compiler would not have allowed that to start with.

Should we then even have this file at all?

src/main/java/org/openrewrite/java/migrate/AddMissingMethodImplementation.java
```diff
@@ -0,0 +1,142 @@
...
+import org.openrewrite.java.MethodMatcher;
+import org.openrewrite.java.tree.J;
+import org.openrewrite.java.tree.JavaType;
+import org.openrewrite.java.tree.TypeTree;
+
+import java.util.List;
+
+@Value
+@EqualsAndHashCode(callSuper = false)
+public class AddMissingMethodImplementation extends Recipe {
```
timtebeek, line 35: I don't immediately see any handling of `abstract` classes in the code or tests. I could see folks have an abstract class that implements an interface for others to extend; we would not want to add implementations to such abstract classes I'd think. What do you think there?

src/main/resources/META-INF/rewrite/java-version-6.yml
```diff
@@ -0,0 +1,40 @@
...
+  with equivalents when there is a clear migration strategy.
+tags:
+  - java6
+recipeList:
+  - org.openrewrite.java.migrate.JREWrapperInterface
+---
+type: specs.openrewrite.org/v1beta/recipe
+name: org.openrewrite.java.migrate.JREWrapperInterface
+displayName: Add missing `isWrapperFor` and `unwrap` methods.
+description: Check for new methods on Java SQL interfaces.
```
timtebeek, line 31:
```suggestion
displayName: Add missing `isWrapperFor` and `unwrap` methods.
description: Add method implementations stubs to classes that implement `java.sql.Wrapper`.
```

src/main/resources/META-INF/rewrite/java-version-7.yml
```diff
@@ -0,0 +1,69 @@
...
+tags:
+  - java7
+recipeList:
+  - org.openrewrite.java.migrate.UpgradeToJava6
+  - org.openrewrite.java.migrate.JREJdbcInterfaceNewMethods
+---
+type: specs.openrewrite.org/v1beta/recipe
+name: org.openrewrite.java.migrate.JREJdbcInterfaceNewMethods
+displayName: Adds missing JDBC interface methods.
+description: Check for new methods on JDBC interfaces.
```
timtebeek, line 32:
```suggestion
displayName: Adds missing JDBC interface methods.
description: Add method implementations stubs to classes that implement JDBC interfaces.
```

src/main/resources/META-INF/rewrite/java-version-6.yml
timtebeek: Ah I only just now see that it's an interface that was added into the hierarchy, and sub classes now each need those method implementations.

src/main/resources/META-INF/rewrite/java-version-6.yml
```diff
@@ -0,0 +1,40 @@
...
+recipeList:
+  - org.openrewrite.java.migrate.AddMissingMethodImplementation:
+      fullyQualifiedClassName: java.sql.Wrapper
+      methodPattern: "*..* isWrapperFor(..)"
+      methodTemplateString: "public boolean isWrapperFor(Class<?> iface) throws java.sql.SQLException { return false; }"
```
timtebeek, line 36: The couple cases I checked I see this method implemented as `var1 != null && var1.isAssignableFrom(this.getClass())`. Should we generate that as the default implementation rather than just `false`?

src/main/resources/META-INF/rewrite/java-version-6.yml
cjobinabo: In Java 1.6 the `java.sql.Wrapper` interface was added as a subinterface to the interfaces listed in the screenshot from my original post. The `java.sql.Wrapper` interface includes a contract for the `isWrapperFor` and `unwrap` methods, meaning an implementation is required by any class implementing `java.sql.Wrapper`. So classes previously implementing `java.sql.CallableStatement`, `java.sql.Connection`, etc are now required to provide an implementation for `isWrapperFor` and `unwrap`. This recipe provides an auto-generated method stub for those methods.

src/main/resources/META-INF/rewrite/java-version-6.yml
cjobinabo: We've typically been using the same name we have in our migration reports as the description in our recipes to make them a bit easier to identify. If you feel strongly about this change, we could reconsider that strategy. Otherwise, using our existing names in the recipe `displayName` or `description` would be preferable.

src/main/resources/META-INF/rewrite/java-version-6.yml
cjobinabo: I didn't want to give any impression that the code here is good to go as is. This is really just meant to be an auto-generated method stub that the developer would fill in based on their requirement, similar to built-in quick fixes IDE's provide.

src/main/resources/META-INF/rewrite/java-version-6.yml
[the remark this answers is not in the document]
cjobinabo: The same comment from above would apply here.

src/main/java/org/openrewrite/java/migrate/AddMissingMethodImplementation.java
cjobinabo: That's a fair point. I can update the recipe to check extended classes as well.

src/main/resources/META-INF/rewrite/java-version-6.yml
timtebeek: As also indicated one to one: large organizations run recipes across multiple repositories, and might only spot check the results. In those cases I'd hope to go for a safer default rather wherever possible, especially when a programmer would otherwise have to reproduce the same themselves. That said I do understand that any template produced here might not be sufficient in all cases, but even then I think it might still be preferable.

What's your stance here given the above?

src/main/resources/META-INF/rewrite/java-version-6.yml
cjobinabo: You make a fair point. I also see that some of the names have been changed in Anu's PR so the consistency is already broken. I'll include these suggestions in my next push.

src/test/java/org/openrewrite/java/migrate/UpgradeToJava6Test.java
```diff
@@ -0,0 +1,137 @@
...
+                  }
+               
+                  public <T> T unwrap(Class<T> iface) throws java.sql.SQLException {
+                      try {
+                          if (isWrapperFor(iface)) {
+                              return (T) this;
+                          }
+                          throw DbException.getInvalidValueException("iface", iface);
+                      } catch (Exception e) {
+                          throw logAndConvert(e);
```
timtebeek, line 129: This `logAndConvert` isn't defined anywhere; we likely need to insert something here that's safe to insert in any context. I don't know the specifics of `unwrap` well enough to know what exactly, but possible it's to wrap the exception into an `SQLException`.

src/test/java/org/openrewrite/java/migrate/UpgradeToJava6Test.java
```diff
@@ -86,52 +92,57 @@ public int getLoginTimeout() throws SQLException {
...
                       try {
-                          if (isWrapperFor(iface)) {
+                          if (iface != null && iface.isAssignableFrom(this.getClass())) {
                               return (T) this;
                           }
-                          throw DbException.getInvalidValueException("iface", iface);
+                          throw new java.lang.Exception();
                       } catch (Exception e) {
-                          throw logAndConvert(e);
+                          throw new java.sql.SQLException();
```
timtebeek, line 139: Likely best to include the cause here to aid debugging, right?
```suggestion
                          throw new java.sql.SQLException(e);
```

src/test/java/org/openrewrite/java/migrate/UpgradeToJava6Test.java
```diff
@@ -86,52 +92,57 @@ public int getLoginTimeout() throws SQLException {
...
+                               
                   public <T> T unwrap(Class<T> iface) throws java.sql.SQLException {
+                      // TODO Auto-generated method stub
                       try {
-                          if (isWrapperFor(iface)) {
+                          if (iface != null && iface.isAssignableFrom(this.getClass())) {
                               return (T) this;
                           }
-                          throw DbException.getInvalidValueException("iface", iface);
+                          throw new java.lang.Exception();
```
timtebeek, line 137: I think a small message here would be appreciated by anyone debugging this; and likely this needs to be a `SQLException` too, right?

src/main/java/org/openrewrite/java/migrate/AddMissingMethodImplementation.java
```diff
@@ -0,0 +1,148 @@
...
+            }
+            return false;
+        }
+
+        @Override
+        public J.ClassDeclaration visitClassDeclaration(J.ClassDeclaration classDecl, ExecutionContext executionContext) {
+            boolean implementsInterface = implementsInterface(classDecl);
+            // Don't make changes to classes that don't match the fully qualified name
+            if (classDecl.getType() == null || !implementsInterface) {
+                return classDecl;
```
timtebeek, line 128: Shouldn't we call super here just to be sure we visit any sub classes too?

cjobinabo: That's interesting, so calling super here would take care of subclasses as well. In the other PR I put out yesterday, I ended up manually searching for subclassed. Additional info about this would be good to have.
So I would just change line 128 to `super.visitClassDeclaration(classDecl, executionContext)`?

timtebeek: Usually calling the super method is the first thing we do in an overridden method, and exclusively use the result from that throughout the method after that. Here's a part of the docs that explains that in some more detail:
https://docs.openrewrite.org/authoring-recipes/recipe-conventions-and-best-practices#be-deliberate-about-lst-traversal

cjobinabo: Thanks, Tim, that link was very helpful. I've gone ahead and made the necessary updates in both PRs. I also included a new test in this PR to confirm that sub-classes are detected and updated.

src/main/java/org/openrewrite/java/migrate/AddMissingMethodImplementation.java
```diff
@@ -0,0 +1,151 @@
...
+                    return matchesInterface(type);
+                }
+            }
+
+            // also check extended class to handle abstract classes
+            TypeTree extendedClass = classDecl.getExtends();
+            if (extendedClass != null) {
+                JavaType.Class type = (JavaType.Class) extendedClass.getType();
+                return matchesInterface(type);
+            }
```
timtebeek, line 119: We likely also don't want to add methods to classes that themselves are abstract, instead pushing that down into actual implementations. Or what are your thoughts on that?

timtebeek: As a continuation of https://github.com/openrewrite/rewrite-migrate-java/pull/312/files#r1364317262

src/main/java/org/openrewrite/java/migrate/AddMissingMethodImplementation.java
```diff
@@ -0,0 +1,148 @@
...
+    // All recipes must be serializable. This is verified by RewriteTest.rewriteRun() in your tests.
+    @JsonCreator
+    public AddMissingMethodImplementation(@NonNull @JsonProperty("fullyQualifiedClassName") String fullyQualifiedClassName,
+                                          @NonNull @JsonProperty("methodPattern") String methodPattern,
+                                          @NonNull @JsonProperty("methodTemplateString") String methodTemplateString) {
+        this.fullyQualifiedClassName = fullyQualifiedClassName;
+        this.methodPattern = methodPattern;
+        this.methodTemplateString = methodTemplateString;
+    }
+
```
timtebeek, line 64: Does the Lombok `@Value` annotation not already provide us with a workable constructor?
```suggestion
```

src/main/java/org/openrewrite/java/migrate/AddMissingMethodImplementation.java
```diff
@@ -0,0 +1,148 @@
...
+        return Preconditions.check(new UsesType<>(fullyQualifiedClassName, true), new ClassImplementationVisitor());
+    }
+
+    public class ClassImplementationVisitor extends JavaIsoVisitor<ExecutionContext> {
+        private final JavaTemplate methodTemplate = JavaTemplate.builder(methodTemplateString).build();
+        private final MethodMatcher methodMatcher = new MethodMatcher(methodPattern, true);
+
+        public boolean matchesInterface(JavaType.Class type) {
+            if (type != null) {
+                if (type.getFullyQualifiedName().equals(fullyQualifiedClassName)) {
```
timtebeek, line 86: I think `TypeUtils` has better methods to check types rather than string comparison. Could we switch to that?

src/main/java/org/openrewrite/java/migrate/AddMissingMethodImplementation.java
```diff
@@ -0,0 +1,148 @@
...
+                    .anyMatch(methodDeclaration -> methodMatcher.matches(methodDeclaration, finalClassDecl));
+
+            // If the class already has method, don't make any changes to it.
+            if (methodExists) {
+                return classDecl;
+            }
+
+            classDecl = classDecl.withBody(methodTemplate.apply(new Cursor(getCursor(), classDecl.getBody()),
+                    classDecl.getBody().getCoordinates().lastStatement()));
+            return classDecl;
```
timtebeek, line 145:
```suggestion
            return classDecl.withBody(methodTemplate.apply(new Cursor(getCursor(), classDecl.getBody()),
                    classDecl.getBody().getCoordinates().lastStatement()));
```

src/main/java/org/openrewrite/java/migrate/AddMissingMethodImplementation.java
cjobinabo: From my experience tests have been failing without the constructor. I can try removing it and see if the tests still work.

src/main/java/org/openrewrite/java/migrate/AddMissingMethodImplementation.java
cjobinabo: I'm not familiar with `TypeUtils` what method did you have in mind?

src/main/java/org/openrewrite/java/migrate/AddMissingMethodImplementation.java
timtebeek: Are you passing `-parameters` to the compiler? Don't know your IDE or setup, but I'd expect this to work without the explicit constructor.

src/main/java/org/openrewrite/java/migrate/AddMissingMethodImplementation.java
timtebeek: `TypeUtils.isAssignableTo`, as applied in https://github.com/openrewrite/rewrite-migrate-java/pull/312/commits/99833e569ff7b862b2c5505ee265056cde32b352
Quite a bit simpler, for what I think is the same output.

src/main/resources/META-INF/rewrite/java-version-7.yml
```diff
@@ -0,0 +1,69 @@
...
+      methodPattern: "*..* getNetworkTimeout()"
+      methodTemplateString: "public int getNetworkTimeout() { \n\t// TODO Auto-generated method stub\n return 0; }"
+  - org.openrewrite.java.migrate.AddMissingMethodImplementation:
+      fullyQualifiedClassName: java.sql.Connection
+      methodPattern: "*..* getSchema()"
+      methodTemplateString: "public java.lang.String getSchema() { \n\t// TODO Auto-generated method stub\n return null; }"
+  - org.openrewrite.java.migrate.AddMissingMethodImplementation:
+      fullyQualifiedClassName: java.sql.Connection
+      methodPattern: "*..* setNetworkTimeout(java.util.concurrent.Executor)"
+      methodTemplateString: "public void isWrapperFor(java.util.concurrent.Executor executor, int milliseconds) { \n\t// TODO Auto-generated method stub\n }"
```
timtebeek, line 49: Should this be 
```suggestion
      methodPattern: "*..* setNetworkTimeout(java.util.concurrent.Executor)"
      methodTemplateString: "public void setNetworkTimeout(java.util.concurrent.Executor executor, int milliseconds) { \n\t// TODO Auto-generated method stub\n }"
```

timtebeek (approved):

PR #315, opened 2023-10-06
`Optional<Stream>.filter(Optional::isPresent).map(Optional::get)` to Java 11 `.flatMap(Optional::stream)`

### LINKED ISSUE #313: get rid of Java 8 Optional Stream idiom

Java 8 Optional class didn't provide ::stream method. As a result of this unfortunate mistake a popular idiom was created to flatten a Stream of Optionals: `.filter(Optional::isPresent).map(Optional::get)`.
This was corrected in Java 9 and the idiom may be replaced with `.flatMap(Optional::stream)`
## What precondition(s) should be checked before applying this recipe?

Java version at least 9.

```java
import java.util.Optional;
import java.util.stream.Stream;

class Scratch {
  public void foo() {
    Stream.of(Optional.empty())
      .filter(Optional::isPresent)
      .map(Optional::get);
  }
}
```

```java
import java.util.Optional;
import java.util.stream.Stream;

class Scratch {
  public void foo() {
    Stream.of(Optional.empty())
      .flatMap(Optional::stream);
  }
}
```

## Any additional context
I already created the recipe as a way to familiarise with OpenRewrite. If you find this useful I am ready to create a PR.

aslaski2:
Fixes #313
## Problem
Java 8 Optional class didn't provide `::stream` method. As a result of this unfortunate mistake a popular idiom was created to flatten a Stream of Optionals: `.filter(Optional::isPresent).map(Optional::get)`.
This was corrected in Java 9 and the idiom may be replaced with `.flatMap(Optional::stream)`

This is my first adventure with OpenRewrite, please check everything carefully.

### Checklist
- [x] I've used the IntelliJ IDEA auto-formatter on affected files **except for the flatMap chain**


aslaski2:
Hi @timtebeek! What happens now? Will someone come and review?

src/main/java/org/openrewrite/java/migrate/util/OptionalStreamRecipe.java
```diff
@@ -0,0 +1,133 @@
...
+
+public class OptionalStreamRecipe extends Recipe {
+  @Override
+  public String getDisplayName() {
+    return "Stream<Optional> idiom recipe";
+  }
+
+  @Override
+  public String getDescription() {
+    return "Migrate Java 8 optional stream idiom .filter(Optional::isPresent).map(Optional::get) to Java 11 .flatMap(Optional::stream).";
```
timtebeek, line 47:
```suggestion
    return "Migrate Java 8 `Optional<Stream>.filter(Optional::isPresent).map(Optional::get)` to Java 11 `.flatMap(Optional::stream)`.";
```

src/main/java/org/openrewrite/java/migrate/util/OptionalStreamRecipe.java
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  @Override
+  public TreeVisitor<?, ExecutionContext> getVisitor() {
+    return new OptionalStreamVisitor();
```
timtebeek, line 52: Probably best to add a quick limiting `Precondition.check` here to only process classes that use the offending classes and methods, for typically better performance. I'd think that `new UsesMethod("java.lang.Optional isPresent()")` could work well here, as that's the one method that's infrequently used, and an essential part of the pattern we want to replace here.

timtebeek (commented):
Impressed with the work you've done here! Really like the attention to detail in handling edge cases, and the suite of tests to show what's correctly handled. I've added some comments that I think would improve this PR still, and bring it closer to a merge.

Let me know if you'd be open to making those changes still!

src/main/java/org/openrewrite/java/migrate/util/OptionalStreamRecipe.java
timtebeek: We allow the use of markdown in our display names and description, which helps readability when we generate the documentation.

src/main/java/org/openrewrite/java/migrate/util/OptionalStreamRecipe.java
```diff
@@ -0,0 +1,133 @@
...
+            .imports("java.util.Optional")
+            .build();
+
+    @Override
+    public J.MethodInvocation visitMethodInvocation(J.MethodInvocation invocation, ExecutionContext ctx) {
+      return getInvocation(invocation, mapMatcher)
+          .flatMap(mapInvocation -> getInvocation(mapInvocation.getSelect(), filterMatcher)
+          .flatMap(filterInvocation -> getReference(mapInvocation.getArguments().get(0), optionalGetMatcher)
+          .flatMap(optionalGetReference -> getReference(filterInvocation.getArguments().get(0), optionalIsPresentMatcher)
+          .flatMap(optionalIsPresentReference -> {
```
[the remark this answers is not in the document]
aslaski2: I introduced chain of flatMaps because a series of nested ifs was very hard to read. I have no problem with reverting this change, but have a look at this change: https://github.com/openrewrite/rewrite-migrate-java/pull/315/commits/8550af0825d893af9ce4cd3dddb60a051f0f112d#diff-eadd116ad853efe772ca5fb5dec23f6417a410e1c5d6aa59b3d498057e335dc8L49
and tell which version do you prefer.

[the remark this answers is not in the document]
timtebeek: I wouldn't necessarily use nested if statements; we can likely use early returns to keep the structure left aligned and relatively easy to follow. Let me know if you'd like to pass that challenge back to me instead; not trying to make you jump through hoops to get this through; we appreciate the work you've put in already! :)

src/main/java/org/openrewrite/java/migrate/util/OptionalStreamRecipe.java
aslaski2: After I added
`return Preconditions.check(new UsesMethod<>("java.lang.Optional isPresent()"), new OptionalStreamVisitor());`
tests fail with
```
java.lang.AssertionError: Expected recipe to complete in 1 cycle, but took 0 cycle.
	at org.openrewrite.test.LargeSourceSetCheckingExpectedCycles.afterCycle(LargeSourceSetCheckingExpectedCycles.java:80)
	at org.openrewrite.RecipeScheduler.scheduleRun(RecipeScheduler.java:88)
	at org.openrewrite.Recipe.run(Recipe.java:279)
        ......
```

timtebeek: Sure! I'm traveling right now, but will try to pick this up Friday; unless @joanvr beats me to it.

timtebeek: Applied just now :)

src/main/java/org/openrewrite/java/migrate/util/OptionalStreamRecipe.java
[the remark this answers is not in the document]
timtebeek: Added this just now with early returns.

src/main/java/org/openrewrite/java/migrate/util/OptionalStreamRecipe.java
```diff
@@ -0,0 +1,97 @@
+/*
+ * Copyright (c) 2023 Atlassian US Inc.
```
timtebeek, line 2:
```suggestion
 * Copyright 2023 the original author or authors.
```

src/test/java/org/openrewrite/java/migrate/util/OptionalStreamRecipeTest.java
```diff
@@ -0,0 +1,432 @@
+/*
+ * Copyright (c) 2023 Atlassian US Inc.
```
timtebeek, line 2:
```suggestion
 * Copyright 2023 the original author or authors.
```

PR #319, opened 2023-10-14
Convert Lombok `@Value` classes to Java records

kevin0x90:
## What's changed?
Adds a recipe for converting lombok value annotated classes to java records.

## What's your motivation?
I am working on a case study to evaluate the effort and benefit of automated refactorings.

@timtebeek 

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/lombok/LombokValueToRecord.java
+++ b/src/main/java/org/openrewrite/java/migrate/lombok/LombokValueToRecord.java
@@ -56,12 +56,13 @@ public class LombokValueToRecord extends ScanningRecipe<Map<String, Set<String>>
 
     private static boolean isRelevantClass(J.ClassDeclaration classDeclaration) {
         return classDeclaration.getType() != null
-               && !isRecord(classDeclaration)
-               && hasOnlyLombokValueAnnotation(classDeclaration)
-               && !hasGenericTypeParameter(classDeclaration)
-               && !hasExplicitMethods(classDeclaration)
-               && !hasExplicitConstructor(classDeclaration)
-               && !hasMemberVariableAnnotations(classDeclaration);
+                && !isRecord(classDeclaration)
+                && hasOnlyLombokValueAnnotation(classDeclaration)
+                && !hasGenericTypeParameter(classDeclaration)
+                && !hasExplicitMethods(classDeclaration)
+                && !hasExplicitConstructor(classDeclaration)
+                && !hasMemberVariableAnnotations(classDeclaration)
+                && !hasStaticMembers(classDeclaration);
     }
 
     @Override
@@ -305,6 +306,14 @@ public class LombokValueToRecord extends ScanningRecipe<Map<String, Set<String>>
                 .map(J.MethodDeclaration::getMethodType)
                 .filter(Objects::nonNull)
                 .anyMatch(JavaType.Method::isConstructor);
+    }
+
+    private static boolean hasStaticMembers(J.ClassDeclaration classDeclaration) {
+        return findAllClassFields(classDeclaration)
+                .map(J.VariableDeclarations::getModifiers)
+                .flatMap(List::stream)
+                .map(J.Modifier::getType)
+                .anyMatch(J.Modifier.Type.Static::equals);
     }
 
     private static boolean hasMemberVariableAnnotations(J.ClassDeclaration classDeclaration) {
--- a/src/main/java/org/openrewrite/java/migrate/lombok/LombokValueToRecord.java
+++ b/src/main/java/org/openrewrite/java/migrate/lombok/LombokValueToRecord.java
@@ -87,13 +87,13 @@ public class LombokValueToRecord extends ScanningRecipe<Map<String, Set<String>>
         return Preconditions.check(check, new ScannerVisitor(acc));
     }
 
-    public String getDisplayName() {
-        return "Convert `@lombok.Value` class to Record";
-    }
-
     @Override
     public TreeVisitor<?, ExecutionContext> getVisitor(Map<String, Set<String>> recordTypesToMembers) {
         return new LombokValueToRecord.LombokValueToRecordVisitor(useExactToString, recordTypesToMembers);
+    }
+
+    public String getDisplayName() {
+        return "Convert `@lombok.Value` class to Record";
     }
 
     private static class LombokValueToRecordVisitor extends JavaIsoVisitor<ExecutionContext> {
--- a/src/main/java/org/openrewrite/java/migrate/lombok/LombokValueToRecord.java
+++ b/src/main/java/org/openrewrite/java/migrate/lombok/LombokValueToRecord.java
@@ -53,7 +53,7 @@ public class LombokValueToRecord extends ScanningRecipe<Map<String, Set<String>>
     boolean useExactToString;
 
     @JsonCreator
-    public LombokValueToRecord(final @JsonProperty("useExactToString") boolean useExactToString) {
+    public LombokValueToRecord(@JsonProperty("useExactToString") boolean useExactToString) {
         this.useExactToString = useExactToString;
     }
 
@@ -73,36 +73,33 @@ public class LombokValueToRecord extends ScanningRecipe<Map<String, Set<String>>
     }
 
     @Override
-    public Map<String, Set<String>> getInitialValue(final ExecutionContext ctx) {
+    public Map<String, Set<String>> getInitialValue(ExecutionContext ctx) {
         return new ConcurrentHashMap<>();
     }
 
     @Override
-    public TreeVisitor<?, ExecutionContext> getScanner(final Map<String, Set<String>> acc) {
-        final TreeVisitor<?, ExecutionContext> check = Preconditions.and(
+    public TreeVisitor<?, ExecutionContext> getScanner(Map<String, Set<String>> acc) {
+        TreeVisitor<?, ExecutionContext> check = Preconditions.and(
                 new UsesJavaVersion<>(17),
                 new MaybeUsesImport<>(LOMBOK_VALUE_IMPORT)
         );
 
         return Preconditions.check(check, new JavaIsoVisitor<ExecutionContext>() {
             @Override
-            public J.ClassDeclaration visitClassDeclaration(
-                    final J.ClassDeclaration classDecl,
-                    final ExecutionContext executionContext
-            ) {
-                final J.ClassDeclaration classDeclaration = super.visitClassDeclaration(classDecl, executionContext);
+            public J.ClassDeclaration visitClassDeclaration(J.ClassDeclaration classDecl, ExecutionContext ctx) {
+                J.ClassDeclaration classDeclaration = super.visitClassDeclaration(classDecl, ctx);
 
                 if (!isRelevantClass(classDeclaration)) {
                     return classDeclaration;
                 }
 
-                final List<J.VariableDeclarations> memberVariables = findAllClassFields(classDeclaration)
+                List<J.VariableDeclarations> memberVariables = findAllClassFields(classDeclaration)
                         .collect(Collectors.toList());
                 if (hasMemberVariableAssignments(memberVariables)) {
                     return classDeclaration;
                 }
 
-                final JavaType.FullyQualified classType = requireNonNull(classDeclaration.getType(),
+                JavaType.FullyQualified classType = requireNonNull(classDeclaration.getType(),
                         "Class type must not be null");
 
                 acc.putIfAbsent(
@@ -115,7 +112,7 @@ public class LombokValueToRecord extends ScanningRecipe<Map<String, Set<String>>
     }
 
     @Override
-    public TreeVisitor<?, ExecutionContext> getVisitor(final Map<String, Set<String>> recordTypesToMembers) {
+    public TreeVisitor<?, ExecutionContext> getVisitor(Map<String, Set<String>> recordTypesToMembers) {
         return new LombokValueToRecord.LombokValueToRecordVisitor(useExactToString, recordTypesToMembers);
     }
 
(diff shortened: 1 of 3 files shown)


timtebeek:
This is looking very promising already! Thanks for getting this started. Could you run `./gradlew lF` to add the license headers? That would make the tests run here.

kevin0x90:
> This is looking very promising already! Thanks for getting this started. Could you run `./gradlew lF` to add the license headers? That would make the tests run here.

Hi,

Sorry for the late reply and thank you for adding the copyright information.

I added more changes to approach the defined acceptability.
During the replacement of the getter methods I ran into some type problems. Maybe you have an idea how to prevent this?

src/main/java/org/openrewrite/java/migrate/lombok/LombokValueToRecord.java
```diff
@@ -0,0 +1,349 @@
...
+    @Override
+    public Set<String> getTags() {
+        return Collections.singleton("lombok");
+    }
+
+    @Override
+    public Duration getEstimatedEffortPerOccurrence() {
+        return Duration.ofMinutes(5);
+    }
+
```
timtebeek, line 86: The default is also five minutes, so no need to override that here.
```suggestion
```

src/main/java/org/openrewrite/java/migrate/lombok/LombokValueToRecord.java
```diff
@@ -0,0 +1,349 @@
...
+    boolean useExactToString;
+
+    @JsonCreator
+    public LombokValueToRecord(final @JsonProperty("useExactToString") boolean useExactToString) {
+        this.useExactToString = useExactToString;
+    }
+
+    @Override
+    public String getDisplayName() {
+        return "Convert Value class to Record";
```
timtebeek, line 69:
```suggestion
        return "Convert `@lombok.Value` class to Record";
```

src/main/java/org/openrewrite/java/migrate/lombok/LombokValueToRecord.java
```diff
@@ -0,0 +1,349 @@
...
+    }
+
+    @Override
+    public String getDescription() {
+        return "Convert Lombok Value annotated classes to standard Java Record classes in Java 17 or higher.";
```
timtebeek, line 74:
```suggestion
        return "Convert Lombok `@Value` annotated classes to standard Java Records.";
```

src/main/java/org/openrewrite/java/migrate/lombok/LombokValueToRecord.java
```diff
@@ -0,0 +1,349 @@
...
+    @Override
+    public boolean causesAnotherCycle() {
+        return true;
+    }
+
+    @Override
+    public int maxCycles() {
+        return 2;
+    }
+
```
timtebeek, line 96: I'm doubtful if we need these overrides; were these added intentionally & necessary for these replacements?
```suggestion
```

src/main/java/org/openrewrite/java/migrate/lombok/LombokValueToRecord.java
```diff
@@ -0,0 +1,349 @@
...
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        final TreeVisitor<?, ExecutionContext> check = Preconditions.and(
+                new UsesJavaVersion<>(17)
```
timtebeek, line 100: Perhaps good to limit recipe execution here early by checking for the presence of `lombok.Value` through something like 
```suggestion
        final TreeVisitor<?, ExecutionContext> check = Preconditions.and(
                new UsesJavaVersion<>(17),
                new UsesType("lombok.Value")
```
Reviewing from GitHub, so the above might still need an import for instance.

src/test/java/org/openrewrite/java/migrate/lombok/LombokValueToRecordTest.java
```diff
@@ -0,0 +1,264 @@
...
+          )
+        );
+    }
+
+    @Test
+    void convertOnlyValueAnnotatedClassWithoutDefaultValuesToRecord() {
+        //language=java
+        rewriteRun(
+          // TODO: find a way to please type validation so this workaround is not required anymore
+          s -> s.typeValidationOptions(TypeValidation.none()),
```
timtebeek, line 76: Probably best to get this resolved before we merge, such that subsequent recipes don't trip over missing types; let us know if you'd need help to resolve the missing types!

src/test/java/org/openrewrite/java/migrate/lombok/LombokValueToRecordTest.java
```diff
@@ -0,0 +1,264 @@
...
+        spec.recipe(new LombokValueToRecord(false))
+          .parser(JavaParser.fromJavaVersion().classpath("lombok"));
+    }
+
+    @Test
+    void valueAnnotatedClassWithUseExactOptionKeepsLombokToString() {
+        //language=java
+        rewriteRun(
+          s -> s.recipe(new LombokValueToRecord(true)),
+          version(
```
timtebeek, line 40: Since (nearly) all tests use this version, we can also set it in the `defaults` method above to simplify each test below a bit.

timtebeek (commented):
Quick first round of feedback; really like that you took this on! Hope to get this closer to a merge soon, and start trying it out on projects.

Did you ever look towards using the Moderne CLI? That might allow you to run it locally across projects quickly, after a first round of building serialized LSTs. Not a requirement, but could give you feedback on how this works on your own projects without the repeated wait for the lossless semantic tree models to be built.

src/main/java/org/openrewrite/java/migrate/lombok/LombokValueToRecord.java
```diff
@@ -0,0 +1,349 @@
...
+        private static boolean hasMemberVariableAssignments(final List<J.VariableDeclarations> memberVariables) {
+            return memberVariables
+                    .stream()
+                    .map(J.VariableDeclarations::getVariables)
+                    .flatMap(List::stream)
+                    .map(J.VariableDeclarations.NamedVariable::getInitializer)
+                    .anyMatch(J.Literal.class::isInstance);
+        }
+
+        private static boolean hasOnlyLombokValueAnnotation(final J.ClassDeclaration cd) {
```
timtebeek, line 341: Should we additionally check if there are any fields annotated with `@With` for instance? Annotations on any field should disqualify the class as well I suppose; not sure if that's already correctly handled, and didn't spot a corresponding case in the (extensive!) tests yet.

src/main/java/org/openrewrite/java/migrate/lombok/LombokValueToRecord.java
```diff
@@ -0,0 +1,349 @@
...
+                            memberVariablesToString(getMemberVariableNames(memberVariables))));
+        }
+
+        private static boolean isRelevantClass(final J.ClassDeclaration classDeclaration) {
+            return !isRecord(classDeclaration)
+                    && hasOnlyLombokValueAnnotation(classDeclaration)
+                    && !hasGenericTypeParameter(classDeclaration)
+                    && !hasExplicitMethods(classDeclaration)
+                    && !hasExplicitConstructor(classDeclaration)
+                    && classDeclaration.getType() != null;
```
timtebeek, line 234: Maybe good to pull this last one up, as it's cheaper to check than some of the others.

src/main/java/org/openrewrite/java/migrate/lombok/LombokValueToRecord.java
kevin0x90: Yes currently these overrides are required since the recipe runs in two phases

- First run: adjust the value classes to records and keep track of the members 
- Second run: adjust the references to getters of the adjusted classes

I haven't found another way to solve the issue of adjusting the references in the same cycle because from what I have read in the documentation the recipe is run in parallel when executed outside of a test environment.

If there is a solution how to solve this with only one run I would be really interested in it.

src/main/java/org/openrewrite/java/migrate/lombok/LombokValueToRecord.java
kevin0x90: I like the suggestion but since I need to run the recipe twice I am not sure if this would break things because from my understanding in the second run the lombok Value classes would be already adjusted and therefore the import would be missing.

I will look into it and adjust if possible.

src/test/java/org/openrewrite/java/migrate/lombok/LombokValueToRecordTest.java
kevin0x90: I would be really happy to get support regarding this. At the moment i suspect in the test somehow even the types are in the same package somehow the environment does not know the adjusted types after the first run.

src/main/java/org/openrewrite/java/migrate/lombok/LombokValueToRecord.java
kevin0x90: Sounds like a good idea. Maybe it would be good to filter for general field annotations related to lombok for this recipe.

src/main/java/org/openrewrite/java/migrate/lombok/LombokValueToRecord.java
knutwannheden: Enter `ScanningRecipe`: https://docs.openrewrite.org/concepts-explanations/recipes#scanning-recipes

Recipes extending this class will run in two phases: The first phase collects data across all sources and the second phase performs the required modifcations.

src/test/java/org/openrewrite/java/migrate/lombok/LombokValueToRecordTest.java
timtebeek: In the `defaults` method you can call `.allSources(src -> src.markers(javaVersion(17)));` on the spec to set that marker on all sources.

src/main/java/org/openrewrite/java/migrate/lombok/LombokValueToRecord.java
timtebeek: I think just any field annotation would disqualify the conversion here; I'm not sure we'd correctly handle Jackson annotations either for instance, so I'd like to start out with the simplest case only: class with only fields and only a single annotation in total, such that we don't accidentally break any code with this recipe. We can always (safely) expand the scope of what we cover later; we can't easily recover trust when we break things. :)

src/main/java/org/openrewrite/java/migrate/lombok/LombokValueToRecord.java
kevin0x90: Thanks for the hint, I refactored the implementation so that a ScanningRecipe is used.

src/main/java/org/openrewrite/java/migrate/lombok/LombokValueToRecord.java
kevin0x90: Addressed with the refactoring to ScanningRecipe

kevin0x90:
> Quick first round of feedback; really like that you took this on! Hope to get this closer to a merge soon, and start trying it out on projects.
> 
> Did you ever look towards using the Moderne CLI? That might allow you to run it locally across projects quickly, after a first round of building serialized LSTs. Not a requirement, but could give you feedback on how this works on your own projects without the repeated wait for the lossless semantic tree models to be built.

I did not yet try out the Moderne CLI but will give it a try for my sample project.

src/main/java/org/openrewrite/java/migrate/lombok/LombokValueToRecord.java
```diff
@@ -0,0 +1,356 @@
...
+    @Override
+    public Map<String, Set<String>> getInitialValue(final ExecutionContext ctx) {
+        return new ConcurrentHashMap<>();
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getScanner(final Map<String, Set<String>> acc) {
+        final TreeVisitor<?, ExecutionContext> check = Preconditions.and(
+                new UsesJavaVersion<>(17),
+                new MaybeUsesImport<>(LOMBOK_VALUE_IMPORT)
```
timtebeek, line 84: Rather than checking the import we can instead check for `UsesType`. Otherwise we might fail to pick up qualified use through `@lombok.Value`.

src/test/java/org/openrewrite/java/migrate/lombok/LombokValueToRecordTest.java
timtebeek: It seems you've worked through most of the review feedback, such that there's less chance for conflicts if we were to step in to help out here. Do you still need help? Typically I'd step through with the debugger and try things that way, but not sure yet as I haven't looked

src/main/java/org/openrewrite/java/migrate/lombok/LombokValueToRecord.java
```diff
@@ -0,0 +1,356 @@
...
+        );
+
+        return Preconditions.check(check, new JavaIsoVisitor<ExecutionContext>() {
+            @Override
+            public J.ClassDeclaration visitClassDeclaration(
+                    final J.ClassDeclaration classDecl,
+                    final ExecutionContext executionContext
+            ) {
+                final J.ClassDeclaration classDeclaration = super.visitClassDeclaration(classDecl, executionContext);
```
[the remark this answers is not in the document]
kevin0x90: Ok, will adjust

timtebeek:
Finally found some more time to review; haven't yet gotten to the missing type information, but did find some simplifications and a case not yet covered around static fields. Not quite sure what records allow there, but might be safest to just not convert those cases. I've also grouped the tests into a nested class, such that it's easier to spot what should and should not be supported.

Let me know if you'd like to fix that failing test yourself, or if you'd at any point want to hand this over. :)

kevin0x90:
> Finally found some more time to review; haven't yet gotten to the missing type information, but did find some simplifications and a case not yet covered around static fields. Not quite sure what records allow there, but might be safest to just not convert those cases. I've also grouped the tests into a nested class, such that it's easier to spot what should and should not be supported.
> 
> Let me know if you'd like to fix that failing test yourself, or if you'd at any point want to hand this over. :)

Hi,

I will fix the failing test and keep on working on this branch whenever I find the time to do so.

src/test/java/org/openrewrite/java/migrate/lombok/LombokValueToRecordTest.java
timtebeek: So the type information is already present on the input, and we're reusing the method invocations from before. As such I don't think we need to hold up this PR over missing type information, as a next run of OpenRewrite using the same sources would actually work. We can leave the missing type information as a nice to have for a future enhancement, with limited use.

timtebeek (approved):
Looks good to me! Thanks again.

Any last minute feedback @knutwannheden or @joanvr ?

knutwannheden:
I can take a look.

src/main/java/org/openrewrite/java/migrate/lombok/LombokValueToRecord.java
```diff
@@ -0,0 +1,308 @@
...
+
+        private static String memberVariablesToString(Set<String> memberVariables) {
+            return memberVariables
+                    .stream()
+                    .map(member -> String.format(TO_STRING_MEMBER_LINE_PATTERN, member, member))
+                    .collect(Collectors.joining(TO_STRING_MEMBER_DELIMITER));
+        }
+
+        private static JavaType.Class buildRecordType(J.ClassDeclaration classDeclaration) {
+            requireNonNull(classDeclaration.getType(), "Class type must not be null");
```
knutwannheden, line 252: Rather than throwing an exception, it would be better if we just made sure, that the recipe doesn't do anything. Typically, a class should indeed by correctly type attributed. But when this recipe is run together with others, we can't always have too high expectations of what the other recipes do, so it could happen that another recipe gets that wrong. So that this gets detected during development, we could add an `assert` statement instead.

src/main/java/org/openrewrite/java/migrate/lombok/LombokValueToRecord.java
```diff
@@ -0,0 +1,308 @@
...
+@EqualsAndHashCode(callSuper = false)
+public class LombokValueToRecord extends ScanningRecipe<Map<String, Set<String>>> {
+
+    private static final AnnotationMatcher LOMBOK_VALUE_MATCHER = new AnnotationMatcher("@lombok.Value");
+
+    @Option(displayName = "useExactToString",
+            description = "When set the `toString` format from Lombok is used in the migrated record.")
+    boolean useExactToString;
+
+    @JsonCreator
```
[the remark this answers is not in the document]
kevin0x90: removed

src/main/java/org/openrewrite/java/migrate/lombok/LombokValueToRecord.java
kevin0x90: addressed

src/main/java/org/openrewrite/java/migrate/lombok/LombokValueToRecord.java
```diff
@@ -0,0 +1,308 @@
...
+import static java.util.Objects.requireNonNull;
+import static java.util.stream.Collectors.toList;
+
+@Value
+@EqualsAndHashCode(callSuper = false)
+public class LombokValueToRecord extends ScanningRecipe<Map<String, Set<String>>> {
+
+    private static final AnnotationMatcher LOMBOK_VALUE_MATCHER = new AnnotationMatcher("@lombok.Value");
+
+    @Option(displayName = "useExactToString",
```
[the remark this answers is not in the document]
kevin0x90: addressed

src/main/java/org/openrewrite/java/migrate/lombok/LombokValueToRecord.java
```diff
@@ -0,0 +1,308 @@
...
+            description = "When set the `toString` format from Lombok is used in the migrated record.")
```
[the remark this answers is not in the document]
kevin0x90: from my perspective this would be fine to make the option non-required but personally I would prefer to keep a primitive boolean that is initialized from the default constructor to false so basically like:
```
@Option(displayName = "Add a `toString()` implementation matching Lombok",
            description = "When set the `toString` format from Lombok is used in the migrated record.",
            required = false)
boolean useExactToString;

public LombokValueToRecord() {
        this(false);
}

public LombokValueToRecord(boolean useExactToString) {
        this.useExactToString = useExactToString;
}
```

@knutwannheden @timtebeek  what do you think?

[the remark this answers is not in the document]
timtebeek: There's some serialization that factors in here as well, where we can't be certain the default constructor is called. In other places having a Boolean field then allows us to differentiate, and perhaps using that uniformly would be best then here as well.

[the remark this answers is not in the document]
knutwannheden: While it may seem a bit odd, I also think it is more important that we stay consistent with respect to other recipes. In the future we may be able to find a better solution for this. But as Tim says, it has to do with the serialization in the SaaS platform.

timtebeek:
Thanks a lot @kevin0x90 ! You can try out this change through [our snapshot versions](https://docs.openrewrite.org/reference/snapshot-instructions), pending the next release.

I'll deploy it to https://app.moderne.io as well, and see how it performs against our own code base as well.

timtebeek:
Ran it on a first few projects; think we might need some minor tweaks and polish. Example diff below
image

Two quick findings:
1. the fields converted to constructor args lack the corresponding indentation
2. we remove the `@Value` annotation anywhere in the compilation unit, not just on the record

timtebeek:
Follow up now in:
- https://github.com/openrewrite/rewrite-migrate-java/pull/334

PR #324, opened 2023-10-21
fix: remove unused variable

SimonVerhoeven:
## What's changed?

I removed an unused variable that was seemingly a remnant from when the new JavaTemplate was introduced

## What's your motivation?

To help maintain the codebase.

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/util/UseEnumSetOf.java
+++ b/src/main/java/org/openrewrite/java/migrate/util/UseEnumSetOf.java
@@ -29,7 +29,6 @@ import org.openrewrite.java.tree.TypeUtils;
 
 import java.time.Duration;
 import java.util.List;
-import java.util.StringJoiner;
 
 public class UseEnumSetOf extends Recipe {
     private static final MethodMatcher SET_OF = new MethodMatcher("java.util.Set of(..)", true);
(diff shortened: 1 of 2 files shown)


src/main/java/org/openrewrite/java/migrate/util/UseEnumSetOf.java
```diff
@@ -66,9 +65,7 @@ public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, Execu
                         if (isAssignmentSetOfEnum(type)) {
                             maybeAddImport("java.util.EnumSet");
 
-                            StringJoiner setOf = new StringJoiner(", ", "EnumSet.of(", ")");
                             List<Expression> args = m.getArguments();
-                            args.forEach(o -> setOf.add("#{any()}"));
```
knutwannheden, line 71: White the variable is indeed unused here, it seems like this currently doesn't work quite right when there are multiple parameters passed to the `Set#of()` method. It looks like a fix was this was started with this `args` variable. By using this `args.toString()` as the template, the recipe should also work for multiple parameters.

SimonVerhoeven: Ah ok, I've updated the test to use multiple parameters, and adapted the recipe

knutwannheden (approved):

PR #330, opened 2023-10-22
chore: remove unused imports

SimonVerhoeven:
## What's changed?

Removed unused imports

## What's your motivation?

Maintenance

### Checklist


timtebeek (approved):

timtebeek:
Thanks again! Guess we should use the platform ourselves more often 😅

PR #332, opened 2023-10-27
Add recipes to migrate boxed primitive constructors deprecated in Java 9

dsibilio:
## What's changed?

Added a set of Refaster recipes to deal with boxed primitive constructors that were [deprecated in Java 9](https://docs.oracle.com/javase%2F9%2Fdocs%2Fapi%2F%2F/deprecated-list.html#constructor).

## What's your motivation?

https://rewriteoss.slack.com/archives/C01A843MWG5/p1698430647117759?thread_ts=1698429230.444949&cid=C01A843MWG5

Unless I missed something, I think I covered all deprecated constructors **except** `new Float(double) -> Float.valueOf((float) double)`. Due to how Refaster generates the code, I end up in a situation like the following:

**Refaster Template**
```java
@RecipeDescriptor(
    name = "Replace deprecated `Float(double)` constructor invocations",
    description = "Replace deprecated `Float(double)` constructor invocations with `Float.valueOf((float) double)`."
)
public static class FloatDoubleConstructor {
    @BeforeTemplate
    public Float doubleConstructor(double value) {
        return new Float(value);
    }

    @AfterTemplate
    public Float valueOf(double value) {
        return Float.valueOf((float) value);
    }
}
```

**Generated Code Error**
```java
error: incompatible types: Double cannot be converted to float
                final JavaTemplate valueOf = Semantics.expression(this, "valueOf", (@Primitive Double value) -> Float.valueOf((float)value)).build();
```

Therefore I was only able to cover `new Float(Double) -> new Float(Double.floatValue())` instead.

Is there any way to force a real primitive in this line via Refaster _(or use e.g.: a String-format template somehow)_?
Do we need to make a separate imperative recipe just to cover the `Float(double)` edge case?

@timtebeek 🙏 

We could use imperative recipes but the change seems simple enough that the Refaster usage seems to be justified and seems to avoid a lot of repeated boilerplate code.

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/lang/MigrateDeprecatedBoxedPrimitiveConstructors.java
+++ b/src/main/java/org/openrewrite/java/migrate/lang/MigrateDeprecatedBoxedPrimitiveConstructors.java
@@ -1,0 +1,285 @@
+/*
+ * Copyright 2023 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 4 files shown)


timtebeek (approved):
Appreciate the quick turn around on this & thanks for contributing.

I don't know of a way to force the case for `new Float(double)` using Refaster; I think we can leave that out for now; it should be uncommon, and perhaps not worth the effort of a separate recipe.

Really like how you cover every case with a single test, and how you took care to match our expected formatting.

PR #333, opened 2023-10-27
Revert "Add recipes to migrate boxed primitive constructors deprecate…

dsibilio:
…d in Java 9 (#332)"

This reverts commit 59f1bb523e72335f7a69fd5a0c35e6d317c4deb3.

## What's your motivation?

While dealing with the `new Float(double)` edge case I actually found out there already was an OpenRewrite recipe in the Java 8 to 11 flow to deal with these deprecated constructors! lol
This makes the new implementation with Refaster unnecessary in the first place.
It was hidden in a separate project which is why I didn't find it at first: https://github.com/openrewrite/rewrite-static-analysis/blob/main/src/main/java/org/openrewrite/staticanalysis/PrimitiveWrapperClassConstructorToValueOf.java & https://github.com/openrewrite/rewrite-migrate-java/blob/main/src/main/resources/META-INF/rewrite/java-version-11.yml#L40.

@timtebeek at least it was good practice and it was a good opportunity for me to familiarize with Refaster, thanks!


timtebeek:
Appreciate the discipline to then revert your own change again! I mist say I slightly prefer the Refaster recipe approach; it's clear and easy to maintain. But that's not enough to maintain two copies, or ask folks to switch between implementations. Thanks again for putting the effort in; when searching earlier I had searched for Integer, and then the other recipe implementation did not pop up. :/

Issue #141, opened 2022-10-30
Migrate from @lombok.Value to Java records


timtebeek:
Convert the simplest cases of [Lombok value classes](https://projectlombok.org/features/Value) into Java Records.
Narrowly scoped initially, as there's a lot of sometimes experimental Lombok annotations that can complicate matters.

Initial classes should:
- not have any other Lombok annotations
- not have any fields with Lombok annotations
- not have an explicit constructor
- not have explicit methods
- not use generics

The conversion should:
- create a record with the same fields, in the same order
- replace any class getter calls with record accessor calls
- remove the lombok.Value annotation & import

The conversion could;
- have a parameter to create an exact `.toString()` replacement

Ideally, given the limited scope, the implementation should not suffer from a lack of Lombok support in the parser. A one-to-one replacement for these simplest of cases should mostly behave the same, with the exception of the `.toString()` method by default.

Similar to #116 & #124.

PR #334, opened 2023-10-31
Polish LombokValueToRecord

timtebeek:
## What's changed?
- Only remove annotations from record class.
- Do not retain final or static modifiers on records.
- Do not convert non static inner classes.

## Any additional context
- #141

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/lombok/LombokValueToRecord.java
+++ b/src/main/java/org/openrewrite/java/migrate/lombok/LombokValueToRecord.java
@@ -19,6 +19,7 @@ import lombok.EqualsAndHashCode;
 import lombok.RequiredArgsConstructor;
 import lombok.Value;
 import org.openrewrite.*;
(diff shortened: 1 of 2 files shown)


src/main/java/org/openrewrite/java/migrate/lombok/LombokValueToRecord.java
```diff
@@ -42,7 +40,7 @@
 @EqualsAndHashCode(callSuper = false)
 public class LombokValueToRecord extends ScanningRecipe<Map<String, Set<String>>> {
 
-    private static final AnnotationMatcher LOMBOK_VALUE_MATCHER = new AnnotationMatcher("@lombok.Value");
+    private static final AnnotationMatcher LOMBOK_VALUE_MATCHER = new AnnotationMatcher("@lombok.Value()");
```
timtebeek, line 43: This was an attempt to have the annotation matcher only match annotations without arguments; that didn't work, but I think it should.

PR #335, opened 2023-10-31
Adds MXbean recipe

cjobinabo:
## What's changed?
Screenshot 2023-10-31 at 2 03 38 PM

@timtebeek @joanvr

### Checklist


src/test/java/org/openrewrite/java/migrate/MXBeanNonPublicTest.java
```diff
@@ -0,0 +1,304 @@
...
+              @MXBean(true)
+              interface Foo {
+              }
+              """,
+            """
+              import javax.management.MXBean;
+                             
+              @MXBean(true)
+              public
+              interface Foo {
```
timtebeek, line 138: Not quite sure where this extra newline comes from; ideally we have that removed before a merge, but since this will likely only affect a small number of interfaces I can also see that as a future improvement.

src/test/java/org/openrewrite/java/migrate/MXBeanNonPublicTest.java
```diff
@@ -0,0 +1,304 @@
...
+                  }
+              
+                  @MXBean
+                  public interface NonPublicMBean {
+                  }
+              }
+              """
+          )
+        );
+    }
```
timtebeek, line 95: What are we really testing here? seems overly complex for the one line we aim to change here. I'm inclined to fully remove this test.
```suggestion
```

src/test/java/org/openrewrite/java/migrate/MXBeanNonPublicTest.java
```diff
@@ -0,0 +1,304 @@
...
+    void mxbeanSuffix() {
+        rewriteRun(
+          //language=java
+          java(
+            """
+              interface TestNonPublicInterfaceMXBean {
+              }
+              """,
+            """
+              public interface TestNonPublicInterfaceMXBean {
```
timtebeek, line 192:
```suggestion
              interface SuffixMXBean {
              }
              """,
            """
              public interface SuffixMXBean {
```

src/test/java/org/openrewrite/java/migrate/MXBeanNonPublicTest.java
```diff
@@ -0,0 +1,304 @@
...
+    void mbeanSuffix() {
+        rewriteRun(
+          //language=java
+          java(
+            """
+              interface TestNonPublicInterfaceMBean {
+              }
+              """,
+            """
+              public interface TestNonPublicInterfaceMBean {
```
timtebeek, line 175:
```suggestion
              interface SuffixMBean {
              }
              """,
            """
              public interface SuffixMBean {
```

src/test/java/org/openrewrite/java/migrate/MXBeanNonPublicTest.java
cjobinabo: Haha yeah, I wasn't able to determine the cause. I was hoping you might know.

src/test/java/org/openrewrite/java/migrate/MXBeanNonPublicTest.java
cjobinabo: This test captures the sub class scenario. The test was a port of our test from our eclipse plugin. The test likely captures scenarios that our other migration rules depended on. I could trim it down if you want.

timtebeek: Trimming it down would help; now it's a bit unclear what elements are and aren't necessary for what we're testing. Minimal tests are also easier to step through with the debugger if there's no excess classes, methods or invocations.

src/test/java/org/openrewrite/java/migrate/MXBeanNonPublicTest.java
timtebeek: Yeah I tried for a bit to figure out a way to fix it; but in the end thought it not worth spending considerable amount of time on, if we're not even sure it replicates beyond this test.

src/test/java/org/openrewrite/java/migrate/MXBeanNonPublicTest.java
cjobinabo: No problem. I'll have that update up soon.

timtebeek (approved):

Issue #243, opened 2023-06-26
Migrate to sequenced collections in Java 21 / JEP 431


timtebeek:
## What problem are you trying to solve?
Java 21 will [introduce sequenced collections](https://openjdk.org/jeps/431), allowing us to potentially migrate away from some outdated code patterns.

## What precondition(s) should be checked before applying this recipe?
Java 21 in use, and collection is a sequenced collection.

## Describe the situation before applying the recipe

```java
class A {
    void foo(List<String> list) {
        String first = list.get(0);
        String last = list.get(list.size() -1);
    }
    void foo(Deque<String> deque) {
        String first = deque.getFirst();
        String last = deque.getLast();
    }
    void foo(SortedSet<String> set) {
        String first = set.first();
        String last = set.last();
    }
    void foo(LinkedHashSet<String> set) {
        String first = set.iterator().next();
        String last; // missing
    }
}
```

## Describe the situation after applying the recipe
```java
class A {
    void foo(List<String> list) {
        String first = list.getFirst();
        String last = list.getLast();
    }
    void foo(Deque<String> deque) {
        String first = deque.getFirst();
        String last = deque.getLast();
    }
    void foo(SortedSet<String> set) {
        String first = set.getFirst();
        String last = set.getLast();
    }
    void foo(LinkedHashSet<String> set) {
        String first = set.getFirst();
        String last; // missing
    }
}
```

## Have you considered any alternatives or workarounds?
No.

## Any additional context
While Java 21 is not yet officially released or supported, we can already define minimized versions of the respective interfaces in our unit tests, such that we can already develop and test these with relative ease, even when running on lower Java versions.

timtebeek:
@nipafx figured you might enjoy an issue such as this one, after our discussion over dinner at JCON. Relatively speaking it's an easy issue to start with, especially when focused on a single use case, while still helping users migrate away from older code patterns onto Java 21, in time with the release.

timtebeek:
This has been mostly implemented in
- https://github.com/openrewrite/rewrite-migrate-java/pull/339

Any further changes can be logged in a new issue, or contributed directly as a draft PR containing a failing test, or even a full PR containing said recipe implementation.

PR #339, opened 2023-11-03
Adopt SequencedCollection First and Last methods

timtebeek:
## What's changed?
Additional recipes to adopt first/last methods in SequencedCollections.

## What's your motivation?
Move away from old code patterns; show benefit of upgrade to Java 21.

## Any additional context
- https://github.com/openrewrite/rewrite-migrate-java/issues/243

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/util/IteratorNext.java
+++ b/src/main/java/org/openrewrite/java/migrate/util/IteratorNext.java
@@ -54,16 +54,17 @@ public class IteratorNext extends Recipe {
                 new JavaIsoVisitor<ExecutionContext>() {
                     @Override
                     public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, ExecutionContext executionContext) {
-                        J.MethodInvocation mi = super.visitMethodInvocation(method, executionContext);
-                        if (NEXT_MATCHER.matches(mi) && ITERATOR_MATCHER.matches(mi.getSelect())) {
-                            Expression select = ((J.MethodInvocation) mi.getSelect()).getSelect();
-                            if (TypeUtils.isAssignableTo("java.util.SequencedCollection", select.getType())) {
-                                return mi.withSelect(select)
-                                        .withName(mi.getName().withSimpleName("getFirst"))
-                                        .withMethodType(mi.getMethodType().withName("getFirst"));
+                        J.MethodInvocation nextInvocation = super.visitMethodInvocation(method, executionContext);
+                        if (NEXT_MATCHER.matches(nextInvocation) && ITERATOR_MATCHER.matches(nextInvocation.getSelect())) {
+                            J.MethodInvocation iteratorInvocation = (J.MethodInvocation) nextInvocation.getSelect();
+                            if (TypeUtils.isAssignableTo("java.util.SequencedCollection", iteratorInvocation.getSelect().getType())) {
+                                J.MethodInvocation getFirstInvocation = nextInvocation.withSelect(iteratorInvocation.getSelect())
+                                        .withName(nextInvocation.getName().withSimpleName("getFirst"))
+                                        .withMethodType(iteratorInvocation.getMethodType().withName("getFirst"));
+                                return getFirstInvocation;
                             }
                         }
-                        return mi;
+                        return nextInvocation;
                     }
                 }
         );
--- a/src/main/java/org/openrewrite/java/migrate/util/IteratorNext.java
+++ b/src/main/java/org/openrewrite/java/migrate/util/IteratorNext.java
@@ -60,7 +60,9 @@ public class IteratorNext extends Recipe {
                             if (TypeUtils.isAssignableTo("java.util.SequencedCollection", iteratorInvocation.getSelect().getType())) {
                                 J.MethodInvocation getFirstInvocation = nextInvocation.withSelect(iteratorInvocation.getSelect())
                                         .withName(nextInvocation.getName().withSimpleName("getFirst"))
-                                        .withMethodType(iteratorInvocation.getMethodType().withName("getFirst"));
+                                        .withMethodType(nextInvocation.getMethodType()
+                                                .withName("getFirst")
+                                                .withDeclaringType(iteratorInvocation.getMethodType().getDeclaringType()));
                                 return getFirstInvocation;
                             }
                         }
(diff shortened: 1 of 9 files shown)


knutwannheden:
What is the reason for not using Refaster for this? Is it because it would require a Java 21 dependency? Possibly the Refaster template could be inside a separate source folder with Java 21 source compatibility?

Alternatively, it could still be interesting to use a JavaTemplate compiled from a lambda to do the matching. Although that reminds me that we probably still have unsolved issues with generics.

timtebeek:
Generics and the Java 21 dependency indeed were the reasons I didn't want to start out with Refaster. There's likely also still some gradle API tooling to sort still, and didn't want to be blocked on multiple fronts for these replacements. We could always reevaluate when support improves, but for now if thought to opt for something folks can already use.

src/main/java/org/openrewrite/java/migrate/util/IteratorNext.java
```diff
@@ -0,0 +1,71 @@
...
+                new JavaIsoVisitor<ExecutionContext>() {
+                    @Override
+                    public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, ExecutionContext executionContext) {
+                        J.MethodInvocation mi = super.visitMethodInvocation(method, executionContext);
+                        if (NEXT_MATCHER.matches(mi) && ITERATOR_MATCHER.matches(mi.getSelect())) {
+                            Expression select = ((J.MethodInvocation) mi.getSelect()).getSelect();
+                            if (TypeUtils.isAssignableTo("java.util.SequencedCollection", select.getType())) {
+                                return mi.withSelect(select)
+                                        .withName(mi.getName().withSimpleName("getFirst"))
+                                        .withMethodType(mi.getMethodType().withName("getFirst"));
```
knutwannheden, line 63: I suspect that the prefix may get lost in this recipe.

Also, the type attribution will be slightly off here, because the method type will have `Iterator` as its declaring type.

timtebeek: Hmm, I had lifted this snippet from the Simplify method chain recipe, but I'll double check, thanks!

timtebeek: I've fixed the method type in 62be57c (I think); thanks for pointing that out! Here's a debugger view of the new situation 
image

timtebeek: I'm wondering what the best approach would be for prefix handling; I don't really expect anyone to use comments between `.iterator()` and `.next()`. They might each be on a new line, but even then it seems more logical to collapse and not retain the whitespace or comments that were there before.

For reference: here's a test that asserts the _current_ behaviour.
```java
@Test
void retainPrefix() {
    rewriteRun(
      //language=java
      java(
        """
          import java.util.*;
              
          class Foo {
              void bar(List<String> collection) {
                  String first =
                    // Collection prefix
                    collection
                    // Iterator prefix
                    .iterator()
                    // Next prefix
                    .next();
              }
          }
          """,
        """
          import java.util.*;
                        
          class Foo {
              void bar(List<String> collection) {
                  String first =
                    // Collection prefix
                    collection
                    // Next prefix
                    .getFirst();
              }
          }
          """
      )
    );
}
```

knutwannheden: I am afraid the type attribution is still not quite correct. Now the method's return type is `Iterator`. I think it would be best if we could either search the owner type for the method (possibly involves searching the inheritance hierarchy upwards) or modify the type so that all its fields have the expected values (owner, name, type, parameters.

Regarding the prefix I will take a look tomorrow. Maybe the problem is just hypothetical.

knutwannheden: Since we are updating the `next()` invocation I believe this code is correct. As with many recipes we have situations where comments will be removed. So we could of course just not update the code if there are any comment that would end up getting removed. But that is maybe overkill.

src/test/java/org/openrewrite/java/migrate/util/ListFirstAndLastTest.java
```diff
@@ -0,0 +1,293 @@
...
+
+
+@Issue("https://github.com/openrewrite/rewrite-migrate-java/issues/243")
+@EnabledForJreRange(min = JRE.JAVA_21)
+class ListFirstAndLastTest implements RewriteTest {
+
+    @Override
+    public void defaults(RecipeSpec spec) {
+        spec.recipe(new ListFirstAndLast())
+          .afterTypeValidationOptions(TypeValidation.builder().methodInvocations(false).build())
```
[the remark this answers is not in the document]
timtebeek: Thanks! That had been bothering me, but didn't quite know what was necessary to resolve this.

src/main/java/org/openrewrite/java/migrate/util/ListFirstAndLast.java
```diff
@@ -0,0 +1,123 @@
...
+                firstOrLast = "Last";
+            } else {
+                return mi;
+            }
+
+            List<Object> arguments = new ArrayList<>();
+            if ("add".equals(operation)) {
+                arguments.add(mi.getArguments().get(1));
+            }
+            return JavaTemplate.builder(operation + firstOrLast + (arguments.isEmpty() ? "()" : "(#{})")).build()
```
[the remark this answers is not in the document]
timtebeek: Thanks! Picked up in https://github.com/openrewrite/rewrite-migrate-java/pull/339/commits/8f420059927a9d61bf56f37097441cab960b1ba2

knutwannheden (approved):

src/main/java/org/openrewrite/java/migrate/util/IteratorNext.java
timtebeek: I liked the suggestion to not make changes when there's a risk of comments being lost; tried for a bit to get that to work, but [in my tests](https://github.com/openrewrite/rewrite-migrate-java/pull/339/commits/7e73ebb882d59cfec0edb48bd43dd8cca12c60cc) the comment on the iterator call is in the `JRightPadded.after` of the select before it, and that makes it inaccessible for now (at least from my understanding). Given that it's improbable I figured document it, such that we can show it was a conscious limitation with likely little chance of negative consequences.

PR #341, opened 2023-11-07
com.sun.net.ssl recipe

cjobinabo:
## What's changed?
Screenshot 2023-11-06 at 10 06 54 PM

It's not immediately clear why tests started failing. sun.interbal.jar has all the necessary dependencies.

@timtebeek @joanvr

### Checklist


timtebeek (approved):
Thanks again! I've taken the liberty to split the tests into separate ones per Yaml recipe, such that it's clearer going forward which recipes contribute what changes.

PR #342, opened 2023-11-08
feat: upgrade lombok

SimonVerhoeven:
## What's changed?

Also upgrade Lombok if present during the upgrade.

## What's your motivation?

Versions before 1.18.30 don't quite play along nicely with Java 21.

-

## Any additional context

-

### Checklist

### DIFF
--- a/src/main/resources/META-INF/rewrite/java-version-21.yml
+++ b/src/main/resources/META-INF/rewrite/java-version-21.yml
@@ -30,6 +30,11 @@ recipeList:
   - org.openrewrite.staticanalysis.ReplaceDeprecatedRuntimeExecMethods
   - org.openrewrite.java.migrate.util.SequencedCollection
   - org.openrewrite.github.SetupJavaUpgradeJavaVersion
+  - org.openrewrite.java.dependencies.UpgradeDependencyVersion:
+      groupId: org.projectlombok
+      artifactId: lombok
+      newVersion: 1.18.30
+      overrideManagedVersion: false
 
 ---
 type: specs.openrewrite.org/v1beta/recipe


timtebeek:
Thanks for the suggestion! I'd expect this to already be covered by the `UpdateLombokToJava11` recipe though.
https://github.com/openrewrite/rewrite-migrate-java/blob/e0ee82b4de937412c5c41e3dbae3f43d76074f22/src/main/resources/META-INF/rewrite/lombok.yml#L25-L28
That recipe is run as part of our Java 11 migration.
https://github.com/openrewrite/rewrite-migrate-java/blob/e0ee82b4de937412c5c41e3dbae3f43d76074f22/src/main/resources/META-INF/rewrite/java-version-11.yml#L45
Which in turn in run by our Java 17 migration.
https://github.com/openrewrite/rewrite-migrate-java/blob/e0ee82b4de937412c5c41e3dbae3f43d76074f22/src/main/resources/META-INF/rewrite/java-version-17.yml#L30
Which finally is run by our Java 21 migration.
https://github.com/openrewrite/rewrite-migrate-java/blob/e0ee82b4de937412c5c41e3dbae3f43d76074f22/src/main/resources/META-INF/rewrite/java-version-21.yml#L27

Are you not seeing the Lombok version bumped when running `UpgradeToJava21` without this change?

SimonVerhoeven:
Not when I tried it on a project last week that was using 1.8.28 (I forgot to note the openrewrite version, sorry). 
With this it works as expected, I dove down a bit for the java 11 => Guava one and that marked for 1.8.* but for 21 we need at least 1.18.30

PR #346, opened 2023-11-09
Recipes to migrate to Java EE 7, Java EE 8 & Jakarta EE 9

ranuradh:
## What's changed?
This PR contains 4 recipes:
**UpdateApacheWSSecurityPackages**
image

**OpenJPAPersistenceProvider:**
image

**ServletIsRequestedSessionIdFromURL**
image

**ApacheDefaultProvider**
image

@timtebeek 
### Checklist


src/test/java/org/openrewrite/java/migrate/jakarta/UpdateApacheWSSecurityPackagesTest.java
```diff
@@ -0,0 +1,71 @@
...
+            public class Test {
+                public static void main(String[] args)  {
+            
+                    WSSecurityException x = new WSSecurityException(1,"");
+            
+                    CryptoBase base = null;
+            
+                }
+            }
+            """));
```
timtebeek, line 69: Can you scrub your tests of anything that's not needed, and format them to conform to the other tests we have?
- text blocks start on a new line
- no package declaration when not necessary
- no unnecessary blank lines
- opening and closing braces each on a new line
- no unused method modifiers (`public static`)
- or unused arguments `(String[] args)`

Having a consistent style helps readability, and not having to read unnecessary elements helps focus on what matters only. Otherwise it might seem something is significant ( such as here `public static void main(String[] args)`), when in reality it's not used at all.

Ideally we automate a lot of this type of feedback, but until then I'd prefer not to have to update these by hand myself. 

```suggestion
          java(
            """
            import org.apache.ws.security.WSSecurityException;
            import org.apache.ws.security.components.crypto.CryptoBase;
            class Foo {
                void bar()  {
                    WSSecurityException x = new WSSecurityException(1,"");
                    CryptoBase base = null;
                }
            }
            """,
            """
            import org.apache.wss4j.common.ext.WSSecurityException;
            import org.apache.wss4j.common.ext.components.crypto.CryptoBase;
            class Foo {
                void bar()  {
                    WSSecurityException x = new WSSecurityException(1,"");
                    CryptoBase base = null;
                }
            }
            """
          )
        );
```

src/main/resources/META-INF/rewrite/java-ee-8.yml
```diff
@@ -0,0 +1,61 @@
...
+      elementName: //message-interpolator
+      newValue: org.hibernate.validator.messageinterpolation.ResourceBundleMessageInterpolator
+  - org.openrewrite.xml.ChangeTagValue:
+      elementName: //traversable-resolver
+      newValue: org.hibernate.validator.engine.resolver.DefaultTraversableResolver
+  - org.openrewrite.xml.ChangeTagValue:
+      elementName: //constraint-validator-factory
+      newValue: org.hibernate.validator.engine.ConstraintValidatorFactoryImpl
+  - org.openrewrite.xml.RemoveXmlTag:
+      xPath: //parameter-name-provider
```
timtebeek, line 61: Same as above: please use the full xpath expressions such that we can more safely run against projects that might contain other XML files with similarly named elements.

src/main/resources/META-INF/rewrite/java-ee-8.yml
```diff
@@ -0,0 +1,61 @@
...
+  - org.openrewrite.java.migrate.javaee8.ApacheDefaultProvider
+
+---
+type: specs.openrewrite.org/v1beta/recipe
+name: org.openrewrite.java.migrate.javaee8.ServletIsRequestedSessionIdFromURL
+displayName: Replace `HttpServletRequestWrapper.isRequestedSessionIdFromUrl()` with `HttpServletRequestWrapper.isRequestedSessionIdFromURL()`
+description: The  method `HttpServletRequestWrapper.isRequestedSessionIdFromUrl()` is deprecated in JavaEE8 and is replaced by `HttpServletRequestWrapper.isRequestedSessionIdFromURL()`
+recipeList:
+  - org.openrewrite.java.ChangeMethodName:
+      methodPattern: javax.servlet.http.HttpServletRequestWrapper isRequestedSessionIdFromUrl(*)
```
timtebeek, line 36: Why the wildcard to match arguments here? Are there ever any arguments to match? Should we match multiple methods? If not, then best to drop the wildcard
```suggestion
      methodPattern: javax.servlet.http.HttpServletRequestWrapper isRequestedSessionIdFromUrl()
```

src/main/resources/META-INF/rewrite/java-ee-8.yml
```diff
@@ -0,0 +1,61 @@
...
+      newMethodName: isRequestedSessionIdFromURL
+      ignoreDefinition: true
+---
+type: specs.openrewrite.org/v1beta/recipe
+name: org.openrewrite.java.migrate.javaee8.ApacheDefaultProvider
+displayName: Flags any `org.apache.bval.jsr*` (bval 1.1) and `org.apache.bval.jsr303*` (bval 1.0) package references
+description: >
+  This recipe flags any org.apache.bval.jsr* (bval 1.1) and org.apache.bval.jsr303* (bval 1.0) package references in validation.xml deployment descriptors.
```
timtebeek, line 44: Please also use markdown backticks for code elements in the description
```suggestion
  This recipe flags any `org.apache.bval.jsr*` (bval 1.1) and `org.apache.bval.jsr303*` (bval 1.0) package references in validation.xml deployment descriptors. 
```

src/test/java/org/openrewrite/java/migrate/jakarta/UpdateApacheWSSecurityPackagesTest.java
timtebeek: This also applies to the other tests in this PR.

timtebeek (changes requested):
Thanks once again for some good migration recipes, and tests covering those recipes.

I've gone through and left more comments this time for items that in the past I've been applying myself, but can't keep up with going forward. Please see what you can do to conform to the code style that we keep in the project, such that it does not detract when reading the implementation.

Also: we need to be mindful of unexpected input files, such that it's probably best to use full xpath expressions rather than shorthands matching anywhere, as a good practice to limit unintended changes.

src/main/resources/META-INF/rewrite/jakarta-ee-9.yml
```diff
@@ -1076,3 +1078,18 @@ recipeList:
       groupId: io.rest-assured
       artifactId: "*"
       newVersion: 5.1.x
+---
+type: specs.openrewrite.org/v1beta/recipe
+name: org.openrewrite.java.migrate.jakarta.UpdateApacheWSSecurityPackages
+displayName: Migrate `org.apache.ws.security` and `org.apache.ws.security.components.crypto` packages to  `org.apache.wss4j.common.ext` and `org.apache.wss4j.common.crypto` packages
+description: >
+  Java EE has been rebranded to Jakarta EE.  This recipe replaces apache security packages to migrate to apache `wss4j`
```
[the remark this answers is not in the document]
ranuradh: updated in latest commit

src/main/resources/META-INF/rewrite/java-ee-7.yml
```diff
@@ -0,0 +1,38 @@
...
+name: org.openrewrite.java.migrate.javaee7.OpenJPAPersistenceProvider
+displayName: Removed OpenJPA providers in the persistence.xml file
+description: >
+  When migrating  to EclipseLink, using OpenJPA providers in EclipseLink results in runtime errors. To resolve these errors, 
+  the recipe removes the flagged OpenJPA provider from the persistence.xml.
+recipeList:
+  - org.openrewrite.xml.ChangeTagValue:
+          elementName: //provider
+          oldValue: org.apache.openjpa.persistence.PersistenceProviderImpl
+          newValue: org.eclipse.persistence.jpa.PersistenceProvider
```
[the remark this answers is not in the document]
ranuradh: updated in latest commit

src/test/java/org/openrewrite/java/migrate/jakarta/UpdateApacheWSSecurityPackagesTest.java
```diff
@@ -0,0 +1,71 @@
...
+    @Override
+    public void defaults(RecipeSpec spec) {
+        spec.parser(JavaParser.fromJavaVersion().
+            classpathFromResources(new InMemoryExecutionContext(), "wss4j-1.6.19", "wss4j-ws-security-common-2.0.0")).
+          recipe(Environment.builder().scanRuntimeClasspath("org.openrewrite.java.migrate.jakarta")
+            .build()
+            .activateRecipes("org.openrewrite.java.migrate.jakarta.UpdateApacheWSSecurityPackages"));
+    }
+    @Test
+    public void updateApache(){
```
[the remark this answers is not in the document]
ranuradh: updated in latest commit

src/test/java/org/openrewrite/java/migrate/jakarta/UpdateApacheWSSecurityPackagesTest.java
ranuradh: Done

src/main/resources/META-INF/rewrite/java-ee-8.yml
ranuradh: done

src/main/resources/META-INF/rewrite/java-ee-8.yml
ranuradh: I wasnt sure about other methods existing,rechecked and we can remove the wild card

src/main/resources/META-INF/rewrite/java-ee-8.yml
ranuradh: done

ranuradh:
Hi @timtebeek thanks for the feedback I made the formatting changes, some of these test files that had the extra spaces etc were from a test files that existed. I thought as long as the tests ran and had the methods, formatting within the tests didnt matter, since they were text files and the Formatter does not update it. I didnt really think using public static void would make any difference hence just used them in a test..., will take note of your inputs for the next batch of recipes..
Regarding "public" in a class..I think the very first recipe I had ..used public hence they continue..have made a note of that as well.. I think this PR should be good now post the updates

src/main/resources/META-INF/rewrite/java-ee-8.yml
```diff
@@ -0,0 +1,62 @@
...
+      elementName: /validation-config/message-interpolator
+      newValue: org.hibernate.validator.messageinterpolation.ResourceBundleMessageInterpolator
+  - org.openrewrite.xml.ChangeTagValue:
+      elementName: /validation-config/traversable-resolver
+      newValue: org.hibernate.validator.engine.resolver.DefaultTraversableResolver
+  - org.openrewrite.xml.ChangeTagValue:
+      elementName: /validation-config/constraint-validator-factory
+      newValue: org.hibernate.validator.engine.ConstraintValidatorFactoryImpl
+  - org.openrewrite.xml.RemoveXmlTag:
+      xPath: /validation-config/parameter-name-provider
```
timtebeek, line 61: Right now we don't match any particular `oldValue`, instead replacing any value with the `newValue`. Is that indeed what we want to happen here, even when the old values are different from what we see in the test? Or should we restrict the recipe to only matching known old values?

timtebeek (approved):
Thanks again! I've posted one last question about using `oldValue` to limit when the recipe makes changes; let me know your thoughts on that and we can merge right after that.

src/main/resources/META-INF/rewrite/java-ee-8.yml
ranuradh: Hi @timtebeek for this particular case we are ok with not having an oldValue match because we want them to move to org.hibernate.validator
```
This rule flags any org.apache.bval.jsr* (bval 1.1) and org.apache.bval.jsr303* (bval 1.0) package references in validation.xml deployment descriptors. Bean Validation 2.0 and later use the Hibernate Validator implementation instead of the Apache BVal implementation which was used for Bean Validation 1.0 and 1.1. The source scanner provides a quick fix that overwrites the configuration to use the Hibernate Validator classes.
```
But for future org.openrewrite.xml.ChangeTagValue it will be good to have a match option.

PR #347, opened 2023-11-13
BeanDiscovery recipe

cjobinabo:
## What's your motivation?
Screenshot 2023-11-13 at 4 46 25 PM

@timtebeek @joanvr

### Checklist


timtebeek (approved):
Thanks again! I've added some polishing commits; the changes that stood out most are:

- do not use fields in visitors for logic; visitors are reused across recipe execution
- reuse the existing ChangeTagAttribute
- extract constants for matchers; saves creating those for every tag
- limit recipe execution to files matching `beans.xml`
- only match xpath at the root; not anywhere else
- add `//language=xml` to tests for code highlighting there

Would help me if you also take the above into account for future PRs. Thanks again!

cjobinabo:
Thanks Tim. I've noted those things down to keep an eye out for in future PRs.

PR #348, opened 2023-11-14
ApplicationPath trailing slash and wildcard no longer supported on Jakarta EE

ranuradh:
## What's changed?
image

@timtebeek @joanvr 

### Checklist


src/test/java/org/openrewrite/java/migrate/jakarta/ApplicationPathWildcardNoLongerAcceptedTest.java
```diff
@@ -0,0 +1,106 @@
...
+        spec.parser(JavaParser.fromJavaVersion()
+          .classpathFromResources(new InMemoryExecutionContext(), "jakarta.ws.rs-api-2.1.3", "jakarta.ws.rs-api-3.1.0"))
+          .recipe(Environment.builder().scanRuntimeClasspath("org.openrewrite.java.migrate.jakarta").build()
+            .activateRecipes("org.openrewrite.java.migrate.jakarta.ApplicationPathWildcardNoLongerAccepted"));
+    }
+    @Test
+    void updateAnnotation() {
+        rewriteRun(
+          //language=java
+          java("""
```
timtebeek, line 39:
```suggestion
          java(
            """
```

src/test/java/org/openrewrite/java/migrate/jakarta/ApplicationPathWildcardNoLongerAcceptedTest.java
```diff
@@ -0,0 +1,106 @@
...
+            package com.test;
+                    
+            import jakarta.ws.rs.ApplicationPath;
+            import jakarta.ws.rs.core.Application;                    
+                       
+            @ApplicationPath("should-flag/*")
+            public class ApplicationPathWithWildcard extends Application {               
+            }
+            """, """
```
timtebeek, line 48:
```suggestion
            """,
            """
```

src/test/java/org/openrewrite/java/migrate/jakarta/ApplicationPathWildcardNoLongerAcceptedTest.java
ranuradh: added in the next commit

src/test/java/org/openrewrite/java/migrate/jakarta/ApplicationPathWildcardNoLongerAcceptedTest.java
ranuradh: added in the next commit

timtebeek (approved):
I've made extensive changes and I hope you agree; I'll only merge after your OK.

Mostly I wanted to reduce the public API we expose, and what shows up in the documentation, as the `RemoveTrailingSlashAnnotation` made assumptions about there only being a single argument, which is not safe to reuse by others when they stumble upon the recipe in the catalog. Instead I've created a dedicated recipe, with updated display name and description, such that it's clear this is a targeted implementation.

We don't currently support constants used as annotation arguments; I'd hope that's rare enough that we can ignore those for now. Otherwise we need to find those constants and update those, which while doable is a lot more effort.

Let me know if you'd want this merged in the current form!

ranuradh:
Hi @timtebeek so the main change was to use the dedicated recipe **ApplicationPathWildcardNoLongerAccepted.java** instead of RemoveTrailingSlashAnnotation. I am fine with this change and we can merge it

ranuradh:
@timtebeek the tests are failing post the changes..

timtebeek:
A rerun fixed it, thanks again!

Issue #194, opened 2023-03-22
Remove calls to `Thread.stop()`


timtebeek:
https://download.java.net/java/early_access/jdk20/docs/api/java.base/java/lang/Thread.html#stop()

As of Java 20 `Thread.stop()` _always_ throws `UnsupportedOperationException`(!).

> **Deprecated, for removal: This API element is subject to removal in a future version.**
This method was originally specified to "stop" a victim thread by causing the victim thread to throw a [ThreadDeath](https://download.java.net/java/early_access/jdk20/docs/api/java.base/java/lang/ThreadDeath.html). It was inherently unsafe. Stopping a thread caused it to unlock all of the monitors that it had locked (as a natural consequence of the ThreadDeath exception propagating up the stack). If any of the objects previously protected by these monitors were in an inconsistent state, the damaged objects became visible to other threads, potentially resulting in arbitrary behavior. Usages of stop should be replaced by code that simply modifies some variable to indicate that the target thread should stop running. The target thread should check this variable regularly, and return from its run method in an orderly fashion if the variable indicates that it is to stop running. If the target thread waits for long periods (on a condition variable, for example), the interrupt method should be used to interrupt the wait. For more information, see [Why are Thread.stop, Thread.suspend and Thread.resume Deprecated?](https://download.java.net/java/early_access/jdk20/docs/api/java.base/java/lang/doc-files/threadPrimitiveDeprecation.html).

It's already been [@Deprecated](https://download.java.net/java/early_access/jdk20/docs/api/java.base/java/lang/Deprecated.html)([since](https://download.java.net/java/early_access/jdk20/docs/api/java.base/java/lang/Deprecated.html#since())="1.2", [forRemoval](https://download.java.net/java/early_access/jdk20/docs/api/java.base/java/lang/Deprecated.html#forRemoval())=true).

From reading the above description it might be too simplistic to simply remove calls without additional guards, but leaving the calls in is guaranteed not to work past v20.

yeikel:
@timtebeek 

Regarding this 

> From reading the above description it might be too simplistic to simply remove calls without additional guards, but leaving the calls in is guaranteed not to work past v20.

Would you accept the simplistic version that just removes it? Or are you hoping to do even more? 

If removal is the initial target, maybe I can take this one

timtebeek:
[...]

PR #354, opened 2023-11-19
Hint at `Thread.stop()` throwing `UnsupportedOperationException`

timtebeek:
## What's changed?
Replace `Thread.stop()` with `throw new UnsupportedOperationException()` on Java 21, with reference to the migration guide.

## What's your motivation?
This helps surface any remaining such migration issues, as the exception is otherwise thrown implicitly.

From the links below a proper replacement will be a lot more complicated; this way we at least already surface the problem to the user, rather than have that be a runtime realization.

## Any additional context
- #194
- https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/Thread.html#stop()
- https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/doc-files/threadPrimitiveDeprecation.html

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/lang/ThreadStopUnsupported.java
+++ b/src/main/java/org/openrewrite/java/migrate/lang/ThreadStopUnsupported.java
@@ -16,7 +16,6 @@
 package org.openrewrite.java.migrate.lang;
 
 import org.openrewrite.ExecutionContext;
-import org.openrewrite.Preconditions;
 import org.openrewrite.Recipe;
 import org.openrewrite.TreeVisitor;
 import org.openrewrite.java.JavaTemplate;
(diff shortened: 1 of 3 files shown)


src/main/java/org/openrewrite/java/migrate/lang/ThreadStopUnsupported.java
```diff
@@ -0,0 +1,69 @@
...
+    @Override
+    public String getDescription() {
+        return "`Thread.stop()` always throws a `new UnsupportedOperationException` in Java 21+. " +
+               "This recipe makes that explicit, as the migration is more complicated." +
+               "See https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/doc-files/threadPrimitiveDeprecation.html .";
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return Preconditions.check(new UsesJavaVersion<>(21),
```
knutwannheden, line 49: Possibly we would want to do this for earlier releases, too. It was deprecated in Java 1.8 or so.

knutwannheden (approved):

src/main/java/org/openrewrite/java/migrate/lang/ThreadStopUnsupported.java
timtebeek: Interesting suggestion.. I'd then lean towards adding the comment for older versions, and only as of Java 21 replace it with that throws, as before that it's still functional. Would you agree with that change?

PR #355, opened 2023-11-19
Recipe- BeansXmlNamespace

ranuradh:
## What's changed?
image

image

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/BeansXmlNamespace.java
+++ b/src/main/java/org/openrewrite/java/migrate/BeansXmlNamespace.java
@@ -1,0 +1,89 @@
+/*
+ * Copyright 2023 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.java.migrate;
+
+import lombok.EqualsAndHashCode;
+import lombok.Value;
+import org.openrewrite.*;
+import org.openrewrite.internal.ListUtils;
+import org.openrewrite.marker.Markers;
+import org.openrewrite.xml.XPathMatcher;
+import org.openrewrite.xml.XmlVisitor;
+import org.openrewrite.xml.tree.Xml;
+
+@Value
+@EqualsAndHashCode(callSuper = true)
+public class BeansXmlNamespace extends Recipe {
+
+    private static final XPathMatcher BEANS_MATCHER = new XPathMatcher("/beans");
+    private static final String NS_SUN = "http://java.sun.com/xml/ns/javaee";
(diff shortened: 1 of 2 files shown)


src/main/java/org/openrewrite/java/migrate/BeansXmlNamespace.java
```diff
@@ -0,0 +1,89 @@
...
+                        nameSpaceValue = attribute.getValueAsString();
+                    } else if (attribute.getKeyAsString().endsWith("schemaLocation")) {
+                        schemaLocation = attribute.getValueAsString();
+                    } else if (DISABLE_VALIDATING.equalsIgnoreCase(attribute.getKeyAsString())) {
+                        return t;
+                    }
+
+                }
+                if (!checkNameSpaceSchemaLocation(nameSpaceValue, schemaLocation)) {
+                    return addAttribute(t, DISABLE_VALIDATING, String.valueOf(true), ctx);
```
timtebeek, line 69: The way I read the screenshots attached to this PR `org.jboss.weld.xml.disableValidating` is a JVM property that you pass in at startup, not something to add to `beans.xml`. I could be mistaken as I lack background here, but figured ask before we continue with this PR.

ranuradh: you are right @timtebeek  ..i think i understood this wrong..our fix is - ```The automated fix for this rule updates incompatible namespaces with the value appropriate to the specified schemaLocation```.  Will update the PR in a bit

timtebeek:
So from what I understand now, we update the `:schemaLocation` to match what we find in `xmlns` inside any `beans.xml`?

src/main/java/org/openrewrite/java/migrate/BeansXmlNamespace.java
```diff
@@ -0,0 +1,90 @@
...
+    private static final XPathMatcher BEANS_MATCHER = new XPathMatcher("/beans");
+    private static final String NS_SUN = "http://java.sun.com/xml/ns/javaee";
+    private static final String NS_JCP = "http://xmlns.jcp.org/xml/ns/javaee";
+    private static final String SUN_SCHEMA_LOCATION = "http://java.sun.com/xml/ns/javaee http://java.sun.com/xml/ns/javaee/beans_1_0.xsd";
+    private static final String JCP_SCHEMA_LOCATION = "http://xmlns.jcp.org/xml/ns/javaee http://xmlns.jcp.org/xml/ns/javaee/beans_1_1.xsd";
+    private static final String DISABLE_VALIDATING = "org.jboss.weld.xml.disableValidating";
+
+    @Override
+    public String getDisplayName() {
+        return "Check valid namespace and schema location in the`beans.xml` file ";
```
timtebeek, line 39: I'm not sure the display name and description fully match what the recipe now does, so I'll propose two changes for you to consider, adapt & adopt or reject.
```suggestion
        return "Change `beans.xml` `schemaLocation` to match XML namespace";
```

timtebeek (approved):
Let me know what you'd want to do with the display name and description; otherwise approved.

ranuradh:
@timtebeek I have updated it

timtebeek:
@ranuradh This recipe is not currently part of any on the larger declarative recipe migrations;
https://github.com/openrewrite/rewrite-migrate-java/tree/main/src/main/resources/META-INF/rewrite
Should this recipe be run in any of the migrations to Jakarta for instance?

timtebeek:
I'm asking since on https://github.com/openrewrite/rewrite-migrate-java/pull/372#discussion_r1422533602 we identified that we can likely do the same with declarative recipes instead, which would be easier to maintain going forward.

PR #363, opened 2023-12-04
`AddJaxwsRuntime` should not always add `jaxws-rt`

knutwannheden:
The Gradle recipe should only add the runtime dependency if the API dependency `jakarta.xml.ws:jakarta.xml.ws-api` is already present. This is already how the Maven recipe operates.


timtebeek:
If this works we appear to have some [similar issues with jaxb](https://github.com/orgs/openrewrite/projects/4/views/10?filterQuery=-label%3A%22parser*%22+jaxb) 😇

PR #366, opened 2023-12-08
CastArraysAsListToList Recipe

leveretconey:
## What's changed?

Add a new recipe convert code like `(Integer[]) Arrays.asList(1, 2, 3).toArray()` to `Arrays.asList(1, 2, 3).toArray(new Integer[0])`.

## What's your motivation?

Arrays.asList(1, 2, 3).toArray() could be cast to Integer[] in JDK8 or earlier but there will be ClassCastException in JDK11 because now toArray() always return an array with Object[] type. During our internal applications' migration process from JDK8 to JDK11, we have identified some similar code patterns, so we hope to add a recipe to automatically fix them.

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/CastArraysAsListToList.java
+++ b/src/main/java/org/openrewrite/java/migrate/CastArraysAsListToList.java
@@ -1,0 +1,96 @@
+/*
+ * Copyright 2023 the original author or authors.
+ * <p>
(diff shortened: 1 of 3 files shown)


timtebeek (approved):
Thanks a lot @leveretconey ! I've pushed some very minor changes, but overall this looks very well done. Glad to have you contribute! :)

PR #374, opened 2023-12-12
Fix #373: Jakarta EE10 bean validation strings in Java

### LINKED ISSUE #373: Jakarta EE10: Find and Replace Bean Validation strings
We do a lot of this...

```java
   @NotNull(message = "Resource Code {javax.validation.constraints.NotNull.message}")
   @Size(min = 1, max = CODE_LENGTH, message = "Resource Code {javax.validation.constraints.Size.message}")
   private String resourceCode;
```

It didn't catch the `javax.validation` props inside the code so it should have done this.

```java
   @NotNull(message = "Resource Code {jakarta.validation.constraints.NotNull.message}")
   @Size(min = 1, max = CODE_LENGTH, message = "Resource Code {jakarta.validation.constraints.Size.message}")
   private String resourceCode;
```

melloware:
Fix #373: Jakarta EE10 bean validation strings in Java

Need some help with this unit test and what the best way to test it is...

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/jakarta/BeanValidationMessages.java
+++ b/src/main/java/org/openrewrite/java/migrate/jakarta/BeanValidationMessages.java
@@ -1,0 +1,80 @@
+package org.openrewrite.java.migrate.jakarta;
+
+import org.openrewrite.ExecutionContext;
+import org.openrewrite.Preconditions;
+import org.openrewrite.Recipe;
+import org.openrewrite.TreeVisitor;
+import org.openrewrite.internal.lang.Nullable;
+import org.openrewrite.java.AnnotationMatcher;
+import org.openrewrite.java.JavaIsoVisitor;
+import org.openrewrite.java.search.UsesType;
+import org.openrewrite.java.tree.J;
+import org.openrewrite.java.tree.JavaType;
+import org.openrewrite.java.tree.TypeUtils;
+
+public class BeanValidationMessages extends Recipe {
+
+    @Override
+    public String getDisplayName() {
+        return "Migrate Bean Validation messages";
+    }
+
+    @Override
+    public String getDescription() {
+        return "Migrate `javax.validation.constraints` messages found in Java files to `jakarta.validation.constraints` equivalents.";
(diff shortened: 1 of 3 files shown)


melloware:
Yes lets assume in JAva files I only want to replace them in annotations!

timtebeek:
Then we'll likely want to write a recipe similar to [this one](https://docs.openrewrite.org/recipes/java/addorupdateannotationattribute) ([source](https://github.com/openrewrite/rewrite/blob/main/rewrite-java/src/main/java/org/openrewrite/java/AddOrUpdateAnnotationAttribute.java)) where we override `visitAnnotation` and replace the value of any annotation argument we encounter.

melloware:
I took a look at this one @timtebeek i have to admit i am struggling a bit.   I basically need to check if the annotation `startsWith` the string `javax.validation.contraints` or `jakarta.validation.constraints` not an exact match of the annotation so I don't have to list them all!

timtebeek:
So the annotation matcher can use `..` and `*` to widen the match such as for instance [seen here](https://github.com/openrewrite/rewrite-logging-frameworks/blob/v2.2.2/src/main/java/org/openrewrite/java/logging/SystemOutToLogging.java#L77):
```java
AnnotationMatcher lombokLogAnnotationMatcher = new AnnotationMatcher("@lombok.extern..*");
```
That ought to allow you to selectively match annotations, and then for those annotations replace the argument values. Happy to help out on a draft pull request; it does not have to function before you open it as a draft here, and then we can tag team to make this work.

melloware:
OK let me open a draft!

timtebeek:
@melloware I think I've got the first bit going in having a test that rewrites what's outlined in #377. Could you check if there's any additional cases that we might need to cover?

melloware (commented):
This looks perfect. This should catch all the scenarios I have seen in the wild.

timtebeek:
> This looks perfect. This should catch all the scenarios I have seen in the wild.

Ah perfect, glad to hear! Thanks again for getting this started; hope this helps you again on any next one!

PR #382, opened 2023-12-27
Fix test failures after latest rewrite-templating changes

timtebeek:
We're now actually doing the correct replacement; not sure how that was missed previously.


knutwannheden:
Weird. But I didn't look into it.

PR #383, opened 2024-01-04
Add useMimeCoder option to UseJavaUtilBase64

leveretconey:
## What's changed?

Add useMimeCoder option to UseJavaUtilBase64 recipe, allowing user to migrate sun.misc.Base64 api using java.lang.Base64.getMimeEncoder()/getMimeDecoder() instead of Base64.getEncoder()/getDecoder(). This option is false by default.

## What's your motivation?

UseJavaUtilBase64 helps user to migrate code from legacy base64 api. This recipe is useful on our applications, but recently we found some problem. If an application that has not been migrated encodes a String using old api and sends the result to another application that has been migrated to Java11 for decoding, there will be an exception, because Base64.getDecoder() does not tolerate line breaks. See the following example:

``` java
import sun.misc.BASE64Encoder;
import java.util.Base64;

public class Main {
    // This application has not migrated yet
    private static String application1(byte[] bytes) {
        return new BASE64Encoder().encode(bytes);
    }

    // This application been migrated using UseJavaUtilBase64
    private static byte[] application2(String s) throws Exception {
        // Current way to migrate, which causes
        // Exception in thread "main" java.lang.IllegalArgumentException: Illegal base64 character a
        return Base64.getDecoder().decode(s);

        // My proposal
        //  return Base64.getMimeDecoder().decode(s);
    }

    public static void main(String[] args) throws Exception {
        String s = "A very long content that will be encoded to multiple lines by sun.misc.Base64Encoder";
        String encoded = application1(s.getBytes());
        byte[] decoded = application2(encoded);
        System.out.println(new String(decoded));
    }
}


```

Therefore I propose to add an option to this recipe that allows migrating to java.lang.Base64.getMimeEncoder()/getMimeDecoder() which tolerates line breaks and solves the problem.

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/UseJavaUtilBase64.java
+++ b/src/main/java/org/openrewrite/java/migrate/UseJavaUtilBase64.java
@@ -16,10 +16,7 @@
 package org.openrewrite.java.migrate;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
-import org.openrewrite.ExecutionContext;
-import org.openrewrite.Preconditions;
-import org.openrewrite.Recipe;
-import org.openrewrite.TreeVisitor;
+import org.openrewrite.*;
 import org.openrewrite.java.ChangeType;
 import org.openrewrite.java.JavaTemplate;
 import org.openrewrite.java.JavaVisitor;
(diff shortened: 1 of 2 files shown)


timtebeek:
Hi @leveretconey ; thanks for the detailed report & contribution once again!

Right now the tests are failing unrelated to your changes; I've logged an issue to have that fixed:
- https://github.com/openrewrite/rewrite-migrate-java/issues/384

Once we've solved the issue on main I'll update your PR and see these changes through.

timtebeek (approved):
Thanks again!

PR #387, opened 2024-01-08
Fix negate ternary

timtebeek:
Types were wrong on the test; the actual fix should go into oprenwrite/rewrite most likely.


timtebeek:
Needs
- https://github.com/openrewrite/rewrite/pull/3896

PR #392, opened 2024-01-12
Add recipe for rule: HttpSessionInvalidate

evie-lau:
## What's changed?

Add recipe to migrate `HttpSession:invalidate()` (pre-java6) with `HttpServletRequest:logout()` (java6 and later)

I wasn't sure about the location of the test file, and just placed it in the `javaee` folder. Please let me know if it should be moved anywhere else.

I'm also wondering if I should update the `java-ee-7.yml` recipeList to include `org.openrewrite.java.migrate.javaee6`?

@cjobinabo 
@AnuRam123 

### Checklist


cjobinabo (approved):
Looks good to me.

src/main/resources/META-INF/rewrite/java-ee-6.yml
```diff
@@ -0,0 +1,40 @@
...
+description: Do not rely on HttpSession `invalidate` method for programmatic security logout. Add the HttpServletRequest `logout` method which was introduced in Java EE 6 as part of the Servlet 3.0 specification.
+recipeList:
+  - org.openrewrite.java.ChangeMethodName:
+      methodPattern: javax.servlet.http.HttpSession invalidate()
+      newMethodName: logout
+      ignoreDefinition: true
+  - org.openrewrite.java.ChangeType:
+      oldFullyQualifiedTypeName: javax.servlet.http.HttpSession
+      newFullyQualifiedTypeName: javax.servlet.http.HttpServletRequest
+      ignoreDefinition: true
```
timtebeek, line 40: I'm wondering if this might end up changing too broadly; while the current test works, we are in fact changing _**any reference**_ to `HttpSession` to a reference to `HttpServletRequest`, not just when used to call `invalidate()`.

When comparing these docs pages it appears that `HttpSession` was retained, and perhaps not a direct replacement for `HttpServletRequest` in all cases
https://docs.oracle.com/javaee/6/api/javax/servlet/http/HttpSession.html
https://docs.oracle.com/javaee/7/api/javax/servlet/http/HttpSession.html
https://docs.oracle.com/javaee/7/api/javax/servlet/http/HttpServletRequest.html

Maybe we need to be more selective, and somehow access the active `HttpServletRequest` to call `logout` there.

timtebeek (changes requested):
Hi @evie-lau and welcome to the project! Good to have you onboard to help automate these changes.

I have a bit of a worry about the current implementation of this recipe; perhaps @cjobinabo can help you through a more subtle approach to change this logout behavior, as changing the type will have unintended effects on other parts of the code too.

src/main/resources/META-INF/rewrite/java-ee-6.yml
evie-lau: That's a good point. I've been encountering similar requirements for some other recipes I'm looking to create, mainly in needing to filter specific instances. 

I didn't notice anything able to do that in the [basic java recipes](https://docs.openrewrite.org/recipes/java/), and was wondering if I would have to create a new recipe to do so.

timtebeek: Yes I feel this most likely would have to be a custom recipe, given how you want to instead call a method on an entirely different type instance, that would need to be accessed from somewhere. Perhaps it would be good to start with a more realistic test, and explore the options from there. For instance I know elsewhere we've inserted comments to tell about necessary migration steps that we can't yet automate; that might be the case here if we can't find a request to call `logout` on.

evie-lau:
Changed to use a custom recipe.

When it finds an instance of `HttpSession.invalidate()`, it looks at the encapsulating MethodDeclaration for the `HttpServletRequest` parameter, and uses that to call `logout()`

I don't like this current implementation, and had trouble finding a cleaner way to do it, so please provide any feedback if it could be improved.

While I think this updated recipe adheres to making changes conservatively to avoid unnecessary changes, I'm not sure if it covers all the cases necessary. But it can always be expanded in the future.

src/main/java/org/openrewrite/java/migrate/javax/HttpSessionInvalidate.java
```diff
@@ -0,0 +1,102 @@
...
+                }
+
+                // Get the HttpServletRequest param
+                J.VariableDeclarations httpServletRequestDeclaration = (J.VariableDeclarations) parentMethod.getParameters().get(servletReqParamIndex);
+
+                // Replace HttpSession.invalidate() with HttpServletRequest.logout()
+                final JavaTemplate logoutTemplate =
+                        JavaTemplate.builder("#{any(javax.servlet.http.HttpServletRequest)}.logout()")
+                                .imports("javax.servlet.http.HttpServletRequest")
+                                .javaParser(JavaParser.fromJavaVersion().classpathFromResources(ctx, "javax.servlet-3.0"))
```
timtebeek, line 74: The `javax.servlet-3.0` jar is only in `src/test/resources/META-INF/classpath`; Perhaps we should move that to `src/main` if we now also need it outside tests.

src/main/java/org/openrewrite/java/migrate/javax/HttpSessionInvalidate.java
```diff
@@ -0,0 +1,102 @@
...
+                                .build();
+                maybeAddImport("javax.servlet.http.HttpServletRequest");
```
timtebeek, line 76: I'm surprised to see this `maybeAddImport`, since we're using a method parameter that should already be there right?
```suggestion
```

src/main/java/org/openrewrite/java/migrate/javax/HttpSessionInvalidate.java
```diff
@@ -0,0 +1,102 @@
...
+                method = logoutTemplate.apply(
+                        getCursor(),
+                        method.getCoordinates().replace(),
+                        httpServletRequestDeclaration.getVariables().get(0)
+                );
+                return method;
+            }
+            return method;
```
timtebeek, line 84: Not sure if we're expecting to match any nested method invocations, but could make sense to do this to be safe:
```suggestion
            }
            return super.visitMethodInvocation(method, ctx);
```

src/main/java/org/openrewrite/java/migrate/javax/HttpSessionInvalidate.java
```diff
@@ -0,0 +1,102 @@
...
+
+        /**
+         * @return the param index position of the HttpServletRequest parameter object
+         */
+        @Nullable
+        private Integer getServletRequestIndex(J.MethodDeclaration parentMethod) {
+            List<JavaType> params = parentMethod.getMethodType().getParameterTypes();
+            for (int i = 0; i < params.size(); ++i) {
+                String paramType = ((JavaType.Class) params.get(i)).getFullyQualifiedName();
+                if (paramType.equals("javax.servlet.http.HttpServletRequest")) {
```
timtebeek, line 95: We prefer to match using `org.openrewrite.java.TypeMatcher` instead of doing String comparisons. That then also handles any missing types correctly.
```suggestion
                if (HTTP_SERVLET_REQUEST_TYPE_MATCHER.matches(params.get(i))) {
```
With the addition of a field
```java
final TypeMatcher HTTP_SERVLET_REQUEST_TYPE_MATCHER = new TypeMatcher("javax.servlet.http.HttpServletRequest");
```

src/main/resources/META-INF/rewrite/java-ee-6.yml
```diff
@@ -0,0 +1,33 @@
...
+  - deprecated
+recipeList:
+  - org.openrewrite.java.migrate.javaee6.HttpSessionInvalidateMethod
+
+---
+type: specs.openrewrite.org/v1beta/recipe
+name: org.openrewrite.java.migrate.javaee6.HttpSessionInvalidateMethod
+displayName: Use HttpServletRequest `logout` method for programmatic security logout in Servlet 3.0
+description: Do not rely on HttpSession `invalidate` method for programmatic security logout. Add the HttpServletRequest `logout` method which was introduced in Java EE 6 as part of the Servlet 3.0 specification.
+recipeList:
```
timtebeek, line 32: There should be no need for this indirection; We can (maybe) move the name and description into the custom recipe.
```suggestion
```

src/test/java/org/openrewrite/java/migrate/javaee/HttpSessionInvalidateTest.java
```diff
@@ -0,0 +1,112 @@
...
+        rewriteRun(
+          //language=java
+          java(
+            """
+              import javax.servlet.http.HttpServletResponse;
+              import javax.servlet.http.HttpSession;
+                              
+              class Foo {
+                  void logOut(HttpSession session, HttpServletResponse res) {
+                      session.logout();
```
timtebeek, line 68: This method can not be found; should we instead use `invalidate` here but not make changes as there's no `HttpServletRequest` to pull from the method arguments?
```suggestion
                      session.invalidate();
```

timtebeek (commented):
Great improvements already; I've added a few more review comments, and our bot did some as well which I've applied.

Understand your hesitancy around finding the method request parameter. Perhaps we can move all of that logic into an appropriately named method with some JavaDoc to document your reservations. That way if we revisit (or bypass) this in the future it's nicely contained.

We can indeed expand the cases covered in the future if needed, by for instance adding a new parameter where applicable (although that may break compilation in rare cases).

src/main/java/org/openrewrite/java/migrate/javax/HttpSessionInvalidate.java
evie-lau: Ah, yes. I had previously worked with an example that didn't include it to begin with. I'm not 100% sure on how applications generally implement their logouts/invalidates, but from most examples online, the `HttpServletRequest` is always there.

src/test/java/org/openrewrite/java/migrate/javaee/HttpSessionInvalidateTest.java
evie-lau: Yes, missed this when copying and changing another test.

src/main/resources/META-INF/rewrite/java-ee-6.yml
evie-lau: Updating this yaml and the name/description in the custom recipe.

src/test/java/org/openrewrite/java/migrate/javaee/HttpSessionInvalidateTest.java
```diff
@@ -0,0 +1,110 @@
...
+                      session.invalidate();
+                      res.sendRedirect("login.html");
+                  }
+              }
+              """
+          )
+        );
+    }
+
+    @Test
```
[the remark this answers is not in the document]
timtebeek:
```suggestion
    @DocumentExample
    @Test
```

timtebeek:
Thanks a lot! Sorry for the noise with those automated reviews; they're new and was still trying things out for a bit. PR looks great.

PR #337, opened 2023-11-02
Run on Java toolchain 21

timtebeek:
Mostly to flush out issues with the build, and to be able to add new `SequencedCollection` recipes for getFirst, getLast & addFirst.

PR #403, opened 2024-01-26
Use `getFirst()` instead of `stream().findFirst().orElseThrow()`

timtebeek:
## Any additional context
- https://github.com/openrewrite/rewrite-migrate-java/pull/337


src/test/java/org/openrewrite/java/migrate/util/IteratorNextTest.java
```diff
@@ -80,7 +80,7 @@ void bar(Collection<String> collection) {
     }
 
     @Test
-    void nextCommentRetained() {
+    void nextCommentList() {
```
timtebeek, line 83:
```suggestion
    void nextCommentLost() {
```

PR #410, opened 2024-02-09
Add @Column along @ElementCollection for OpenJPA to EclipseLink migration

evie-lau:
## What's changed?
Add custom recipe to help with OpenJPA -> EclipseLink migration.
If `@ElementCollection` exists without a `@Column` annotation with the name attribute, add/update `@Column` annotation with `name = "element`

## What's your motivation?
https://wiki.eclipse.org/EclipseLink/Examples/JPA/Migration/OpenJPA/Mappings#.40ElementCollection

## Any additional context
This is an initial implementation that seems to mostly work.. except it duplicates the `@Column` annotation on the last test when adding the annotation, trying to fix it.

### Checklist


timtebeek (approved):
Nice work, thanks! I understood @cjobinabo would like to review as well, so I'll leave it up to him when we merge.

src/main/java/org/openrewrite/java/migrate/javax/AddColumnAnnotation.java
```diff
@@ -0,0 +1,82 @@
...
+                            maybeAddImport("javax.persistence.Column");
+                            return JavaTemplate.builder("@Column(name = \"element\")")
+                                    .javaParser(JavaParser.fromJavaVersion().classpathFromResources(ctx, "javax.persistence-api-2.2"))
+                                    .imports("javax.persistence.Column")
+                                    .build()
+                                    .apply(getCursor(), multiVariable.getCoordinates().addAnnotation(Comparator.comparing(J.Annotation::getSimpleName)));
+                        }
+
+                        // Update existing @Column annotation
+                        doAfterVisit(new AddOrUpdateAnnotationAttribute(
```
evie-lau, line 73: Since `doAfterVisit` applies to the whole file, this may unintentionally change an existing `@Column` annotation on a varDecl that does not have an `@ElementCollection` annotation.

timtebeek: A way around here might be to directly call visit methods on the `.getVisitor()` returned; haven't tried that specifically here yet, but could be a way around.

src/main/java/org/openrewrite/java/migrate/javax/AddColumnAnnotation.java
```diff
@@ -0,0 +1,82 @@
...
+                                    .javaParser(JavaParser.fromJavaVersion().classpathFromResources(ctx, "javax.persistence-api-2.2"))
+                                    .imports("javax.persistence.Column")
+                                    .build()
+                                    .apply(getCursor(), multiVariable.getCoordinates().addAnnotation(Comparator.comparing(J.Annotation::getSimpleName)));
+                        }
+
+                        // Update existing @Column annotation
+                        J.VariableDeclarations newVar = (J.VariableDeclarations) new AddOrUpdateAnnotationAttribute(
+                                "javax.persistence.Column", "name", "element", true)
+                                .getVisitor().visit(multiVariable, ctx);
```
evie-lau, line 75: I can't figure out why this is giving me this error, even after changing the cast type to J.Annotation which it should be returning.
```
java.lang.ClassCastException: class java.lang.String cannot be cast to class org.openrewrite.java.tree.J (java.lang.String is in module java.base of loader 'bootstrap'; org.openrewrite.java.tree.J is in unnamed module of loader 'app')
```
But in the meantime, the custom code I had in 4653e046ca504fe547e60fe4f3edf1ebfc0f386d works with all the current tests.

timtebeek: Fixed it locally with https://github.com/openrewrite/rewrite-migrate-java/pull/410/commits/c0873546dafaac8c2bcd288dfad63eebfc4f777c ; when calling out to other visitors, especially when they operate on other LST elements, it helps to pass in the active cursor. Hope that helps!

evie-lau: Great! Glad to see we can keep this cleaner solution!

evie-lau: Curious why this one needed to pass the cursor, whereas some of the other recipes didn't run into any issues without it (like [this PR](https://github.com/openrewrite/rewrite-migrate-java/pull/416/files)).

timtebeek: Hmm; comparing the two I don't immediately have an answer either. 😕

src/test/java/org/openrewrite/java/migrate/javax/AddColumnAnnotationTest.java
```diff
@@ -0,0 +1,262 @@
...
+import org.openrewrite.DocumentExample;
+import org.openrewrite.InMemoryExecutionContext;
+import org.openrewrite.java.JavaParser;
+import org.openrewrite.java.migrate.javax.AddColumnAnnotation;
+import org.openrewrite.test.RecipeSpec;
+import org.openrewrite.test.RewriteTest;
+
+import static org.openrewrite.java.Assertions.java;
+
+class AddColumnAnnotationTest implements RewriteTest {
```
cjobinabo, line 28: We should also have tests for the following scenarios:

- @ElementCollection is used with @Transient - shouldn't flag (this case isn't explicitly mentioned in our rule help but the [OpenJPA to EclipseLink mapping guide](https://wiki.eclipse.org/EclipseLink/Examples/JPA/Migration/OpenJPA/Mappings#.40javax.persistence.ElementCollection) suggests we shouldn't make any assumptions when @Transient is used `It is undefined in the JPA specification to define a mapping (@Basic, @OneToOne, @ManyToOne, @OneToMany, @ManyToMany, ...) and transient. While OpenJPA allows this, EclipseLink correctly throws and exception indicating the conflict in the configuration. Transient configuration on an attribute is intended to prevent default mappings from being applied when calculating the mappings for an entity. With default mappings unmapped attributes are assumed according to the specification and @Transient prevents these assumptions.`)
- case where you expect your recipe to add @Column to a nested class (I see the recipe code doesn't call the super classes `visitVariableDeclarations` method so the code may be overlooking nested classes. 
- no annotation is used on a method in the Entity class - shouldn't flag

src/main/java/org/openrewrite/java/migrate/javax/AddColumnAnnotation.java
```diff
@@ -0,0 +1,80 @@
...
+
+import java.util.Comparator;
+
+@Value
+@EqualsAndHashCode(callSuper = false)
+public class AddColumnAnnotation extends Recipe {
+
+    @Override
+    public String getDisplayName() {
+        return "ElementCollection annotations must be accompanied by a defined Column annotation";
```
cjobinabo, line 40: For consistency with your next PR `ElementCollection` should probably be updated to ``@ElementCollection`` and `Column` to ``@Column``.

src/test/java/org/openrewrite/java/migrate/javax/AddColumnAnnotationTest.java
evie-lau: By the last point, did you mean if the enclosing Class is not annotated with `@Entity`?

cjobinabo: I had something else in mind, but the `doNotAddColumnMappingIfClassNotEntity` test you added is better.

cjobinabo:
Thanks for making the updates, Evie. It looks like the code should be able to handle nested classes, but I didn't see any tests to confirm that the functionality worked. Were there any restrictions to adding that test? I also noticed a previous PR of mine included the sub-class check but didn't include a test for it, even though I remember doing work for the test. I vaguely recall some issues with the formatting of the expected output.

src/main/java/org/openrewrite/java/migrate/javax/AddColumnAnnotation.java
```diff
@@ -0,0 +1,98 @@
...
+                    public J.VariableDeclarations visitVariableDeclarations(J.VariableDeclarations multiVariable, ExecutionContext ctx) {
+                        // Exit if var does not have @ElementCollection or has @Transient
+                        if (FindAnnotations.find(multiVariable, "@javax.persistence.ElementCollection").isEmpty()
+                            || !FindAnnotations.find(multiVariable, "@javax.persistence.Transient").isEmpty()) {
+                            return multiVariable;
+                        }
+
+                        // Confirm direct parent class is @Entity
+                        J.ClassDeclaration parentClass = getCursor().dropParentUntil(parent -> parent instanceof J.ClassDeclaration).getValue();
+                        if (FindAnnotations.find(parentClass, "javax.persistence.Entity").isEmpty()) {
```
evie-lau, line 75: I found that this is somehow finding the annotation on the inner class as well, returning true for the `parentClass`, when it shouldn't. Fails this test case: 
image

Is this the intended behavior for `FindAnnotations.find`? @timtebeek

evie-lau: For now, changed this code to manually search `parentClass.getLeadingAnnotations()`. But question still remains

timtebeek: Hmm; that's unexpected to me, but I'm still out. Tagging @knutwannheden to have a look when he can.

cjobinabo (approved):
The logic here now looks to be consistent with our existing tool. Thanks for making the updates,

PR #413, opened 2024-02-13
Recipe to remove Id annotation from Embeddable class when referenced by EmbeddedId

evie-lau:
## What's changed?
Add new recipe and tests for OpenJPA -> EclipseLink migration

Remove `@Id` annotation if...
- parent class is `@Embeddable` and
- parent class referenced by `@EmbeddedId`

## What's your motivation?
https://wiki.eclipse.org/EclipseLink/Examples/JPA/Migration/OpenJPA/Mappings#.40EmbeddedId_and_.40Id_in_the_Same_Entity

### Checklist


timtebeek (approved):
Implementation looks perfect to me; great to see! I've done some very minor touch ups to the docs, code highlighting inside text blocks, and formatting. Good to merge if you ask me, but let's wait for @cjobinabo to return before we merge.

src/test/java/org/openrewrite/java/migrate/javax/RemoveEmbeddableIdTest.java
```diff
@@ -0,0 +1,163 @@
...
+import org.junit.jupiter.api.Test;
+import org.openrewrite.DocumentExample;
+import org.openrewrite.InMemoryExecutionContext;
+import org.openrewrite.java.JavaParser;
+import org.openrewrite.test.RecipeSpec;
+import org.openrewrite.test.RewriteTest;
+
+import static org.openrewrite.java.Assertions.java;
+
+class RemoveEmbeddableIdTest implements RewriteTest {
```
cjobinabo, line 27: It would be good to include the following scenarios:

- `Class1` annotated with `@Entity` and defines a field of type `Class2` annotated with `@Embedded`. Then `Class2`'s definition is annotated with `@Embeddable` and defines a field of any type annotated with `@Id`
- `@Embeddable ` is used on a sub class where this subclass uses `@Id`

evie-lau: Does this recipe also have to handle the `@Embedded` case? The rule only mentioned `@EmbeddableId`

evie-lau: In the second case, is `Class1` defining `Class2` or `Class2SubClass` with `@EmbeddableId`?
And should the recipe be expected to run?

src/test/java/org/openrewrite/java/migrate/javax/RemoveEmbeddableIdTest.java
```diff
@@ -160,4 +160,50 @@ public class EmbeddableObject {
...
+    void removeEmbeddableIdFromSubClass() {
+        //language=java
+        rewriteRun(
+          java(
+            """
+              import javax.persistence.EmbeddedId;
+              import javax.persistence.Entity;
+
+              @Entity
+              public class MainEntity {
```
cjobinabo, line 174: I should have been clearer about describing the sub-class test case. I was looking for something like this:
```
              import javax.persistence.Embeddable;
              import javax.persistence.Id;
              public class SomeClass {
                        @Embeddable
                        public class EmbeddableObject {
                            @Id
                            private int field;
                        }
              }
```

evie-lau: If `SomeClass` also had it's own `@Id` annotated field, should that be left unchanged by this recipe?

evie-lau: Conversely, if the outer `SomeClass` is `@Embeddable`, has an `@Id` field, and has an inner class without `@Embeddable` and an inner field with `@Id`, should that inner field annotation be removed as well?

cjobinabo (approved):
Thanks for making the updates

PR #415, opened 2024-02-14
Add default constructor to Entity class

evie-lau:
## What's changed?
New recipe for OpenJPA -> EclipseLink migration
If class is tagged with `@Entity` or `@MappedSuperclass`, add default no-arg constructor if missing.

## What's your motivation?
https://wiki.eclipse.org/EclipseLink/Examples/JPA/Migration/OpenJPA/Mappings#No-arg_Constructor

### Checklist


src/test/java/org/openrewrite/java/migrate/javaee/AddDefaultConstructorToEntityClassTest.java
```diff
@@ -0,0 +1,201 @@
...
+                  private int id;
+
+                  public void doNothing() {
+                  }
+
+                  public MissingNoArgConstructorEntity(int id) {
+                      this.id = id;
+                  }
+              }
+              ""","""
```
timtebeek, line 179:
```suggestion
              """,
              """
```

src/main/java/org/openrewrite/java/migrate/javax/AddDefaultConstructorToEntityClass.java
```diff
@@ -0,0 +1,90 @@
...
+                        if (hasDefaultConstructor) {
+                            return cd;
+                        }
+
+                        // Add default constructor with empty body
+                        // For testing simplicity, adding as last statement in class body
+                        cd = cd.withBody(defaultConstructorTemplate.apply(
+                                new Cursor(getCursor(), cd.getBody()),
+                                cd.getBody().getCoordinates().lastStatement(),
+                                cd.getSimpleName()
```
evie-lau, line 82: This should be generating the correct `public <classname>(){}` constructor, and it did at one point.
But it suddenly started making `public Template <classname>(){}`, and I hadn't made any changes to the recipe code.

evie-lau: For reference, 
```
final JavaTemplate defaultConstructorTemplate = JavaTemplate.builder("public #{}(){}").build();
```

evie-lau: Fixed by using `.contextSensitive()` on the JavaTemplate builder

src/main/java/org/openrewrite/java/migrate/javax/AddDefaultConstructorToEntityClass.java
```diff
@@ -0,0 +1,93 @@
...
+ */
+package org.openrewrite.java.migrate.javax;
+
+import lombok.EqualsAndHashCode;
+import lombok.Value;
+import org.openrewrite.Cursor;
+import org.openrewrite.ExecutionContext;
+import org.openrewrite.Preconditions;
+import org.openrewrite.Recipe;
+import org.openrewrite.TreeVisitor;
```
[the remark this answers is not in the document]
evie-lau: @timtebeek Is it actually preferred to use the wildcard import?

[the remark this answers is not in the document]
timtebeek: We try to follow the IntelliJ defaults for formatting and imports, which use a wildcard import for five or more imports from the same class I believe. That way folks can use their short cuts to organize imports and format without giving that too much thought. The automation is here to be consistent going forward, in what I hope is a frictionless experience for the most part.

timtebeek (approved):
Great work once again! Good to see you worked out the issue with context sensitivity when injecting a constructor. Indeed that'll be down to how we generate and extract the code snippet from JavaTemplate. If you're ever curious what gets generated you can add this little snippet to the code to get the generated template.
image

```java
class __P__ {
    static native <T> T p();
    static native <T> T[] arrp();
    static native boolean booleanp();
    static native byte bytep();
    static native char charp();
    static native double doublep();
    static native int intp();
    static native long longp();
    static native short shortp();
    static native float floatp();
}

class __M__ {
    static native Object any(Object o);

    static native Object any(java.util.function.Predicate<Boolean> o);

    static native <T> Object anyT();
}

public class MissingNoArgConstructorEntity {
    private int id;

    public MissingNoArgConstructorEntity(int id) {
    }

    /*__TEMPLATE__*/
    public MissingNoArgConstructorEntity() {
    }/*__TEMPLATE_STOP__*/
}
```
For constructors specifically we need that context sensitivity, as otherwise the generated template does not compile.
image


Hope that helps you on your next one!

timtebeek:
Once again we'll wait for @cjobinabo to merge when convenient there! :)

src/main/java/org/openrewrite/java/migrate/javax/AddDefaultConstructorToEntityClass.java
```diff
@@ -0,0 +1,79 @@
...
+               "have a default, no-argument constructor. The OpenJPA implementation automatically generates the " +
+               "no-argument constructor, but the EclipseLink implementation does not.";
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return Preconditions.check(
+                Preconditions.or(
+                        new UsesType<>("javax.persistence.Entity", true),
+                        new UsesType<>("javax.persistence.MappedSuperclass", true)),
```
cjobinabo, line 47: I didn't see any mentioned of the `@MappedSuperclass` annotation in the OpenJPA to EclipseLink JPA Migration guide. Did you find additional information that suggested that we should be treating `@MappedSuperclass` the same as `@Entity`?

src/test/java/org/openrewrite/java/migrate/javax/AddDefaultConstructorToEntityClassTest.java
```diff
@@ -0,0 +1,286 @@
...
+import org.junit.jupiter.api.Test;
+import org.openrewrite.DocumentExample;
+import org.openrewrite.InMemoryExecutionContext;
+import org.openrewrite.java.JavaParser;
+import org.openrewrite.test.RecipeSpec;
+import org.openrewrite.test.RewriteTest;
+
+import static org.openrewrite.java.Assertions.java;
+
+class AddDefaultConstructorToEntityClassTest implements RewriteTest {
```
cjobinabo, line 27: We should include a sub-class example where an update is expected here as well.

src/main/java/org/openrewrite/java/migrate/javax/AddDefaultConstructorToEntityClass.java
evie-lau: Our rule in WAMT says:
> This rule scans for entity classes with the @Entity or @MappedSuperclass annotations. Entity classes with constructors that do not also have a no-argument constructor are flagged. The quick fix adds an empty no-argument constructor to the class.

Should that be taken to mean that `@MappedSuperclass` still isn't modified?

src/test/java/org/openrewrite/java/migrate/javax/AddDefaultConstructorToEntityClassTest.java
evie-lau: Do you mean a subclass of an `@Entity` class? and should the subclass be annotated with `@Entity` as well?

src/main/java/org/openrewrite/java/migrate/javax/AddDefaultConstructorToEntityClass.java
evie-lau: After we discussed, seems we should add the default constructor to a MappedSuperclass

src/main/java/org/openrewrite/java/migrate/javax/AddDefaultConstructorToEntityClass.java
```diff
@@ -0,0 +1,83 @@
...
+                                .anyMatch(constructor -> constructor.getParameters().get(0) instanceof J.Empty)) {
+                            return super.visitClassDeclaration(classDecl, ctx);
+                        }
+
+                        // Add default constructor with empty body
+                        return classDecl.withBody(JavaTemplate.builder("public #{}(){}")
+                                .contextSensitive()
+                                .build()
+                                .apply(new Cursor(getCursor(), classDecl.getBody()),
+                                        classDecl.getBody().getCoordinates().addMethodDeclaration(Comparator.comparing(J.MethodDeclaration::getSimpleName)),
```
evie-lau, line 75: Noticing intermittent discrepencies on the insertion order for `addMethodDeclaration` using this comparator, which causes intermittent unit test failures. 

Maybe should just pick a consistent insertion point for tests? Even if the resulting code might not look as organized...

evie-lau:
Reverted recent change because `@Entity` annotation can never be used on an inner class

src/main/java/org/openrewrite/java/migrate/javax/AddDefaultConstructorToEntityClass.java
evie-lau: Switched to using `firstStatement` to guarantee order for tests.

PR #416, opened 2024-02-15
Use JoinColumn annotation for relationship mappings

evie-lau:
## What's changed?
OpenJPA -> EclipseLink migration recipe
When a `@Column` annotated attribute has a relationship mapping (`@OneToOne` or `@ManyToOne`), replace with `@JoinColumn` and drop certain annotation attributes.

## What's your motivation?
https://wiki.eclipse.org/EclipseLink/Examples/JPA/Migration/OpenJPA/Mappings#.40Column_in_Relationship_Mappings

### Checklist


evie-lau:
I noticed the `doAfterVisit` is applying changes to the whole file instead of the filtered VariableDeclaration.
Is there a way to apply those other recipes specifically to the current item? Or would I have to build my own JavaTemplate for it?

src/test/java/org/openrewrite/java/migrate/javaee/UseJoinColumnForMappingTest.java
```diff
@@ -0,0 +1,206 @@
...
+                  @Column(name="amount", length=512, precision=12, scale=2)
+                  private double amount;
+
+                  @ManyToOne
+                  @Column(name="account", length=512, precision=12, scale=2)
+                  private Account account;
+              }
+              """,
+            """
+              import javax.persistence.*;
```
evie-lau, line 181: After the recipe ran, there would be 5 imports from `javax.persistence`, and the outputted code used wildcard by default.

timtebeek (approved):
Near perfect; Great to see you reuse as much as possible. We'll wait for @cjobinabo to return before a merge.

src/main/java/org/openrewrite/java/migrate/javax/UseJoinColumnForMapping.java
```diff
@@ -0,0 +1,79 @@
...
+                        if (FindAnnotations.find(multiVariable, "javax.persistence.Column").isEmpty()
+                            || (FindAnnotations.find(multiVariable, "javax.persistence.OneToOne").isEmpty()
+                                && FindAnnotations.find(multiVariable, "javax.persistence.ManyToOne").isEmpty())) {
+                            return multiVariable;
+                        }
+
+                        // Change @Column to @JoinColumn
+                        // The javax.persistence.Column attributes length, precision, and scale are not copied.
+                        maybeRemoveImport("javax.persistence.Column");
+                        maybeAddImport("javax.persistence.JoinColumn"); // TODO: exceeding 5+ javax.persistence imports turns into wildcard import in tests
```
timtebeek, line 68: We detect the import style (or fallback to defaults) before we apply those; use of wildcards (or not) should thus not be a concern of any particular recipe. :)
```suggestion
                        maybeAddImport("javax.persistence.JoinColumn");
```

src/main/java/org/openrewrite/java/migrate/javax/UseJoinColumnForMapping.java
```diff
@@ -0,0 +1,79 @@
...
+                        // Exit if not annotated with @Column and a relationship mapping annotation
+                        if (FindAnnotations.find(multiVariable, "javax.persistence.Column").isEmpty()
+                            || (FindAnnotations.find(multiVariable, "javax.persistence.OneToOne").isEmpty()
+                                && FindAnnotations.find(multiVariable, "javax.persistence.ManyToOne").isEmpty())) {
+                            return multiVariable;
+                        }
+
+                        // Change @Column to @JoinColumn
+                        // The javax.persistence.Column attributes length, precision, and scale are not kept.
+                        maybeRemoveImport("javax.persistence.Column");
```
cjobinabo, line 67: Your strings are reused a few times. Maybe it makes sense to define them as a class variable which can be referenced instead.

src/test/java/org/openrewrite/java/migrate/javax/UseJoinColumnForMappingTest.java
```diff
@@ -0,0 +1,182 @@
...
+import org.junit.jupiter.api.Test;
+import org.openrewrite.DocumentExample;
+import org.openrewrite.InMemoryExecutionContext;
+import org.openrewrite.java.JavaParser;
+import org.openrewrite.test.RecipeSpec;
+import org.openrewrite.test.RewriteTest;
+
+import static org.openrewrite.java.Assertions.java;
+
+class UseJoinColumnForMappingTest implements RewriteTest {
```
cjobinabo, line 27: The following these cases should be ported over from our bucket:

- @OneToOne used with @Column (with a few attributes provided)
- @ManyToOne and @Column, both without any attributes
- @OneToOnee and @Column, both without any attributes

We should also test the use of these annotations within a sub class here as well.

evie-lau: Did you mean subclass extending the `@Entity` class? or an inner class inside the `@Entity` class?

cjobinabo (approved):
Looks good

timtebeek (approved):

PR #418, opened 2024-02-18
ClassCastException in OptionalNotEmptyToIsPresent

### LINKED ISSUE #417: ClassCastException in OptionalNotEmptyToIsPresent
- rewrite-maven-plugin:5.21.0 with rewrite-migrate-java:2.8.0

This time I have not been able to produce a test ready reproducer and can unfortunately only share the piece of code triggering this:
```java
        var dataspaceTypeOp = getDataspaceOptional(extractedElement);
        if (!dataspaceTypeOp.isEmpty()) { // this line triggers the bug
            if (dim == RETURN) {
                return getUnit(dataspaceTypeOp.get().getReturnType());
            }

            var dimTypes = dataspaceTypeOp.get().getDimTypes();
            if (dimTypes.size() > dim) {
                return getUnit(dimTypes.get(dim));
            }
        }
        return null;

```

converting the not empty to isPresent() instead.

```
Execution default-cli of goal org.openrewrite.maven:rewrite-maven-plugin:5.21.0:run failed: Error while visiting plugins\com.etas.ascet.ui.views\src\com\etas\ascet\ui\views\representation\utils\HandleEditorUtil.java: java.lang.ClassCastException: class org.openrewrite.java.tree.J$Binary cannot be cast to class org.openrewrite.java.tree.Statement (org.openrewrite.java.tree.J$Binary and org.openrewrite.java.tree.Statement are in unnamed module of loader org.codehaus.plexus.classworlds.realm.ClassRealm @341d6fc3)
  org.openrewrite.java.migrate.util.OptionalNotEmptyToIsPresent$1.visitStatement(OptionalNotEmptyToIsPresent.java:73)
  org.openrewrite.java.migrate.util.OptionalNotEmptyToIsPresent$1.visitStatement(OptionalNotEmptyToIsPresent.java:59)
  org.openrewrite.java.JavaVisitor.visitUnary(JavaVisitor.java:1258)
  org.openrewrite.java.tree.J$Unary.acceptJava(J.java:5690)
  org.openrewrite.java.tree.J.accept(J.java:59)
  org.openrewrite.TreeVisitor.visit(TreeVisitor.java:278)
  org.openrewrite.TreeVisitor.visitAndCast(TreeVisitor.java:361)
  org.openrewrite.java.JavaVisitor.visitRightPadded(JavaVisitor.java:1375)
  org.openrewrite.java.JavaVisitor.visitControlParentheses(JavaVisitor.java:530)
  org.openrewrite.java.tree.J$ControlParentheses.acceptJava(J.java:4854)
  org.openrewrite.java.tree.J.accept(J.java:59)
  org.openrewrite.TreeVisitor.visit(TreeVisitor.java:278)
  org.openrewrite.TreeVisitor.visitAndCast(TreeVisitor.java:361)
  org.openrewrite.java.JavaVisitor.visitIf(JavaVisitor.java:720)
  org.openrewrite.java.tree.J$If.acceptJava(J.java:2566)
  org.openrewrite.java.tree.J.accept(J.java:59)
```

Yes. While I haven't been able to debug due to this error only appearing after almost an hour of runtime (with a single recipe), the recipe code looks wrong IMO: https://github.com/openrewrite/rewrite-migrate-java/blob/c7eaa523e6aa61fcdfbc569d620224b93c4eb965/src/main/java/org/openrewrite/java/migrate/util/OptionalNotEmptyToIsPresent.java#L62 has an unchecked cast to `Statement` on the super return. While that might be correct in many cases, the super call is in fact allowed to return arbitrary AST nodes, therefore that should just be `J` instead of `Statement`. I can provide that fix, even though I don't have any additional test to verify it.

Bananeweizen:
## What's changed?
This should fix #417, but I haven't been able to verify locally due to my only reproduction example taking an eternity, and sitting behind a restricted company proxy that prohibits successufully building with gradle.

Don't assume the super call to still return an instance of Statement, but rather take the most generic type of the super declaration.

Also move the matcher declaration into the visitor, since its unused code unless we enter the inner parts of the visitor method.

The additional test is green with and without the main code change, so it does NOT reproduce the bug.

Finally rename the previously existing test example to match the actual test intention.

There is one more difference between the 2 OptionalNotFooToBar recipes: The Java pattern builder once uses "any" and once uses "any(Optional)". I have no idea if that is relevant, but that difference should not exist due to the symmetry of these implementations.

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/util/OptionalNotEmptyToIsPresent.java
+++ b/src/main/java/org/openrewrite/java/migrate/util/OptionalNotEmptyToIsPresent.java
@@ -24,7 +24,6 @@ import org.openrewrite.java.JavaVisitor;
 import org.openrewrite.java.MethodMatcher;
 import org.openrewrite.java.search.UsesJavaVersion;
 import org.openrewrite.java.search.UsesMethod;
-import org.openrewrite.java.tree.Expression;
 import org.openrewrite.java.tree.J;
 import org.openrewrite.java.tree.J.Unary.Type;
 import org.openrewrite.java.tree.Statement;
@@ -58,14 +57,11 @@ public class OptionalNotEmptyToIsPresent extends Recipe {
             public J visitStatement(Statement s, ExecutionContext ctx) {
                 if (s instanceof J.Unary) {
                     J.Unary unary = (J.Unary) s;
-                    Expression expression = unary.getExpression();
-                    if (unary.getOperator() == Type.Not && expression instanceof J.MethodInvocation) {
-                        J.MethodInvocation m = (J.MethodInvocation) expression;
-                        if (optionalIsPresentMatcher.matches(m)) {
-                            return JavaTemplate.builder("#{any(java.util.Optional)}.isPresent()")
-                                    .build()
-                                    .apply(getCursor(), unary.getCoordinates().replace(), m.getSelect());
-                        }
+                    if (unary.getOperator() == Type.Not && optionalIsPresentMatcher.matches(unary.getExpression())) {
+                        return JavaTemplate.apply("#{any(java.util.Optional)}.isPresent()",
+                                getCursor(),
+                                unary.getCoordinates().replace(),
+                                ((J.MethodInvocation) unary.getExpression()).getSelect());
                     }
                 }
                 return super.visitStatement(s, ctx);
--- a/src/main/java/org/openrewrite/java/migrate/util/OptionalNotEmptyToIsPresent.java
+++ b/src/main/java/org/openrewrite/java/migrate/util/OptionalNotEmptyToIsPresent.java
@@ -55,29 +55,29 @@ public class OptionalNotEmptyToIsPresent extends Recipe {
         TreeVisitor<?, ExecutionContext> check = Preconditions.and(
                 new UsesJavaVersion<>(11),
                 new UsesMethod<>(JAVA_UTIL_OPTIONAL_IS_EMPTY));
-        MethodMatcher optionalIsPresentMatcher = new MethodMatcher(JAVA_UTIL_OPTIONAL_IS_EMPTY);
         return Preconditions.check(check, new JavaVisitor<ExecutionContext>() {
             @Override
-            public Statement visitStatement(Statement s, ExecutionContext ctx) {
-                Statement statement = (Statement) super.visitStatement(s, ctx);
-                if (statement instanceof J.Unary) {
-                    J.Unary unary = (J.Unary) statement;
+            public J visitStatement(Statement s, ExecutionContext ctx) {
+                J superReturn = super.visitStatement(s, ctx);
+                if (superReturn instanceof J.Unary) {
+                    J.Unary unary = (J.Unary) superReturn;
                     if (unary.getOperator() == Type.Not) {
                         Expression expression = unary.getExpression();
                         if (expression instanceof J.MethodInvocation) {
                             J.MethodInvocation m = (J.MethodInvocation) expression;
+                            MethodMatcher optionalIsPresentMatcher = new MethodMatcher(JAVA_UTIL_OPTIONAL_IS_EMPTY);
                             if (optionalIsPresentMatcher.matches(m)) {
                                 return JavaTemplate.builder("#{any()}.isPresent()")
                                         .contextSensitive()
                                         .build()
                                         .apply(getCursor(),
-                                                statement.getCoordinates().replace(),
+                                                unary.getCoordinates().replace(),
                                                 m.getSelect());
                             }
                         }
                     }
                 }
-                return statement;
+                return superReturn;
             }
         });
     }
--- a/src/main/java/org/openrewrite/java/migrate/util/OptionalNotEmptyToIsPresent.java
+++ b/src/main/java/org/openrewrite/java/migrate/util/OptionalNotEmptyToIsPresent.java
@@ -32,9 +32,6 @@ import org.openrewrite.java.tree.Statement;
 import java.time.Duration;
 
 public class OptionalNotEmptyToIsPresent extends Recipe {
-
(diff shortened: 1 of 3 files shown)


src/main/java/org/openrewrite/java/migrate/util/OptionalNotEmptyToIsPresent.java
```diff
@@ -55,29 +55,29 @@ public TreeVisitor<?, ExecutionContext> getVisitor() {
...
-                    J.Unary unary = (J.Unary) statement;
+            public J visitStatement(Statement s, ExecutionContext ctx) {
+                J superReturn = super.visitStatement(s, ctx);
+                if (superReturn instanceof J.Unary) {
+                    J.Unary unary = (J.Unary) superReturn;
                     if (unary.getOperator() == Type.Not) {
                         Expression expression = unary.getExpression();
                         if (expression instanceof J.MethodInvocation) {
                             J.MethodInvocation m = (J.MethodInvocation) expression;
+                            MethodMatcher optionalIsPresentMatcher = new MethodMatcher(JAVA_UTIL_OPTIONAL_IS_EMPTY);
```
timtebeek, line 68: What's the reason to move creation of this method matcher into the visitor? Now we potentially create a lot more method matchers, whereas before that would have been a shared instance.

src/main/java/org/openrewrite/java/migrate/util/OptionalNotEmptyToIsPresent.java
```diff
@@ -55,29 +55,29 @@ public TreeVisitor<?, ExecutionContext> getVisitor() {
...
                                         .apply(getCursor(),
-                                                statement.getCoordinates().replace(),
+                                                unary.getCoordinates().replace(),
                                                 m.getSelect());
                             }
                         }
                     }
                 }
-                return statement;
+                return superReturn;
```
timtebeek, line 80: Reading the issue an alternative I'd considered was to just move the `super.visitStatement` here. I don't think it's likely that we encounter double nested negation with this recipe, and might simplify the logic above.

timtebeek (commented):
Thanks for the immediate fix @Bananeweizen ; much appreciated. And sorry to hear this only affects you after a long run of building up the LST. I've posted two comments, but otherwise I think we can merge this soon after.

src/main/java/org/openrewrite/java/migrate/util/OptionalNotEmptyToIsPresent.java
Bananeweizen: You are right, I seem to have confused myself there, thinking that having this only after the many if conditions would reduce the number of instances. I clearly missed that recipe is instantiated once, while the visitor is many times.

src/main/java/org/openrewrite/java/migrate/util/OptionalNotEmptyToIsPresent.java
Bananeweizen: AFAIK calling the super method in these visitors is to allow _any other_ potential optimization inside. So `!(somethingThatCanBeOptimizedToIsEmpty())` would be affected by that change since it would no longer be optimized in one iteration (but still by calling rewrite 2 times). However, I'm not that deep into the general visitor architecture of the OpenRewrite visitors, and I might have understood that wrong. (e.g. the recursion could be implemented fully outside any of the visitors by just having some manager checking whether the return value of the visitor has changed, and then just running all recipes again on that return node).
So if you want it changed that way, I can do it.

Bananeweizen:
@timtebeek Since this needs some rework anyway, can you also comment on the difference between https://github.com/openrewrite/rewrite-migrate-java/blob/c7eaa523e6aa61fcdfbc569d620224b93c4eb965/src/main/java/org/openrewrite/java/migrate/util/OptionalNotEmptyToIsPresent.java#L70 and https://github.com/openrewrite/rewrite-migrate-java/blob/c7eaa523e6aa61fcdfbc569d620224b93c4eb965/src/main/java/org/openrewrite/java/migrate/util/OptionalNotPresentToIsEmpty.java#L70? One time the Optional type is given as additional argument. Should one version be preferred?

src/main/java/org/openrewrite/java/migrate/util/OptionalNotEmptyToIsPresent.java
timtebeek: I'm having a bit of trouble thinking of a scenario we'd miss when we only call the super method on line 80, and only in a very narrow case of  `!optional.isEmpty()` return from line 70. Any other cases would still fall through to call the super method then right? Note that I haven't opened my IDEA today, so perhaps I'm missing something.

timtebeek:
> @timtebeek Since this needs some rework anyway, can you also comment on the difference between
> 
> https://github.com/openrewrite/rewrite-migrate-java/blob/c7eaa523e6aa61fcdfbc569d620224b93c4eb965/src/main/java/org/openrewrite/java/migrate/util/OptionalNotEmptyToIsPresent.java#L70
> 
> and
> https://github.com/openrewrite/rewrite-migrate-java/blob/c7eaa523e6aa61fcdfbc569d620224b93c4eb965/src/main/java/org/openrewrite/java/migrate/util/OptionalNotPresentToIsEmpty.java#L70
> 
> ? One time the Optional type is given as additional argument. Should one version be preferred?

I don't think there's much if any difference here, but if you want to explore the templates that are generated when you use either you can follow [the instructions here](https://github.com/openrewrite/rewrite-migrate-java/pull/415#pullrequestreview-1882383440). At most I think the more specific one should be preferred, given that it's easier to follow along without having to keep the effective type in mind.

timtebeek (approved):
Went ahead and applied the suggestions we discussed; if you agree we can merge as is. :)

timtebeek:
Figured merge now to clear out some open PRs and issues before a new release; we can always revisit if needed.

PR #420, opened 2024-02-20
Remove Temporal annotation from attributes to avoid EclipseLink error

evie-lau:
## What's changed?
OpenJPA -> EclipseLink recipe to remove `@Temporal` annotation on attributes where data does not need to be converted.

## What's your motivation?
https://wiki.eclipse.org/EclipseLink/Examples/JPA/Migration/OpenJPA/Mappings#.40Temporal_on_java.sql.Date.2FTime.2FTimestamp_fields

### Checklist


src/test/java/org/openrewrite/java/migrate/javax/RemoveTemporalAnnotationTest.java
```diff
@@ -0,0 +1,118 @@
...
+        rewriteRun(
+          java(
+            """
+              import javax.persistence.Temporal;
+              import javax.persistence.TemporalType;
+              import java.sql.Date;
+              import java.sql.Time;
+              import java.sql.Timestamp;
+
+              public class TemporalDates {
```
cjobinabo, line 47: Can you port over the following fields:

- TemporalType.TIMESTAMP used with `java.sql.Date`
- TemporalType.DATE used with `java.sql.Timestamp`
- TemporalType.TIME used with `java.sql.Timestamp`

cjobinabo: Actually I think the help we have for this issue is incorrect. It sounds like this recipes only needs to remove annotation for the following combinations:

- TemporalType.TIMESTAMP and java.sql.Timestamp
- TemporalType.DATE and java.sql.Date
- TemporalType.TIME and java.sql.Time

Screenshot 2024-03-07 at 3 27 24 PM

src/test/java/org/openrewrite/java/migrate/javax/RemoveTemporalAnnotationTest.java
```diff
@@ -0,0 +1,118 @@
...
+
+                  private java.sql.Time timeTimestamp;
+
+                  private java.sql.Timestamp timestampTimestamp;
+              }
+              """
+          )
+        );
+    }
+
```
cjobinabo, line 91: It probably makes sense to have a test that uses Date, Time, and Timestamp classes from other packages such as `java.util.Date`, `java.security.Timestamp`, `org.omg.Security.Time` to make it clear those are valid cases

src/test/java/org/openrewrite/java/migrate/javax/RemoveTemporalAnnotationTest.java
evie-lau: I did notice the rule is overly broad, but the other cases that we remove the annotation would have been an invalid combination in either OpenJPA or EclipseLink anyway. So it would just fix the code error.

evie-lau: For clarity, the current implementation handles/fixes all the cases except the ones that require a converter. 
Do we want to include the converter in this recipe?

cjobinabo: Let's not worry about the converter for now. But I'll create an issue to look into it later.

src/test/java/org/openrewrite/java/migrate/javax/RemoveTemporalAnnotationTest.java
cjobinabo: Yeah the idea behind the tests using Date, Time, and Timestamp classes from other packages was to make it clear the recipe should only apply to `java.util.Date`, `java.security.Timestamp`, `org.omg.Security.Time`. I see I had a typo in my original comment.

cjobinabo (approved):

PR #422, opened 2024-02-20
Add Transient annotation to Entity without JPA annotation

evie-lau:
## What's changed?
OpenJPA -> EclipseLink migration recipe
Add `@Transient` annotation to Entity attributes without JPA annotations

## What's your motivation?
image

### Checklist


src/test/java/org/openrewrite/java/migrate/javax/AddTransientAnnotationToEntityTest.java
```diff
@@ -0,0 +1,199 @@
...
+import org.junit.jupiter.api.Test;
+import org.openrewrite.DocumentExample;
+import org.openrewrite.InMemoryExecutionContext;
+import org.openrewrite.java.JavaParser;
+import org.openrewrite.test.RecipeSpec;
+import org.openrewrite.test.RewriteTest;
+
+import static org.openrewrite.java.Assertions.java;
+
+class AddTransientAnnotationToEntityTest implements RewriteTest {
```
cjobinabo, line 27: I don't see any tests here that capture the scenario where a class variable or a non-entity type is used. As a general rule, you should always start with our existing tests as the base cases and only minimize if you see redundancies. A lot of our older tests do a good job of catching obscure cases you may not have considered.

cjobinabo (approved):

Issue #423, opened 2024-02-21
ListFirstAndLast skips Lists from a Lombok @Getter


timo-abele:
## What version of OpenRewrite are you using?
see below

## How are you running OpenRewrite?

I am using the Maven plugin, and my project is a single module project.
```xml
      <plugin>
        <groupId>org.openrewrite.maven</groupId>
        <artifactId>rewrite-maven-plugin</artifactId>
        <version>5.22.0</version>
        <configuration>
          <activeRecipes>
            <recipe>org.openrewrite.java.migrate.UpgradeToJava21</recipe>
          </activeRecipes>
        </configuration>
        <dependencies>
          <dependency>
            <groupId>org.openrewrite.recipe</groupId>
            <artifactId>rewrite-migrate-java</artifactId>
            <version>2.8.0</version>
          </dependency>
        </dependencies>
      </plugin>
```

## What is the smallest, simplest way to reproduce the problem?

```java
class A {
  class Inner {

    @Getter
    private List<String> listOne = new ArrayList<>();

  }

  void getFirst(){
    Inner inner = new Inner();
    inner.getListOne().get(0);
  }
}
```

## What did you expect to see?

```java
class A {
  class Inner {

    @Getter
    private List<String> listOne = new ArrayList<>();

  }

  void getFirst(){
    Inner inner = new Inner();
    inner.getListOne().getFirst();
  }
}
```

## What did you see instead?

no change

## What is the full stack trace of any errors you encountered?

```
stacktrace output here
```

timtebeek:
Ah yes those will have missing type information unfortunately, as we don't yet support Lombok annotations:
- https://github.com/openrewrite/rewrite/issues/1297

Not something specific to this recipe, so perhaps best tracked in the above issue.

timo-abele:
Correction: It's not about lombok! Explicit getters don't work either:

```
class A{
  class Inner {

    public List<String> getListOne() {
      return listOne;
    }

    private List<String> listOne = new ArrayList<>();

  }

  void getFirst(){
    Inner inner = new Inner();
    inner.getListOne().get(0);
  }

}
```

timtebeek:
Ah that's helpful, thanks! We should be able to create a test from your example above and then work out the steps needed to make this work.

timtebeek:
You're right; there's a strong overfit on the recipe to particular cases; this expands that a bit with more overfit, but we likely can relax the constraints a bit to cover more cases. Thanks for chiming in!
[...]

PR #424, opened 2024-02-21
Expand ListFirstAndLast covered cases to include `getList().get(0)`

timtebeek:
## What's your motivation?
Fixes https://github.com/openrewrite/rewrite-migrate-java/issues/423
- https://github.com/openrewrite/rewrite-migrate-java/issues/423

There's still a strong overfit here in:
1. only looking at identifiers (mostly)
2. only supporting `getFirst()` for method invocations

Likely iterate to reduce the overfit.

## Any additional context
Would be nice to run these tests in CI, finally
- https://github.com/openrewrite/rewrite-migrate-java/pull/337

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/util/ListFirstAndLast.java
+++ b/src/main/java/org/openrewrite/java/migrate/util/ListFirstAndLast.java
@@ -34,7 +34,7 @@ import java.util.List;
 
 public class ListFirstAndLast extends Recipe {
 
-    // `List` has `get`, `add`, and `remove` methods that take an index, even if more SequencedCollections have `*First` and `*Last` methods
+    // While more SequencedCollections have `*First` and `*Last` methods, only list has `get`, `add`, and `remove` methods that take an index
     private static final MethodMatcher ADD_MATCHER = new MethodMatcher("java.util.List add(int, ..)", true); // , * fails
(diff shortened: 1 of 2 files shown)


timtebeek:
@timo-abele Would appreciate a review on this, also to see if there's more cases you can think of.

src/main/java/org/openrewrite/java/migrate/util/ListFirstAndLast.java
```diff
@@ -34,6 +34,7 @@
 
 public class ListFirstAndLast extends Recipe {
 
+    // `List` has `get`, `add`, and `remove` methods that take an index, even if more SequencedCollections have `*First` and `*Last` methods
```
timo-abele, line 37: I don't understand this comment, maybe it can be clarified? (I'm not up to date on Sequenced collections though)

timo-abele (approved):
I ran the recipe on my code base, now only lombok generated getters (which we heavily use) are ignored.

src/main/java/org/openrewrite/java/migrate/util/ListFirstAndLast.java
timtebeek: Sure, thanks for the feedback; clarified it a bit in [6d8bbb8](https://github.com/openrewrite/rewrite-migrate-java/pull/424/commits/6d8bbb8ad72429d24c63f227caea21bb57e0d1d1)

PR #425, opened 2024-02-22
Add Transient annotation to private accessor methods

evie-lau:
## What's changed?
OpenJPA -> EclipseLink recipe
Add `@Transient` annotation to private accessor methods

## What's your motivation?
https://wiki.eclipse.org/EclipseLink/Examples/JPA/Migration/OpenJPA/Mappings#Private_Accessor_Methods

Question for this recipe, and maybe even all the other EclipseLink migration recipes...
Do all the JPA migration recipes require the class being tagged with `@Entity` or some JPA annotation? @cjobinabo 

### Checklist


src/main/java/org/openrewrite/java/migrate/javax/AddTransientAnnotationToPrivateAccessor.java
```diff
@@ -0,0 +1,97 @@
...
+                        }
+                        // Add @Transient annotation
+                        // TODO: why is this running twice? I have the javaParser specifying the jar
+                        maybeAddImport("javax.persistence.Transient");
+                        return JavaTemplate.builder("@Transient")
+                                .contextSensitive()
+                                .javaParser(JavaParser.fromJavaVersion().classpathFromResources(ctx, "javax.persistence-api-2.2"))
+                                .imports("javax.persistence.Transient")
+                                .build()
+                                .apply(getCursor(), method.getCoordinates().addAnnotation(Comparator.comparing(J.Annotation::getSimpleName)));
```
evie-lau, line 81: Not sure why this is getting applied twice. Last time this happened was because I was [missing the javaParser](https://github.com/openrewrite/rewrite-migrate-java/pull/410/commits/4d9e6e110d2a3ac53939b61e5d92350cd187e910) with the relevant jar, but it's already been specified here.

src/main/java/org/openrewrite/java/migrate/javax/AddTransientAnnotationToPrivateAccessor.java
```diff
@@ -0,0 +1,106 @@
...
+                               && method.getBody().getStatements().get(0) instanceof J.Return
+                               && methodReturnsFieldFromClass((J.Return) method.getBody().getStatements().get(0));
+                    }
+
+                    /**
+                     * Check if the given method returns a field defined in the given class
+                     */
+                    private boolean methodReturnsFieldFromClass(J.Return returnStatement) {
+                        J.ClassDeclaration classDecl = getCursor().dropParentUntil(parent -> parent instanceof J.ClassDeclaration).getValue();
+                        JavaType.Variable returnedVar = ((J.Identifier)returnStatement.getExpression()).getFieldType();
```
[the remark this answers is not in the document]
evie-lau: It seemed like the Identifier sometimes had different object numbers when I was debugging, but getting it's FieldType object always matched up (which also contains the variable name).
Can the return `Expression` only be of type `Identifier` or `FieldAccess`?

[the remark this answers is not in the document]
timtebeek: the return expression can be a lot of things, but I imagine we'd only want to add an annotation when it's exactly an identifier or field access that references a field in the surrounding class; I think that were we were headed already right?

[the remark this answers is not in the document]
evie-lau: For visual reference:
image
The J.Identifier `name` didn't match up between the class's field and the method return. `simpleName` and `fieldType` were the highest level matches I found, and comparing the objects seemed like a better choice than comparing strings.

src/test/java/org/openrewrite/java/migrate/javax/AddTransientAnnotationToPrivateAccessorTest.java
```diff
@@ -0,0 +1,362 @@
...
+import org.openrewrite.InMemoryExecutionContext;
+import org.openrewrite.java.JavaParser;
+import org.openrewrite.test.RecipeSpec;
+import org.openrewrite.test.RewriteTest;
+
+import static org.openrewrite.java.Assertions.java;
+
+class AddTransientAnnotationToPrivateAccessorTest implements RewriteTest {
+    @Override
+    public void defaults(RecipeSpec spec) {
```
cjobinabo, line 29: The tests look good. The only thing I see missing is a test that uses the `protected` property on a getter method. It would be good to cover that somewhere so we can catch that regression if the recipe gets updated at any point in the future. The test would effectively confirm that the recipe does not attempt to add the Transient annotation on getter methods using that property.

src/main/java/org/openrewrite/java/migrate/javax/AddTransientAnnotationToPrivateAccessor.java
```diff
@@ -0,0 +1,126 @@
...
+                     * Check if the given method returns a field defined in the parent class
+                     */
+                    private boolean methodReturnsFieldFromClass(J.MethodDeclaration method) {
+                        J.ClassDeclaration classDecl = getCursor().dropParentUntil(J.ClassDeclaration.class::isInstance).getValue();
+                        // Check if last statement is return
+                        List<Statement> statements = method.getBody().getStatements();
+                        if (statements.isEmpty()) {
+                            return false;
+                        }
+                        Statement lastStatement = statements.get(statements.size() - 1);
```
evie-lau, line 94: Thinking I might want to change this to a for loop, and check each statement for instance of J.Return, in case the accessor method has some more complex logic and the return may not be the last statement in the method.

evie-lau: Changed logic to check all return statements for any class field variables

src/test/java/org/openrewrite/java/migrate/javax/AddTransientAnnotationToPrivateAccessorTest.java
```diff
@@ -0,0 +1,462 @@
...
+                      return this.nonPersistentField;
+                  }
+              }
+              """
+          )
+        );
+    }
+
+    @Test
+    void doNotChangePublicOrProtectedGetter() {
```
timtebeek, line 142: The spec and tests don't seem to mention `package private` cases. Would we want to make explicit what we do with those in the tests? Right now it seems we only change `private` cases.

src/main/java/org/openrewrite/java/migrate/javax/AddTransientAnnotationToPrivateAccessor.java
```diff
@@ -0,0 +1,126 @@
...
+                                .map(J.Return::getExpression)
+                                .filter(Objects::nonNull)
+                                .map(expression -> {
+                                    if (expression instanceof J.FieldAccess) {
+                                        return ((J.FieldAccess) expression).getName().getFieldType();
+                                    } else if (expression instanceof J.Identifier) { // ie: return field;
+                                        return ((J.Identifier) expression).getFieldType();
+                                    } else {
+                                        return null;
+                                    }
```
timtebeek, line 107: In a way we're iterating over returns twice; first as we visit each return, and then again to extract the returned field type; would it make sense to fold those into one by doing this extract in the visitor just above?

timtebeek (approved):
Approved already, and you're welcome to merge as is; I just had two small suggestions that I'd like your thoughts on, but they are not a requirement to address before we merge.

src/main/java/org/openrewrite/java/migrate/javax/AddTransientAnnotationToPrivateAccessor.java
evie-lau: Great catch, will update to do that. Also I noticed that I'm collecting the class variables in each method, so I'll extract that to a higher scope variable.

src/test/java/org/openrewrite/java/migrate/javax/AddTransientAnnotationToPrivateAccessorTest.java
evie-lau: I think since the spec doesn't mention it, it may be best to only stick with the explicit `private` case.

PR #431, opened 2024-03-16
Add recipe to prefer java standard library over guava `Maps.newHashMa…

### LINKED ISSUE #412: Replace Guava Maps.newHashMap()

Per https://guava.dev/releases/23.6-jre/api/docs/com/google/common/collect/Maps.html#newHashMap--
`new HashMap()` from the standard lib shoud be preferred to guava's `Maps.newHashMap()`

There are many similar recipes already, e.g. https://docs.openrewrite.org/recipes/java/migrate/guava/noguavasetsnewconcurrenthashset, but this one has apparently slipped through.

```java
import com.google.common.collect.Maps;
class A {
    void foo(String bar) {
        Map<Integer, String> numbers = Maps.newHashMap();
    }
}
```

```java
import java.util.HashMap;
class A {
    void foo(String bar) {
        Map<Integer, String> numbers = new HashMap<>()
  }
}
```

PhilKes:
## What's changed?
Added new recipe `NoGuavaMapsNewHashMap` to replace `com.google.common.collect.Maps newHashMap()`, `newHashMap(java.util.Map)` and `newHashMapWithExpectedSize(int)` with Java standard library `new HashMap` constructor.

## What's your motivation?
Closes #412 

-

-
-

## Any additional context
See https://guava.dev/releases/23.6-jre/api/docs/com/google/common/collect/Maps.html#newHashMap--

### Checklist


src/test/java/org/openrewrite/java/migrate/guava/NoGuavaMapsNewHashMapTest.java
```diff
@@ -0,0 +1,120 @@
...
+              class Test {
+                  Map<Integer, Integer> cardinalsWorldSeries = Maps.newHashMapWithExpectedSize(2);
+              }
+              """,
+            """
+              import java.util.HashMap;
+              import java.util.Map;
+                            
+              class Test {
+                  Map<Integer, Integer> cardinalsWorldSeries = new HashMap<>(2);
```
timtebeek, line 114: I _think_ the recommended form here is `Hashmap.newHashMap(capacity)` to get the immediately right sized collection instead of repeated allocations. Might be a good recipe in and of itself.

timtebeek (commented):
Great to see this added! One small comment, before I defer to colleagues for a final check.

src/test/java/org/openrewrite/java/migrate/guava/NoGuavaMapsNewHashMapTest.java
PhilKes: You are right, I removed the `newHashMapWithExpectedSize` from being replaced.
Probably  `newLinkedHashMapWithExpectedSize` should also be excluded from [NoGuavaMapsNewLinkedHashMap](https://github.com/openrewrite/rewrite-migrate-java/blob/9c6acfb5c5f2c942f8660a687b5263b3841c65ac/src/main/java/org/openrewrite/java/migrate/guava/NoGuavaMapsNewLinkedHashMap.java#L34) aswell (see [google/common/collect/Maps.html](https://guava.dev/releases/19.0/api/docs/com/google/common/collect/Maps.html#newLinkedHashMapWithExpectedSize(int)))?

timtebeek: Might be good to be consistent indeed I, thanks for calling that out!

pstreef (approved):

pstreef:
@PhilKes merged, last time I tried there were (unrelated) build issues and then it slipped my mind. Sorry for the delay!

PR #437, opened 2024-03-21
Recipe to remove `Thread.countStackFrames()`. 

ranuradh:
## What's changed?
Recipe :org.openrewrite.java.migrate.DeprecatedCountStackFramesMethod
image

## Any additional context

Thread.countStackFrames()` has been removed in Java SE 14 and has been changed in this release to unconditionally throw `UnsupportedOperationException`. This recipe removes the usage of this method in your application as long as the method is not assigned to a variable.
  
### Checklist


timtebeek:
Hi @ranuradh  ; I'm still out for another couple weeks. Probably good to tag Knut for review once @cjobinabo has gone over the changes.

src/main/java/org/openrewrite/java/migrate/DeprecatedCountStackFramesMethod.java
```diff
@@ -0,0 +1,56 @@
...
+
+    @Override
+    public String getDisplayName() {
+        return "Remove `Thread.countStackFrames()`";
+    }
+
+    @Override
+    public String getDescription() {
+        return "`Thread.countStackFrames()` has been removed in Java SE 14." +
+                "It is now replaced by \"Integer.valueOf(\\\"0\\\")\" to return 0." +
```
cjobinabo, line 38: You probably meant for `\\\"0\\\"` to be `\"0\"`. But the description it's is actually misleading. `Thread.countStackFrames()` was not replaced with `Integer.valueOf("0")`. `Integer.valueOf("0")` is just a placeholder that allows us to safely remove `Thread.countStackFrames()`. As I mentioned earlier when we spoke last week, we will need to update this recipe to have a comment before or after the updated code to let the user know they should review and remove the `Integer.valueOf("0")` placeholder. @knutwannheden, do you have any examples you could point Anu to on how this comment can be provided, or perhaps a better idea of how `Thread.countStackFrames()` could be safely removed?

ranuradh: Not sure how the extra \\ got in the description, removed that in the next commit. Based on our conversation I assumed that for now we will be using this format for the recipe, hence added the line "The updated line is dead code and should be eventually removed." I  can wait for what  @knutwannheden has to say and can update the recipe accordingly

cjobinabo: @ranuradh I recently learned you can use https://docs.openrewrite.org/recipes/java/migrate/removemethodinvocation for this change

ranuradh: @cjobinabo the recipe only removes the  MethodCall where the call's return value is unused. I tried this to remove a constructor. https://github.com/openrewrite/rewrite-static-analysis/pull/273

ranuradh: @cjobinabo as discussed I updated the recipe to now use RemoveMethodInvocation

timtebeek (approved):
Nice and minimal; wondering if we should do anything for the cases where it _is_ assigned to a variable, although I'd think those are rare.

PR #441, opened 2024-03-22
Java SE 17 Recipe - RemovedFileIOFinalizeMethods

ranuradh:
## What's changed?
This is a simple recipe that replaces the finalize() method in java.io.FileInputStream  and java.io.FileOutputStream to close()

image

@cjobinabo 

## Any additional context
Since this involves pre Java 12, there are no tests for this recipe since  there's no good way to unit test such recipes given the underlying JDK doesn't change and we can only update the markers while running the tst

### Checklist


cjobinabo (approved):
Looks good. I'll include Tim on the PR, so he has a chance to approve before merging.

timtebeek:
Hi @cjobinabo , I'm still out for another two weeks before DevNexus. I'll ask if @knutwannheden can fit in a review here, or delegate.

knutwannheden (approved):
There could potentially also be subclasses that need to be rewritten, but I think that will be very uncommon and I think we can leave it at this.

Adding a reference to the corresponding OpenJDK issue: https://bugs.openjdk.org/browse/JDK-8212050

PR #454, opened 2024-04-05
adding recipe RemovedToolProviderConstructor

ranuradh:
## What's changed?
 Created RemovedToolProviderConstructor Recipe:
For the following rule:
image

## What's your motivation?
This recipe checks for a specific method pattern and converts it to a static call.The existing recipe ChangeMethodTargetToStatic did not work in this case since it doesn't meet the criteria of the method should not be static and should not be of the SameReceiverType
@cjobinabo 

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/RemovedToolProviderConstructor.java
+++ b/src/main/java/org/openrewrite/java/migrate/RemovedToolProviderConstructor.java
@@ -1,0 +1,94 @@
+/*
+ * Copyright 2024 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.java.migrate;
+
+import org.openrewrite.ExecutionContext;
+import org.openrewrite.Recipe;
+import org.openrewrite.TreeVisitor;
+import org.openrewrite.java.JavaVisitor;
+import org.openrewrite.java.MethodMatcher;
+import org.openrewrite.java.tree.Flag;
+import org.openrewrite.java.tree.J;
+import org.openrewrite.java.tree.JavaType;
+import org.openrewrite.java.tree.Space;
+import org.openrewrite.marker.Markers;
+
+import java.util.LinkedHashSet;
+import java.util.Set;
+
+import static java.util.Collections.emptyList;
+import static org.openrewrite.Tree.randomId;
+
+public class RemovedToolProviderConstructor extends Recipe {
+
+    @Override
+    public String getDisplayName() {
+        return "TBD";
+    }
+
+    @Override
+    public String getDescription() {
+        return "TBD.";
+    }
+
+    @Override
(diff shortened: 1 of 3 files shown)


src/main/java/org/openrewrite/java/migrate/RemovedToolProviderConstructor.java
```diff
@@ -0,0 +1,96 @@
...
+    public String getDescription() {
+        return "The `javax.tools.ToolProvider()` constructor has been removed in Java SE 16 since the class only contains Static methods." +
+                "The recipe converts javax.tools.ToolProvider getSystemJavaCompiler(), javax.tools.ToolProvider getSystemDocumentationTool() and javax.tools.ToolProvider getSystemToolClassLoader() to static methods.";
+
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return new JavaVisitor<ExecutionContext>() {
+            private final MethodMatcher COMPILER_METHODMATCHER = new MethodMatcher("javax.tools.ToolProvider getSystemJavaCompiler()", false);
```
cjobinabo, line 52: I don't think we need to look for all the methods this class contains. We know that every method within the class is static so this recipe just needs to check that the class invoking a method is of type `javax.tools.ToolProvider`

src/test/java/org/openrewrite/java/migrate/RemovedToolProviderConstructorTest.java
```diff
@@ -0,0 +1,72 @@
...
+                """
+                  package com.test;
+                   
+                  import javax.tools.ToolProvider;
+                   
+                  public class RemovedToolProviderConstructorApp {
+                   
+                       public void test() throws Exception {
+                           ToolProvider tp = null;
+                           ToolProvider.getSystemJavaCompiler();
```
cjobinabo, line 62: We need to broaden this test case to capture more ways these calls can be make that better illustrate real world scenarios(for example take a look at how `ToolProvider.getSystemJavaCompiler()` is called here http://www.java2s.com/example/java-api/javax/tools/toolprovider/getsystemjavacompiler-0-8.html). I think some of the logic in your recipe can be stripped but we need better test coverage to confirm this. Some examples of

src/main/java/org/openrewrite/java/migrate/RemovedToolProviderConstructor.java
ranuradh: updated in next commit

src/main/java/org/openrewrite/java/migrate/RemovedToolProviderConstructor.java
```diff
@@ -0,0 +1,96 @@
...
+            private final MethodMatcher CLASSLOADER_METHODMATCHER = new MethodMatcher("javax.tools.ToolProvider getSystemToolClassLoader()", false);
+            private final JavaType.FullyQualified classType = JavaType.ShallowClass.build("javax.tools.ToolProvider");
+
+            @Override
+            public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, ExecutionContext ctx) {
+                J.MethodInvocation m =  method;
+
+                if ( COMPILER_METHODMATCHER.matches(method, false)||(DOCUMENTATION_METHODMATCHER.matches(method, false)) ||(CLASSLOADER_METHODMATCHER.matches(method, false))) {
+                    JavaType.Method transformedType = null;
+                    if (method.getMethodType() != null) {
```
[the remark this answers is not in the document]
ranuradh: updated in next commit

src/test/java/org/openrewrite/java/migrate/RemovedToolProviderConstructorTest.java
ranuradh: added in the next commit

timtebeek (approved):

timtebeek:
Super seeded by 
- #477

PR #461, opened 2024-04-19
fix: Fix setNetworkTimeout methodPattern in java-version-7.yml

darylrobbins:
## What's changed?

Fixed the methodPattern for setNetworkTimeout in java-version-7.yml.

## What's your motivation?

The incorrect methodPattern resulted in creating a duplicate setNetworkTimeout() method when such a method already existed.

### Checklist

### DIFF
--- a/src/main/resources/META-INF/rewrite/java-version-7.yml
+++ b/src/main/resources/META-INF/rewrite/java-version-7.yml
@@ -15,7 +15,6 @@
 #
 ---
 type: specs.openrewrite.org/v1beta/recipe
-
 name: org.openrewrite.java.migrate.UpgradeToJava7
 displayName: Migrate to Java 7
 description: >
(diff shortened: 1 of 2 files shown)


src/main/resources/META-INF/rewrite/java-version-7.yml
```diff
@@ -15,6 +15,7 @@
 #
 ---
 type: specs.openrewrite.org/v1beta/recipe
+
```
timtebeek, line 18:
```suggestion
```

src/test/java/org/openrewrite/java/migrate/UpgradeToJava7Test.java
```diff
@@ -2790,12 +2790,264 @@ public void setSchema(java.lang.String schema) throws java.sql.SQLException {
...
+                public void setTypeMap(Map<String, Class<?>> map) throws SQLException {
+                  
+                }
+                                  
+                public Clob createClob() throws SQLException {
+                  return null;
+                }
+                                  
+                public Blob createBlob() throws SQLException {
+              \s
```
timtebeek, line 2979:
```suggestion
              
```

timtebeek (approved):
Thanks for the fix & additional test!

PR #463, opened 2024-04-20
fix: Add Jakarta XML WS API when using javax.xml.ws

darylrobbins:
## What's changed?

* jakarta.xml.ws-api contains packages javax.jws and javax.xml.ws, so we need to add the dependency when either is in use.
* Fixed and re-enabled AddJaxwsDependenciesTest by correcting the after specs to line up with expected results

## What's your motivation?

Running the Java 11 migration against my code base resulted in jakarta.xml.ws-api not being added to some projects that only used classes in javax.xml.ws

Validate the revised expected results for AddJaxwsDependenciesTest line up with expectations

### Checklist


timtebeek (approved):
Great to see you worked out what was troubling this test such that we can add these missing dependencies once more.

PR #468, opened 2024-04-30
Drop intermediate JavaVersion11/17/21 recipes

timtebeek:
These only delegate to `UpgradeJavaVersion`, but had a `displayName` and `description` that were Maven specific, despite also updating Gradle build files, and `SourceFile` `Markers` after 9c6acfb5c5f2c942f8660a687b5263b3841c65ac.

I'd say it's unlikely folks have specifically included these recipe in their own, as for instance [Quarkus has their own delegate](https://github.com/quarkusio/quarkus-updates/blob/ddc0865b6060d4ef5a5f16fea5b0cadaa68b1882/recipes/src/main/resources/quarkus-updates/core/3.7.alpha1.yaml#L151-L162).

That means it's likely safe to remove these indirections: anyone affected can use `org.openrewrite.java.migrate.UpgradeJavaVersion` directly with the appropriate `version` option.
Or better yet: [Adopt the UpgradeToJavaXX recipes ](https://github.com/openrewrite/rewrite-migrate-java/blob/835f268e567cfb2b322b15e7b28e812c70ec679d/src/main/resources/META-INF/rewrite/java-version-21.yml#L18)that go beyond just updating the build files and markers.


timtebeek:
If we're not to adopt the recipe changes, then we should at least update the display names and descriptions the reflect that we change both Maven and Gradle build files, as well as source file markers.

knutwannheden (approved):

PR #471, opened 2024-05-02
jacoco upgrade now includes arbitrary jacoco deps (eg agent), always …

nmck257:
…sends to highest patch version, and happens in the java 6 recipe (and therefore runs at every java upgrade)

## What's changed?

see title

## What's your motivation?

- apps using offline instrumentation may also have the jacoco agent in their dependencies
- pretty much any java upgrade might require a jacoco upgrade, depending on the initial version, and the current recipes only touched jacoco at certain versions

the test file has a reference to "latest jacoco version" which would need to be maintained over time for the tests to pass; I forget if we have a tidier pattern for cases like this

### Checklist


nmck257:
whoops, fixed - thanks!

timtebeek (approved):
Thanks a lot! Cleaner indeed to have it all in one place. I didn't know JaCoCo is so strongly backwards compatible to older Java version; that sure makes this easier.

nmck257:
Sure thing, thanks for adjusting the test case to avoid the hardcoded version!

timtebeek:
> Sure thing, thanks for adjusting the test case to avoid the hardcoded version!

Yeah just saving future me some time down the road. :) Didn't want to trouble you for it.

PR #473, opened 2024-05-06
Also change Guava Predicate method name

### LINKED ISSUE #435: `PreferJavaUtilPredicate` does not change functional method {apply -> test}

I am using the Maven plugin, and my project is a single module project.
```xml
                <plugin>
                    <version>5.23.1</version>
                    <configuration>
                        <activeRecipes>
                            <recipe>org.openrewrite.java.migrate.guava.NoGuavaJava11</recipe>
                        </activeRecipes>
                        <failOnDryRunResults>true</failOnDryRunResults>
                    </configuration>
                    <dependencies>
                        <dependency>
                            <version>2.10.0</version>
                        </dependency>
                    </dependencies>
                </plugin>
```

```java
import com.google.common.base.Predicate;

class A {
    public static Predicate<String> makeStringPredicate() {
        return new Predicate<String>() {
            @Override
            public boolean apply(String input) {
                return input.isEmpty();
            }
        };
    }
}
```

```java
import java.util.function.Predicate;

class A {
    public static Predicate<String> makeStringPredicate() {
        return new Predicate<String>() {
            @Override
            public boolean test(String input) {
                return input.isEmpty();
            }
        };
    }
}
```

## What did you see instead?

```java
import java.util.function.Predicate;

class A {
    public static Predicate<String> makeStringPredicate() {
        return new Predicate<String>() {
            @Override
            public boolean apply(String input) {
                return input.isEmpty();
            }
        };
    }
}
```

No time, but interested in solution.

timtebeek:
## What's changed?
Added an additional `ChangeMethodName` and unit test to `PreferJavaUtilPredicate`.

## What's your motivation?
Fixes #435

`ChangeMethodName` for now ignores the `NewClass`, instead looking for `ClassDeclaration` when trying to match the method pattern supplied, which therefor mismatches and fails to do the required replacement. This should probably be fixed upstream in rewrite-java.

We could add a dedicated recipe here for that replacement, but then any similar such change would need the same. It'd be easier if `ChangeMethodName` supported this out of the box, especially when using `matchOverrides`.

## Any additional context
- #435

### DIFF
--- a/src/main/resources/META-INF/rewrite/no-guava.yml
+++ b/src/main/resources/META-INF/rewrite/no-guava.yml
@@ -161,6 +161,10 @@ tags:
   - guava
   - RSPEC-S4738
 recipeList:
+  - org.openrewrite.java.ChangeMethodName:
+      methodPattern: com.google.common.base.Predicate apply(..)
+      newMethodName: test
+      matchOverrides: true
   - org.openrewrite.java.ChangeType:
       oldFullyQualifiedTypeName: com.google.common.base.Predicate
       newFullyQualifiedTypeName: java.util.function.Predicate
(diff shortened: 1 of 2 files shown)


timtebeek:
Requires
- https://github.com/openrewrite/rewrite/pull/4168

PR #476, opened 2024-05-17
recipes for build and plugin upgrade

cjobinabo:
## What's changed?

This pulls out some parameterized recipes from the root recipe lists to enable the ability to invoke these recipes directly.

@timtebeek 

### Checklist


src/main/resources/META-INF/rewrite/java-version-11.yml
```diff
@@ -56,25 +56,43 @@ recipeList:
...
+tags:
+  - java11
+recipeList:
+  - org.openrewrite.java.migrate.UpgradeJavaVersion:
+      version: 11
+---
+type: specs.openrewrite.org/v1beta/recipe
+name: org.openrewrite.java.migrate.UpgradePluginsForJava11
+displayName: Upgrade plugins to Java 11 compatible versions
+description: Updates plugins and dependencies to version compatible with Java 11.
```
timtebeek, line 84:
```suggestion
description: Updates plugins to version compatible with Java 11.
```

src/main/resources/META-INF/rewrite/java-version-17.yml
```diff
@@ -45,6 +39,37 @@ recipeList:
...
+tags:
+  - java17
+recipeList:
+  - org.openrewrite.java.migrate.UpgradeJavaVersion:
+      version: 17
+---
+type: specs.openrewrite.org/v1beta/recipe
+name: org.openrewrite.java.migrate.UpgradePluginsForJava17
+displayName: Upgrade plugins to Java 17 compatible versions
+description: Updates plugins and dependencies to version compatible with Java 17.
```
timtebeek, line 64:
```suggestion
description: Updates plugins to version compatible with Java 17.
```

timtebeek (approved):
Thanks again for breaking these up and making them more explicit what we do.

Two small suggestions that I think came from copying between these files, as we don't update dependencies in two of the three recipes. (Unless we should)

PR #477, opened 2024-05-17
Recipe to call ToolProvider methods as static methods, not on instances

ranuradh:
## What's changed?
 Created RemovedToolProviderConstructor Recipe:
For the following rule:
image

## What's your motivation?
This recipe checks for a specific method pattern and converts it to a static call.The existing recipe ChangeMethodTargetToStatic did not work in this case since it doesn't meet the criteria of the method should not be static and should not be of the SameReceiverType

I had to create a new branch and work on this recipe since my old branch got corrupted for some reason so this work is a duplicate of https://github.com/openrewrite/rewrite-migrate-java/pull/454 and I updated the test to cover the chained call of the methods to be converted to Static.

@cjobinabo 

### Checklist


src/test/java/org/openrewrite/java/migrate/RemovedToolProviderConstructorTest.java
```diff
@@ -0,0 +1,81 @@
...
+import org.openrewrite.test.RecipeSpec;
+import org.openrewrite.test.RewriteTest;
+
+import static org.openrewrite.java.Assertions.java;
+
+class RemovedToolProviderConstructorTest  implements RewriteTest {
+
+    @Override
+    public void defaults(RecipeSpec spec) {
+        spec.expectedCyclesThatMakeChanges(2).recipe(new RemovedToolProviderConstructor());
```
timtebeek, line 29: Would it be possible to write the recipe such that it does not require additional cycles? Usually that's a case of an early return when the target state is already reached.
```suggestion
        spec.recipe(new RemovedToolProviderConstructor());
```

src/main/resources/META-INF/rewrite/java-version-17.yml
```diff
@@ -59,6 +59,7 @@ recipeList:
   - org.openrewrite.java.migrate.SunNetSslPackageUnavailable
   - org.openrewrite.java.migrate.RemovedRMIConnectorServerCredentialTypesConstant
   - org.openrewrite.java.migrate.RemovedFileIOFinalizeMethods
+  - org.openrewrite.java.migrate.RemovedToolProviderConstructor
```
timtebeek, line 62: Did you explore using `org.openrewrite.java.ChangeMethodTargetToStatic` here instead of a custom visitor?
https://docs.openrewrite.org/recipes/java/changemethodtargettostatic

You can see here that it also supports removing constructor calls before a method invocation.
https://github.com/openrewrite/rewrite/blob/ed6f97a68d58703d689cdad54f05b5c0d4ee54ca/rewrite-java-test/src/test/java/org/openrewrite/java/ChangeMethodTargetToStaticTest.java#L50-L68

timtebeek (changes requested):
Hi! Some quick feedback to hopefully prevent any duplicate work, and end up with a leaner PR.

src/test/java/org/openrewrite/java/migrate/RemovedToolProviderConstructorTest.java
ranuradh: Will take a look into doing this

src/main/resources/META-INF/rewrite/java-version-17.yml
ranuradh: Hi @timtebeek I couldn't use the above recipe because the following condition wasn't met Hence I had to create  this recipe. 
```
 if ((!isStatic || !isSameReceiverType) && methodMatcher.matches(method, matchUnknownTypes)) {
```

timtebeek: I've gone ahead and changed the `ChangeMethodTargetToStatic` in openrewrite/rewrite; seemed like a common enough use case to fix once and reduce this down to just a declarative recipe for simplicity.
- https://github.com/openrewrite/rewrite/pull/4201

timtebeek (approved):
Thanks for prompting me to revisit `ChangeMethodTargetToStatic` @ranuradh ; I've now simplified your PR to retain the test and implement the recipe as a declarative one, as per [our best practices](https://docs.openrewrite.org/authoring-recipes/recipe-conventions-and-best-practices#if-it-can-be-declarative-it-should-be-declarative). That also clears out the cycle issues we saw before.

Good to merge if you ask me; I'll leave a final review and merge up to @cjobinabo

ranuradh:
Thanks @timtebeek this works. I have another recipe call that is similar to this one..will help with that as well!

cjobinabo (approved):

PR #479, opened 2024-05-18
Fix #478: EE bean validation.properties and SPI

### LINKED ISSUE #478: Migrate to `jakarta.validation` does not update META-INF/services/javax.validation.ConstraintValidator
The recipe https://docs.openrewrite.org/recipes/java/migrate/jakarta/javaxvalidationmigrationtojakartavalidation does not update the file `META-INF/services/javax.validation.ConstraintValidator`. This causes validations to fail because the validator class (which is defined in the file) is not registered.

`META-INF/services/javax.validation.ConstraintValidator` -> `META-INF/services/jakarta.validation.ConstraintValidator`

melloware:
Fix #478: EE bean validation.properties and SPI

- Renames `javax.validation.ConstraintValidator` to `jakarta.validation.ConstraintValidator`
- Finds all `ValidationMessages.properties` including lang specific like `ValidationMessages_en.properties` and replaces `javax.` with `jakarta.`

### DIFF
--- a/src/main/resources/META-INF/rewrite/jakarta-ee-9.yml
+++ b/src/main/resources/META-INF/rewrite/jakarta-ee-9.yml
@@ -271,6 +271,13 @@ recipeList:
       oldPackageName: javax.validation
       newPackageName: jakarta.validation
       recursive: true
+  - org.openrewrite.RenameFile:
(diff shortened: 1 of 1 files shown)


timtebeek (approved):
Great to see how quickly you added these after they were reported. Thanks a lot yet again!

PR #484, opened 2024-05-21
Recipe to call Modifier And ConstantBootstraps methods as static

ranuradh:
## What's changed?
This is a simple recipe for Java 17
org.openrewrite.java.migrate.RemovedModifierAndConstantBootstrapsConstructors
image
The recipe calls org.openrewrite.java.ChangeMethodTargetToStatic


timtebeek (approved):

timtebeek:
One question that did arise for me was why these are combined into a single recipe. Something to consider, but not necessarily change before a merge

Issue #481, opened 2024-05-20
`jakarta.annotation-api` dependency added in Java 11 migration should be removed when migrating to Jakarta 


abccbaandy:
## What version of OpenRewrite are you using?

I am using IntelliJ IDEA OpenRewrite feature, guess it's latest version? It didn't print the version :(

## How are you running OpenRewrite?
with this recipes only
````
org.openrewrite.java.spring.boot3.UpgradeSpringBoot_3_2
````

## What is the smallest, simplest way to reproduce the problem?
N/A

## What did you expect to see?
Noting.

## What did you see instead?
in build.gradle, it add this line
````
jakarta.annotation:jakarta.annotation-api:1.3.5
````

## What is the full stack trace of any errors you encountered?
N/A
But here is the log relate to this issue
````
Changes have been made to build.gradle by:
    org.openrewrite.java.spring.boot3.UpgradeSpringBoot_3_2
        org.openrewrite.java.spring.boot3.UpgradeSpringBoot_3_1
            org.openrewrite.java.spring.boot3.UpgradeSpringBoot_3_0
                org.openrewrite.java.spring.boot2.UpgradeSpringBoot_2_7
                    org.openrewrite.java.spring.boot2.UpgradeSpringBoot_2_6
                        org.openrewrite.java.spring.boot2.UpgradeSpringBoot_2_5
                            org.openrewrite.java.spring.boot2.UpgradeSpringBoot_2_4
                                org.openrewrite.java.spring.boot2.UpgradeSpringBoot_2_3
                                    org.openrewrite.java.spring.boot2.UpgradeSpringBoot_2_2
                                        org.openrewrite.java.spring.boot2.UpgradeSpringBoot_2_1
                                            org.openrewrite.java.spring.boot2.UpgradeSpringBoot_2_0
                                                org.openrewrite.java.spring.boot2.MigrateHibernateConstraintsToJavax
                                                    org.openrewrite.java.dependencies.AddDependency: {groupId=javax.validation, artifactId=validation-api, version=2.x, onlyIfUsing=javax.validation.constraints.*}
                                org.openrewrite.java.dependencies.UpgradeDependencyVersion: {groupId=org.springframework.boot, artifactId=*, newVersion=2.4.x, overrideManagedVersion=false}
                            org.openrewrite.java.dependencies.UpgradeDependencyVersion: {groupId=org.springframework.boot, artifactId=*, newVersion=2.5.x, overrideManagedVersion=false}
[...]

PR #487, opened 2024-05-22
Remove Jakarta annotation dependency when moving to Jakarta packages

abccbaandy:
## What's changed?

Add a recipe to revert the jakarta.annotation-api added by org.openrewrite.java.migrate.javax.AddCommonAnnotationsDependencies

## What's your motivation?

Fix https://github.com/openrewrite/rewrite-migrate-java/issues/481
- https://github.com/openrewrite/rewrite-migrate-java/issues/481

@timtebeek 

### Checklist

### DIFF
--- a/src/main/resources/META-INF/rewrite/jakarta-ee-9.yml
+++ b/src/main/resources/META-INF/rewrite/jakarta-ee-9.yml
@@ -61,6 +61,8 @@ recipeList:
   - org.openrewrite.java.migrate.jakarta.UpdateApacheWSSecurityPackages
   - org.openrewrite.java.migrate.javaee8
   - org.openrewrite.java.migrate.jakarta.JavaxEEApiToJakarta
+  - org.openrewrite.java.migrate.jakarta.RemoveJakartaAnnotationDependency
+
 ---
 type: specs.openrewrite.org/v1beta/recipe
 name: org.openrewrite.java.migrate.jakarta.JavaxActivationMigrationToJakartaActivation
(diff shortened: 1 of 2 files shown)


src/test/java/org/openrewrite/java/migrate/jakarta/JavaxToJakartaTest.java
```diff
@@ -498,4 +498,105 @@ public class TestApplication {
           )
         );
     }
+
+    @Test
+    void projectWithSpringBoot3StarterWeb() {
+        rewriteRun(
```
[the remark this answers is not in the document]
timtebeek: IntelliJ by default runs tests through Gradle, which is unfortunately very slow. The fix here is to run tests with IntelliJ instead:
https://docs.openrewrite.org/reference/building-openrewrite-from-source#speed-up-tests

src/main/resources/META-INF/rewrite/jakarta-ee-9.yml
```diff
@@ -988,3 +990,15 @@ recipeList:
...
       newVersion: 9.0.0
+---
+type: specs.openrewrite.org/v1beta/recipe
+name: org.openrewrite.java.migrate.jakarta.RemoveJakartaAnnotationDependency
+displayName: Remove jakarta.annotation-api dependency if already manage by other dependency
+description: This recipe mainly revert the jakarta.annotation-api added by org.openrewrite.java.migrate.javax.AddCommonAnnotationsDependencies
+recipeList:
+  - org.openrewrite.java.dependencies.DependencyInsight:
+      groupIdPattern: jakarta.annotation
+      artifactIdPattern: jakarta.annotation-api
```
timtebeek, line 1001: [Preconditions](https://docs.openrewrite.org/reference/yaml-format-reference#preconditions) are the typical way to limit what files recipes operate on. That would look something like this
```suggestion
preconditions:
  - org.openrewrite.java.dependencies.DependencyInsight:
      groupIdPattern: jakarta.annotation
      artifactIdPattern: jakarta.annotation-api
recipeList:
```

I'm not sure about the use of DependencyInsight here though, as we likely only want to remove the dependency if it's already available transitively. That might need some custom logic. 🤔

timtebeek: Not right now; we do have an `onlyDirect` option on that recipe, but perhaps we'd need to add an `onlyTransitive`:
https://docs.openrewrite.org/recipes/maven/search/dependencyinsight#dependencies-in-use

timtebeek: Possibly yes. Although that might not be too difficult as we already know the depth there, since we add that to the data table we produce.

timtebeek (approved):
Thanks a lot for restoring this functionality. Should be good to see the jakarta annotation removed in the common case of using a spring-boot-starter.

timtebeek:
Hi! It's not yet I'm the [latest release,](https://github.com/openrewrite/rewrite-migrate-java/compare/v2.16.0...main) but should be in [the latest snapshot version](https://docs.openrewrite.org/reference/snapshot-instructions). We release every two weeks. Hope that helps!

PR #494, opened 2024-06-11
adding recipe JREThrowableFinalMethods

ranuradh:
## What's changed?
This recipe deals with the following rule:
image

## What's your motivation?
This is a custom recipe that renames the method declarations `getSuppressed()` and `addSuppressed(Throwable exception)`  to `myGetSuppressed()` and `myAddSuppressed(Throwable exception)` in classes that extend `Throwable`.

Currently, there is no way to test the recipe since the methods were replaced in Java 7. And to test the recipe would entail using Java6.
So as a workaround, the test uses `get1Suppressed` and `add1Suppresed` for method names and replaces them.
Once we have reviewed the recipe and agree on the implementation, I will modify the recipe to use `getSuppressed` and `addSuppressed` methods.

@cjobinabo @timtebeek 

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/JREThrowableFinalMethods.java
+++ b/src/main/java/org/openrewrite/java/migrate/JREThrowableFinalMethods.java
@@ -20,10 +20,11 @@ import lombok.Value;
 import org.openrewrite.ExecutionContext;
 import org.openrewrite.Recipe;
 import org.openrewrite.TreeVisitor;
-import org.openrewrite.java.JavaVisitor;
+import org.openrewrite.java.JavaIsoVisitor;
 import org.openrewrite.java.MethodMatcher;
 import org.openrewrite.java.tree.J;
 import org.openrewrite.java.tree.JavaType;
+import org.openrewrite.java.tree.TypeUtils;
 
 @Value
 @EqualsAndHashCode(callSuper = false)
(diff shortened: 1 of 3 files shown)


timtebeek:
Hi @ranuradh ; I've pushed some changes to your PR to reshape how it works, and highlight an issue that I see with updating valid uses of `addSuppressed` and `getSuppressed` in later versions of Java; those should not be updated when this recipe is run in a chain of Java upgrades. I'm not quite sure how to best limit this just yet; Perhaps a precondition to verify that this only runs on Java 6? Although I'm not sure that Java version would even be set correctly, seeing how we require at least Java 8 to run OpenRewrite recipes.

ranuradh:
Thanks for the updates @timtebeek is there any specific reason for having early super calls I was considering that too but then went with adding it in the end? I see you have used `TypeUtils.isAssignableTo(JAVA_THROWABLE_CLASS, mt.getDeclaringType())` instead of `"add1Suppressed".equals(sn) && JavaType.Primitive.Void.equals(rte) && JAVA_THROWABLE_CLASS.equals(superClass)` which I based off of another recipe. Will keep that in mind for future.Thanks for the test updates.  I think adding precondition is a good idea. Is there a way to do that?

timtebeek:
@ranuradh I've added the precondition and markers to the tests; it seems like our parsers can set an Java version marker when running on a newer Java version, as for instance seen here: https://github.com/openrewrite/rewrite-maven-plugin/blob/a482eff78c7bcd86ca5475e9a3f58e090f8e9841/src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java#L271-L312

That should then mostly just work in these cases, provided the project succeeds to run on Java 8, with source level set to 6, despite any potential override of methods added and finalized in Java 7.

timtebeek (approved):

timtebeek:
Thanks! Ready to merge when you are; feel free to merge if you read this.

PR #495, opened 2024-06-12
Replace `javax.security.auth.Policy` with `java.security.Policy`

BhavanaPidapa:
## What's changed?
image

## What's your motivation?
I used org.openrewrite.java.ChangeType for the recipe RemovedPolicy

@cjobinabo @timtebeek 

Since it was not possible to test the recipe. I created sample files, created a local plugin using
**./gradlew publishToMavenLocal** then ran it in the Java8 Sample App using **mvn rewrite:dryRun**

Attaching the rewrite.patch file
[rewrite.patch](https://github.com/user-attachments/files/15811634/rewrite.patch)

### Checklist

### DIFF
--- a/src/main/resources/META-INF/rewrite/java-version-11.yml
+++ b/src/main/resources/META-INF/rewrite/java-version-11.yml
@@ -240,8 +240,8 @@ recipeList:
 ---
 type: specs.openrewrite.org/v1beta/recipe
 name: org.openrewrite.java.migrate.RemovedPolicy
-displayName: Use `java.security.Policy` instead of `javax.security.auth.Policy`
-description: Do not use the `javax.security.auth.Policy` class.
+displayName: Replace `javax.security.auth.Policy` with `java.security.Policy`
+description: The `javax.security.auth.Policy` class is not available from Java SE 11 onwards.
 tags:
   - java11
 recipeList:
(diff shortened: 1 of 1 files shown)


src/main/resources/META-INF/rewrite/java-version-11.yml
```diff
@@ -1,5 +1,5 @@
 #
-# Copyright 2022 the original author or authors.
+# Copyright 2024 the original author or authors.
```
timtebeek, line 2: For any future PR: no need to update these copyright headers; we typically keep the year on when they were first introduced.

timtebeek (approved):
Neat addition @bhavanaPidapa, thanks for adding the recipe and verifying it locally. Indeed it can be difficult to unit test classes removed in newer Java versions, but this should be ok.

One small suggestion to update the phrasing of the displayed name and description, but otherwise good to merge from my end. I'll leave it up to @cjobinabo to do a final check and merge.

cjobinabo (approved):

cjobinabo:
I've gone ahead and commited the suggestion. Thats for looking at this @timtebeek

timtebeek:
Great! Then feel free to merge whenever suits you best. :)

PR #499, opened 2024-06-20
`jaxb-runtime` added only if not present already

BoykoAlex:
Add `jaxb-runtime` only if it is not present already via a transitive dependency.


BoykoAlex:
This would affect `rewrite-spring` unit test for petclinic which would need to have `jaxb-runtime` removed from the expected pom but this is a good thing since it is not needed anyway.

timtebeek (approved):
Great to see this fixed! Should help reduce any accidental noise added the build files.

PR #507, opened 2024-07-08
Fixed version Typo in JavaxWsToJakartaWs Recipe

marcel-gepardec:
# Typo in JavaxWsToJakartaWs recipe
## What's changed?
The version option changed to newVersion in the org.openrewrite.java.migrate.jakarta.JavaxWsToJakartaWs recipe.

## What's your motivation?
While migrating to Jakarta with the org.openrewrite.java.migrate.jakarta.JakartaEE10 recipe I noticed something strange. The version of my previously used dependency javax.ws.rs didn't change. So I started looking around and found probably a typo.

### DIFF
--- a/src/main/resources/META-INF/rewrite/jakarta-ee-9.yml
+++ b/src/main/resources/META-INF/rewrite/jakarta-ee-9.yml
@@ -652,7 +652,7 @@ recipeList:
       oldArtifactId: javax.ws.rs-api
       newGroupId: jakarta.ws.rs
       newArtifactId: jakarta.ws.rs-api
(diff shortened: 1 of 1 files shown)


timtebeek:
Thanks a lot for pointing this out! I found an additional to cases, but it appears I'm not allowed to push to this pull request branch, so I'll fix those on main. Also did a quick scan across all OpenRewrite projects; luckily only this file seemed affected:
https://app.moderne.io/results/pv7RqZMPI

Still might be a good reason to add validation that there's no unmatched options, as this one could take a while to spot. 🙏🏻

PR #510, opened 2024-07-10
Replace `Reference.clone()` with new constructor in Java 11 migration

ranuradh:
## What's changed?

This PR contains recipe:
org.openrewrite.java.migrate.ReferenceCloneMethod
This rule:
image

## What's your motivation?
The custom recipe, replaces the calls to `java.lang.ref.Reference.clone()` with a constructor call passing in the referent and reference queue as parameters.

@timtebeek 
@cjobinabo 

### Checklist

### DIFF
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -29,6 +29,12 @@ develocity {
             fileFingerprints = true
         }
 
+        publishing {
+            onlyIf {
+                authenticated
+            }
+        }
+
         uploadInBackground = !isCiServer
     }
 }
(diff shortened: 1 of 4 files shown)


timtebeek (approved):
Thanks for the new recipe @AnuRam123 ; good to see you back!

I've made some changes to your recipe to also clear out the fully qualified types, and remove the type casts that are no longer necessary with the new constructors. Hope you agree!

ranuradh:
Thanks @timtebeek yes I was looking into TypeCast and wasn't sure how to go about it. This recipe is now perfect!

PR #518, opened 2024-07-26
fix: UseEnumSetOf should ignore array and varargs parameters

### LINKED ISSUE #516: UseEnumSetOf rewrite breaks for varargs parameter
rewrite-migrate-java 2.20.0

I can reproduce the issue with the following testcase:
```java

  @Test
  void enumSetOfVarargs() {
    rewriteRun(
        spec -> spec.recipe(new UseEnumSetOf()),
        java(
            """
            package com.helloworld;

            import java.util.concurrent.TimeUnit;
            import java.util.Set;

            public class Main {

              public Set<TimeUnit> method(final TimeUnit... units) {
                final Set<TimeUnit> asSet = Set.of(units);
                return asSet;
              }

            }"""));
  }
}

```

Either to bail out and keep it as is. 

Or alternatively we could translate the varargs parameter into a collection first, e.g.
```java
public Set<TimeUnit> method(final TimeUnit... units) {
  final Set<TimeUnit> asSet = EnumSet.noneOf(TimeUnit.class);
  asSet.addAll(Arrays.asList(units));
  return asSet;
}
```
Note that the more convenient-looking `EnumSet.copyOf(asList(units))` throws an exception for empty array.

## What did you see instead?
```java
public Set<TimeUnit> method(final TimeUnit... units) {
  final Set<TimeUnit> asSet = EnumSet.of(units);
  return asSet;
}
```
(doesn't compile)

mbruggmann:
EnumSet doesn't have the equivalent of `Set.of(arr[])`. For now, let's just skip those cases and bail out.

Fixes #516.

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/util/UseEnumSetOf.java
+++ b/src/main/java/org/openrewrite/java/migrate/util/UseEnumSetOf.java
@@ -67,6 +67,10 @@ public class UseEnumSetOf extends Recipe {
                             maybeAddImport("java.util.EnumSet");
 
                             List<Expression> args = m.getArguments();
+                            if (isArrayParameter(args)) {
+                                return m;
+                            }
+
                             StringJoiner setOf = new StringJoiner(", ", "EnumSet.of(", ")");
                             args.forEach(o -> setOf.add("#{any()}"));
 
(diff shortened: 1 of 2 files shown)


timtebeek (approved):
Great to see, thanks a lot! Nice to have this quick turn around on a report & fix.

mbruggmann:
Likewise, thanks @timtebeek!

PR #526, opened 2024-08-07
Recipe to replace PropertyChangeListener methods to ConfigurationListener Methods

BhavanaPidapa:
## What's changed?
image
Replacing the addPropertyListener with addConfigurationListener and removePropertyListener with removeConfigurationListener.

## What's your motivation?
I used org.openrewrite.java.ChangeMethodName for the recipe RemovedLogManagerPropertyChangeListenerMethods.

@timtebeek @cjobinabo 

Since it was not possible to test the recipe. I created sample files, created a local plugin using
**./gradlew publishToMavenLocal** then ran it in the Java8 Sample App using **mvn rewrite:dryRun**

Attaching the rewrite.patch file
[rewrite.patch](https://github.com/user-attachments/files/16532662/rewrite.patch)

### Checklist


timtebeek:
Hi! Thanks for getting this started; I'm looking over the APIs and wondering if this changes is correct & complete.

## Java 8
https://docs.oracle.com/javase/8/docs/api/java/util/logging/LogManager.html#addPropertyChangeListener-java.beans.PropertyChangeListener-
> public void addPropertyChangeListener([PropertyChangeListener](https://docs.oracle.com/javase/8/docs/api/java/beans/PropertyChangeListener.html) l)
                                           throws [SecurityException](https://docs.oracle.com/javase/8/docs/api/java/lang/SecurityException.html)
public void removePropertyChangeListener([PropertyChangeListener](https://docs.oracle.com/javase/8/docs/api/java/beans/PropertyChangeListener.html) l)
                                              throws [SecurityException](https://docs.oracle.com/javase/8/docs/api/java/lang/SecurityException.html)

 ## Java 21
https://docs.oracle.com/en/java/javase/21/docs/api/java.logging/java/util/logging/LogManager.html#addConfigurationListener(java.lang.Runnable)

> public [LogManager](https://docs.oracle.com/en/java/javase/21/docs/api/java.logging/java/util/logging/LogManager.html) addConfigurationListener([Runnable](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/Runnable.html) listener)
public void removeConfigurationListener([Runnable](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/Runnable.html) listener)

Notice how the argument types are different; where one takes an argument that you could write as `event -> { ... }`, whereas the other takes a runnable `() -> { ... }`. The changes made here up to now are just changing the method name, but leaving the arguments as they are, which would then break compilation.

Could you explore what more is needed for a complete migration? We likely can't use the `event` in any capacity inside `{ ... }`, since the replacement method does not have that as input to work with.

timtebeek (commented):
Let's reevaluate the argument handling here, as this likely now breaks compilation.

BhavanaPidapa:
Hey @timtebeek, I will investigate more on this issue and I will get back to you.

BhavanaPidapa:
We don't plan to tackle this using a recipe.

PR #527, opened 2024-08-08
Additional No guava recipes

Laurens-W:
## What's changed?

Replace guava TreeMap with Java TreeMap
Replace guava Primitive asList with Arrays asList

## What's your motivation?

Customer feedback about no guava recipes

Test coverage

@timtebeek 

### Checklist


src/main/java/org/openrewrite/java/migrate/guava/NoGuavaListsNewArrayList.java
```diff
@@ -87,6 +89,11 @@ public J visitMethodInvocation(J.MethodInvocation method, ExecutionContext ctx)
                     maybeAddImport("java.util.ArrayList");
                     return newArrayListCapacity.apply(getCursor(), method.getCoordinates().replace(),
                             method.getArguments().get(0));
+                } else if (NEW_ARRAY_LIST_EXPECTED.matches(method)) {
+                    maybeRemoveImport("com.google.common.collect.Lists");
+                    maybeAddImport("java.util.ArrayList");
+                    return newArrayListCapacity.apply(getCursor(), method.getCoordinates().replace(),
+                            method.getArguments().get(0));
```
timtebeek, line 96: Ah I'd left this out previously, as there's some nuance here that would be lost, since capacity != size. Best to tackle this separately.

src/test/java/org/openrewrite/java/migrate/guava/NoGuavaMapsNewTreeMapTest.java
```diff
@@ -0,0 +1,93 @@
...
+    void replaceWithNewTreeMapWithMap() {
+        //language=java
+        rewriteRun(
+          java(
+            """
+              import com.google.common.collect.*;
+              import java.util.Collections;
+              import java.util.Map;
+              import java.util.SortedMap;
+
```
timtebeek, line 73: IntelliJ has an annoying bug here where the auto format introduces excess white space. I've found it only stops doing so when you completely remove any whitespace on those lines, all the way to the beginning of the line.
```suggestion

```

On my machine I've enabled my IDE to show whitespace, such that these issues and tabs stand out more, I suggest you do the same.
image

And then if you use that `F2` repeately before a commit you'll see prompts like this one to fix it everywhere.
image

Laurens-W: Must have missed it in this case, I have enabled show whitespaces now :)

src/test/java/org/openrewrite/java/migrate/guava/NoGuavaPrimitiveAsListTest.java
```diff
@@ -0,0 +1,243 @@
...
+
+              class Test {
+                  List<Short> shorts = Arrays.asList((short) 1, (short) 2);
+              }
+              """
+          )
+        );
+    }
+
+    @DocumentExample
```
timtebeek, line 217: We typically only have a single test as an example, as we previously showed these in the docs and having too many might clutter the docs; let's remove these from all but one representative example, which is then also the one to copy when folks report an issue.
```suggestion
```

src/main/java/org/openrewrite/java/migrate/guava/NoGuavaPrimitiveAsList.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        JavaVisitor<ExecutionContext> javaVisitor = new AbstractRefasterJavaVisitor() {
+
+            @Override
+            public J visitMethodInvocation(J.MethodInvocation elem, ExecutionContext ctx) {
+                if (METHOD_MATCHER.matches(elem)) {
+                    maybeRemoveImport("com.google.common.primitives.*");
```
timtebeek, line 51: As you'll have found the wildcard doesn't work here; let's remove this line that suggests it might.
```suggestion
```

src/main/java/org/openrewrite/java/migrate/guava/NoGuavaPrimitiveAsList.java
```diff
@@ -0,0 +1,88 @@
...
+                    maybeRemoveImport("com.google.common.primitives.Bytes");
+                    maybeAddImport("java.util.Arrays");
+                    StringBuilder sb = new StringBuilder();
+                    sb.append("Arrays.asList(");
+                    for (int i = 0; i < elem.getArguments().size(); i++) {
+                        sb.append("#{any()}");
+                        if (i != elem.getArguments().size() - 1) {
+                            sb.append(", ");
+                        }
+                    }
```
timtebeek, line 68: This alternative is a little more expressive perhaps.
```suggestion
                    sb.append(String.join(",", Collections.nCopies(elem.getArguments().size(), "#{any()}")));
```

timtebeek: That might then also remove the need for the StringBuilder, as that part can be inlined in `"Arrays.asList(" + joined + ')'`.

src/main/java/org/openrewrite/java/migrate/guava/NoGuavaPrimitiveAsList.java
```diff
@@ -0,0 +1,88 @@
...
+                }
+                return super.visitMethodInvocation(elem, ctx);
+            }
+
+        };
+        return Preconditions.check(
+                Preconditions.and(
+                        new UsesType<>("java.util.List", true),
+                        new UsesMethod<>("com.google.common.primitives.* asList(..)")
+                ),
```
timtebeek, line 84: The second one implies the first right? No need to check both in that case I think.
```suggestion
                new UsesMethod<>("com.google.common.primitives.* asList(..)"),
```

Laurens-W: Sounds fair!

timtebeek (commented):
Good additions, thanks! Couple small pointers but then I think we should be good to merge.

src/test/java/org/openrewrite/java/migrate/guava/NoGuavaPrimitiveAsListTest.java
```diff
@@ -0,0 +1,236 @@
...
+              import java.util.List;
+              
+              class Test {
+                  List<Byte> shorts = Bytes.asList((byte) 1, (byte) 2);
+              }
+              """,
+            """
+              import java.util.Arrays;
+              import java.util.List;
+
```
timtebeek, line 227: Couple more instances here of trailing whitespace that might then trip up the formatter on a next pass through. I might just have to write a recipe to clear these out. 😅 
```suggestion
```

Laurens-W: I think I already resolved those in https://github.com/openrewrite/rewrite-migrate-java/pull/527/commits/0eb99f393d2a8396721a0557dbd4a01ab49f3b19 ?

timtebeek (approved):
Great to see these additional cases covered and added to our NoGuava composite recipe.

PR #528, opened 2024-08-08
Clear out empty newlines

timtebeek:
Reduces friction around code changes due to IntelliJ formatter not handling text blocks well.


src/test/java/org/openrewrite/java/migrate/SunNetSslPackageUnavailableTest.java
```diff
@@ -16,15 +16,12 @@
 package org.openrewrite.java.migrate;
 
 import org.junit.jupiter.api.Test;
-import org.openrewrite.DocumentExample;
```
Laurens-W, line 19: This means there's no happy flow example?

Laurens-W (approved):

src/test/java/org/openrewrite/java/migrate/SunNetSslPackageUnavailableTest.java
timtebeek: Good catch, thanks!

PR #538, opened 2024-08-15
Handle generics in PreferJavaUtilObjectsRequireNonNull

mbruggmann:
## What's changed?
Before this patch, the `PreferJavaUtilObjectsRequireNonNull` recipe would only migrate usage of exactly `checkNotNull(Object)` but not any other (sub)types such as `checkNotNull(String)`, `checkNotNull(CustomType)` etc.

We now match invocations of checkNotNull with exactly one or two arguments which makes the recipe more broadly applicable.

Note that there is also a three argument version that doesn't have a straight-forward replacement, I skipped that for now and added a testcase.

## What's your motivation?
Migrate more usage of Preconditions.checkNotNull throughout our codebase.

The proposed implementation using refaster doesn't deal well with preserving the static import if one was used before. I'm not sure how to best handle that.

@timtebeek since you added refaster support.

I have also tried modifying the method pattern to be `checkNotNull(*)` in no-guava.yml but that did not work as expected. I couldn't find other examples of using a wildcard match on an argument, maybe it isn't supported?

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/guava/NoGuavaRefaster.java
+++ b/src/main/java/org/openrewrite/java/migrate/guava/NoGuavaRefaster.java
@@ -42,7 +42,7 @@ public class NoGuavaRefaster {
     }
 
     @RecipeDescriptor(
-            name = "`Preconditions.checkNotNull` to `Objects.requireNonNull`",
+            name = "`Preconditions.checkNotNull` with message to `Objects.requireNonNull`",
             description = "Migrate from Guava `Preconditions.checkNotNull` to Java 8 `java.util.Objects.requireNonNull`."
     )
     public static class PreconditionsCheckNotNullWithMessageToObjectsRequireNonNull {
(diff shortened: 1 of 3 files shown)


timtebeek:
Thanks for the clear improvement @mbruggmann ! Wasn't aware we only did those very selective replacements before. Great feedback!

We indeed do not yet respect any static import hints on our Refaster reimplementation; that was a conscious choice at the time allowing us to move faster
- https://github.com/openrewrite/rewrite-templating/pull/53

We'd of course hope to circle back in rewrite-templating to pick up and respect those hints, but haven't gotten around to that just yet. Until then we can expect the format here, and perhaps follow up with other recipes like [UseStaticImport](https://docs.openrewrite.org/recipes/java/usestaticimport).

mbruggmann:
Nice, thanks for fixing this up and getting it merged @timtebeek 🙇🏻  
We do indeed have a separate collection of `UseStaticImport` recipes that we apply separately, so at least in our case it doesn't make a difference how the NoGuava recipe behaves in that regard.

PR #541, opened 2024-08-26
Recipes to migrate from Javax, Jakarta and JetBrains nullability annotations to JSpecify

sullis:
## What's changed?

add recipe for jspecify annotations

## What's your motivation?

jspecify 1.0.0 was released in July 2024

https://jspecify.dev/blog/release-1.0.0/

https://www.infoq.com/news/2024/08/jspecify-java-nullability/

n/a

n/a

n/a

## Any additional context

jspecify mailing list 
https://groups.google.com/g/jspecify-discuss/c/h73MkdhAPew

### Checklist

### DIFF
--- a/src/main/resources/META-INF/rewrite/jspecify.yml
+++ b/src/main/resources/META-INF/rewrite/jspecify.yml
@@ -19,7 +19,7 @@ type: specs.openrewrite.org/v1beta/recipe
 name: org.openrewrite.java.jspecify.MigrateToJspecify
 displayName: Migrate to JSpecify
 description: >-
-  This recipe will migrate to JSpecify annotations.
+  This recipe will migrate to JSpecify annotations from various other nullability annotation standards.
 tags:
   - java
 recipeList:
@@ -51,8 +51,8 @@ recipeList:
 ---
 type: specs.openrewrite.org/v1beta/recipe
 name: org.openrewrite.java.jspecify.MigrateFromJakartaAnnotationApi
-displayName: Migrate from jakarta annotation API to JSpecify
-description: Migrate from jakarta annotation API to JSpecify.
+displayName: Migrate from Jakarta annotation API to JSpecify
+description: Migrate from Jakarta annotation API to JSpecify.
 recipeList:
   - org.openrewrite.java.dependencies.AddDependency:
       groupId: org.jspecify
@@ -72,8 +72,8 @@ recipeList:
 ---
 type: specs.openrewrite.org/v1beta/recipe
 name: org.openrewrite.java.jspecify.MigrateFromJetbrainsAnnotations
-displayName: Migrate from jetbrains annotations to JSpecify
-description: Migrate from jetbrains annotations to JSpecify.
+displayName: Migrate from JetBrains annotations to JSpecify
+description: Migrate from JetBrains annotations to JSpecify.
 recipeList:
   - org.openrewrite.java.dependencies.AddDependency:
       groupId: org.jspecify
--- a/src/main/resources/META-INF/rewrite/jspecify.yml
+++ b/src/main/resources/META-INF/rewrite/jspecify.yml
@@ -25,6 +25,7 @@ tags:
 recipeList:
   - org.openrewrite.java.jspecify.MigrateFromJavaxAnnotationApi
   - org.openrewrite.java.jspecify.MigrateFromJakartaAnnotationApi
+  - org.openrewrite.java.jspecify.MigrateFromJetbrainsAnnotations
 
 ---
 type: specs.openrewrite.org/v1beta/recipe
@@ -47,7 +48,6 @@ recipeList:
       onlyIfUsing: org.jspecify.annotations..*
   - org.openrewrite.staticanalysis.java.MoveFieldAnnotationToType:
       annotationType: org.jspecify.annotations.*
-
 ---
 type: specs.openrewrite.org/v1beta/recipe
 name: org.openrewrite.java.jspecify.MigrateFromJakartaAnnotationApi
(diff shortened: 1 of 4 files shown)


src/main/resources/META-INF/rewrite/jspecify.yml
```diff
@@ -0,0 +1,69 @@
...
+      newFullyQualifiedTypeName: org.jspecify.annotations.Nullable
+      ignoreDefinition: true
+  - org.openrewrite.java.ChangeType:
+      oldFullyQualifiedTypeName: javax.annotation.Nonnull
+      newFullyQualifiedTypeName: org.jspecify.annotations.NonNull
+      ignoreDefinition: true
+  - org.openrewrite.java.dependencies.AddDependency:
+      groupId: org.jspecify
+      artifactId: jspecify
+      version: latest.release
```
timtebeek, line 46: Might be best to add an `onlyIfUsing` expression here, such that we only add the dependency where needed.

src/main/resources/META-INF/rewrite/jspecify.yml
```diff
@@ -0,0 +1,69 @@
...
+  - java
+recipeList:
+  - org.openrewrite.java.jspecify.MigrateFromJavaxAnnotationApi
+  - org.openrewrite.java.jspecify.MigrateFromJakartaAnnotationApi
+
+---
+type: specs.openrewrite.org/v1beta/recipe
+name: org.openrewrite.java.jspecify.MigrateFromJavaxAnnotationApi
+displayName: Migrate to JSpecify
+description: Migrate from javax Annotation API to JSpecify.
```
timtebeek, line 33: Probably best to give each recipe a unique name, otherwise the links in the docs would look the same in a list of recipes.

src/main/resources/META-INF/rewrite/jspecify.yml
sullis: I just added `onlyIfUsing` but it doesn't seem to be working.  

I tried this:
```
 onlyIfUsing: org.jspecify.annotations..*
```

I also tried this:
```
 onlyIfUsing: org.jspecify.annotations.*
```

src/main/resources/META-INF/rewrite/jspecify.yml
sullis: Done

wakingrufus:
Nice! Any thoughts on including org.jetbrains.annotations.*?

sullis:
> Nice! Any thoughts on including org.jetbrains.annotations.*?

Great idea.  I am adding `MigrateFromJetbrainsAnnotations`

src/main/resources/META-INF/rewrite/jspecify.yml
```diff
@@ -0,0 +1,92 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+---
+type: specs.openrewrite.org/v1beta/recipe
+name: org.openrewrite.java.jspecify.MigrateToJspecify
+displayName: Migrate to jspecify
```
wmdietl, line 20: It would be nice to use `JSpecify` for consistency, when in user-visible text.

sullis: Fixed.

sullis:
@timtebeek  can you look at the CI build failure?

sullis:
> @timtebeek can you look at the CI build failure?
> 
> ```
> MigrateToJspecifyTest > migrateFromJavaxAnnotationApiToJspecify() FAILED
>     org.opentest4j.AssertionFailedError: [The recipe should have made the following change to "pom.xml"] 
> ```

Perhaps I am using 'onlyIfUsing' incorrectly.   I'm not 100% sure if my implementation is correct.

src/main/resources/META-INF/rewrite/jspecify.yml
timtebeek: There's a bit of an implementation detail here with how that `onlyIfUsing` is implemented, and how scanning recipes work. Scanning recipes run a first separate scan across all source files, before making informed changes downstream. We only do a single scan across all files, as otherwise we could potentially have to do a lot of passes through very large projects, slowing things down considerably. The `AddDependency` recipe is a scanning recipe, that evaluates the `onlyIfUsing` in that initial pass through, seeing the files as they are before the `ChangeType` recipes make their changes. That means that if we evaluate `org.jspecify.annotations.*`, then it's not (yet) seeing those annotations being used, so it's not adding a dependency, even if effectively in the end we do end up using those annotations. The trick then is to look at the "before" situation instead, looking for each of the old annotations, and adding the JSpecify dependency if found, as we expect those annotations to be converted in subsequent steps. Hope that helps explain the issue!

timtebeek:
> > @timtebeek can you look at the CI build failure?
> > ```
> > MigrateToJspecifyTest > migrateFromJavaxAnnotationApiToJspecify() FAILED
> >     org.opentest4j.AssertionFailedError: [The recipe should have made the following change to "pom.xml"] 
> > ```
> 
> Perhaps I am using 'onlyIfUsing' incorrectly. I'm not 100% sure if my implementation is correct.

All good; took me a little while longer to get to this than I had hoped, but I've fixed the issue just now with a write up in https://github.com/openrewrite/rewrite-migrate-java/pull/541#discussion_r1738281451

timtebeek (approved):
Thanks a lot for adding these recipes @sullis ! Should make it a lot easier for folks to make the switch as well. :)

PR #545, opened 2024-09-06
Modified the RemovedSubjectMethods recipe

BhavanaPidapa:
## What's changed?
The recipe should change doAs() method with callAs() method. The existing recipe was replacing the wrong Value. Now, updated the recipe with correct values. 

## What's your motivation?
I used org.openrewrite.java.ChangeMethodName for the recipe RemovedSubjectMethods.

@timtebeek @cjobinabo 

### Checklist


timtebeek:
This makes sense looking at https://docs.oracle.com/javase/8/docs/api/javax/security/auth/Subject.html 
Thanks for the quick fix!

PR #549, opened 2024-09-11
Correctly override parent property to set Java version

### LINKED ISSUE #523: `maven.compiler.release` set in parent is not recognized in child module
`maven.compiler.release` which is defined in parent is not recognized in child module and patch for explicit overriding property is generated by `UpgradeToJava17` recipe.

I am using
- rewrite-migrate-java 2.21.0
- Maven plugin 5.37.1

I am using the Maven plugin, and my project is a multi-module project.

https://github.com/pzrep/openrewrite-migrate-compiler-release-in-child/

No patch generated by recipe as property inherited from parent is good enough.

## What did you see instead?
Patch generated by recipe (step: Show applied changes): https://github.com/pzrep/openrewrite-migrate-compiler-release-in-child/actions/runs/10196871940/job/28208511361?pr=2

```diff
diff --git a/child/pom.xml b/child/pom.xml
index 96186f3..2570020 100644
--- a/child/pom.xml
+++ b/child/pom.xml
@@ -8,5 +8,8 @@
         <version>1-SNAPSHOT</version>
     </parent>
     <artifactId>openrewrite-migrate-compiler-release-in-child.child</artifactId>
+    <properties>
+        <maven.compiler.release>17</maven.compiler.release>
+    </properties>
```

https://github.com/openrewrite/rewrite-migrate-java/commit/8d15935263b802d709861c0509c17dfbfae8109b#commitcomment-144892557

DidierLoiseau:
## What's changed?
* Added a test case on `UpdateMavenProjectPropertyJavaVersionTest` expecting no change when upgrading to Java 17 but parent is already spring-boot 3.3 (declares `java.version` and `maven.compiler.release` with value 17 already) .
* Partially reverted 12c8f372d77f2b4612e18d1e420816612106fdc8
* Implemented a fix to only override properties that actually parse to a number, and to add `maven.release.plugin` only when no release property is found

## What's your motivation?
- Reproduce & fix #523

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/maven/UpdateMavenProjectPropertyJavaVersion.java
+++ b/src/main/java/org/openrewrite/java/migrate/maven/UpdateMavenProjectPropertyJavaVersion.java
@@ -79,7 +79,6 @@ public class UpdateMavenProjectPropertyJavaVersion extends Recipe {
     @Override
     public TreeVisitor<?, ExecutionContext> getVisitor() {
         return new MavenIsoVisitor<ExecutionContext>() {
-            final Set<String> propertiesExplicitlyReferenced = new HashSet<>();
             boolean compilerPluginConfiguredExplicitly;
 
             @Override
@@ -99,7 +98,7 @@ public class UpdateMavenProjectPropertyJavaVersion extends Recipe {
                 MavenResolutionResult mrr = getResolutionResult();
                 Map<String, String> currentProperties = mrr.getPom().getRequested().getProperties();
                 for (String property : JAVA_VERSION_PROPERTIES) {
-                    if (currentProperties.containsKey(property) || !propertiesExplicitlyReferenced.contains(property)) {
+                    if (currentProperties.containsKey(property)) {
                         continue;
                     }
                     d = (Xml.Document) new AddProperty(property, String.valueOf(version), null, false)
@@ -130,12 +129,8 @@ public class UpdateMavenProjectPropertyJavaVersion extends Recipe {
             @Override
             public Xml.Tag visitTag(Xml.Tag tag, ExecutionContext ctx) {
                 Xml.Tag t = super.visitTag(tag, ctx);
-                Optional<String> s = t.getValue()
-                        .map(it -> it.replace("${", "").replace("}", "").trim())
-                        .filter(JAVA_VERSION_PROPERTIES::contains);
-                if (s.isPresent()) {
-                    propertiesExplicitlyReferenced.add(s.get());
-                } else if (JAVA_VERSION_XPATH_MATCHERS.stream().anyMatch(matcher -> matcher.matches(getCursor()))) {
+
+                if (JAVA_VERSION_XPATH_MATCHERS.stream().anyMatch(matcher -> matcher.matches(getCursor()))) {
                     Optional<Float> maybeVersion = t.getValue().flatMap(
                             value -> {
                                 try {
--- a/src/main/java/org/openrewrite/java/migrate/maven/UpdateMavenProjectPropertyJavaVersion.java
+++ b/src/main/java/org/openrewrite/java/migrate/maven/UpdateMavenProjectPropertyJavaVersion.java
@@ -23,14 +23,12 @@ import org.openrewrite.Recipe;
 import org.openrewrite.TreeVisitor;
 import org.openrewrite.maven.AddProperty;
 import org.openrewrite.maven.MavenIsoVisitor;
-import org.openrewrite.maven.tree.MavenResolutionResult;
 import org.openrewrite.xml.XPathMatcher;
 import org.openrewrite.xml.tree.Xml;
 
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
-import java.util.Optional;
 import java.util.stream.Collectors;
 
 @Value
--- a/src/main/java/org/openrewrite/java/migrate/maven/UpdateMavenProjectPropertyJavaVersion.java
+++ b/src/main/java/org/openrewrite/java/migrate/maven/UpdateMavenProjectPropertyJavaVersion.java
@@ -90,22 +90,22 @@ public class UpdateMavenProjectPropertyJavaVersion extends Recipe {
                 Xml.Document d = super.visitDocument(document, ctx);
 
                 // Return early if the parent appears to be within the current repository, as properties defined there will be updated
-                Optional<String> pathToLocalParent = d.getRoot().getChild("parent")
+                if (d.getRoot().getChild("parent")
                         .flatMap(parent -> parent.getChild("relativePath"))
-                        .flatMap(Xml.Tag::getValue);
-                if (pathToLocalParent.isPresent()) {
+                        .flatMap(Xml.Tag::getValue)
+                        .isPresent()) {
                     return d;
                 }
 
                 // Otherwise override remote parent's properties locally
-                MavenResolutionResult mrr = getResolutionResult();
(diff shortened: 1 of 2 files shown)


timtebeek:
Much appreciated to have this runnable @DidierLoiseau ! Did you already look into a potential fix following [the commit you referenced](https://github.com/openrewrite/rewrite-migrate-java/commit/8d15935263b802d709861c0509c17dfbfae8109b)?

I'm a little behind on reviews, but will try to get to this one as well if you can't work out a fix already, as I understand the frustration of seeing a needless `maven.compiler.release` introduced.

DidierLoiseau:
> Much appreciated to have this runnable @DidierLoiseau ! Did you already look into a potential fix following [the commit you referenced](https://github.com/openrewrite/rewrite-migrate-java/commit/8d15935263b802d709861c0509c17dfbfae8109b)?

No, I’m new to OpenRewrite. I just quickly glanced over the commits of v21.0 and noticed one was a big refactoring/rewrite of `UpgradeJavaVersion`.

There is a side-question to this which is: which property should actually be set by `UpgradeJavaVersion` when an override of the parent is actually necessary? Because when the parent has specific/custom properties handling this (such as Spring’s `java.version`), it would be better to override that property than to set `maven.compiler.release`. For instance in Spring, this affects the configuration of `kotlin-maven-plugin`.

DidierLoiseau:
> Much appreciated to have this runnable @DidierLoiseau ! Did you already look into a potential fix following [the commit you referenced](https://github.com/openrewrite/rewrite-migrate-java/commit/8d15935263b802d709861c0509c17dfbfae8109b)?

In the end it seems to be commit 12c8f372 combined with that commit which caused the issue: the former was only taking into account properties **used** in the current pom, while the latter was adding `maven.compiler.release` when no property was updated.

I don’t really understand the purpose of 12c8f372: if the parent defines a property, it’s certainly to be used _by the parent_ (like in Spring Boot) so we have to update it. I doubt there is a use case where you wouldn’t want that, but that commit didn’t reference a specific issue number.

I thus implemented a fix that reverted it – all properties from the (remote) parent are now overridden when they have a numeric value. With Spring Boot for example, this means that only `java.version` will be overridden (if needed) but not `maven.compiler.release`.

If this is not satisfying, I think the only solution would be to parse the `maven.compiler.*` properties + the `maven-compiler-plugin` configuration to determine which variable actually defines the version (at least maintaining `JAVA_VERSION_PROPERTIES` wouldn’t be needed any more…).

There are still 2 limitations (already there before):
* If the child overrides a parent property, it does not try to check if the parent already has the desired value to remove it. For instance, let’s say someone working with Spring Boot 2 first upgraded to Java 11 by setting `java.version` to 11. Now they want to upgrade to SB 3 and thus Java 17. The recipe will not detect that the SB3 parent already sets `java.version` to 17, it will thus just update it in the child. I think it’s easy to work around this with `RemoveProperty`.
* If the child overrides a local parent property, no update is performed, so it could stay in an old version defined in the child.

src/main/java/org/openrewrite/java/migrate/maven/UpdateMavenProjectPropertyJavaVersion.java
```diff
@@ -65,21 +65,20 @@ public String getDisplayName() {
...
-               "If none of these properties are in use and the maven compiler plugin is not otherwise configured adds the `maven.compiler.release` property.";
+                " * `java.version`\n" +
+                " * `jdk.version`\n" +
+                " * `javaVersion`\n" +
+                " * `jdkVersion`\n" +
+                " * `maven.compiler.source`\n" +
+                " * `maven.compiler.target`\n" +
+                " * `maven.compiler.release`\n" +
+                " * `release.version`\n\n" +
+                "If none of these properties are in use and the maven compiler plugin is not otherwise configured, adds the `maven.compiler.release` property.";
```
DidierLoiseau, line 76: IntelliJ still gives me slightly different formatting here, I don’t know why…

src/main/java/org/openrewrite/java/migrate/maven/UpdateMavenProjectPropertyJavaVersion.java
```diff
@@ -97,28 +96,30 @@ public Xml.Document visitDocument(Xml.Document document, ExecutionContext ctx) {
...
-                        .stream()
-                        .noneMatch(JAVA_VERSION_PROPERTIES::contains)) {
+                if (!foundProperty && version >= 9 && !compilerPluginConfiguredExplicitly) {
                     d = (Xml.Document) new AddProperty("maven.compiler.release", String.valueOf(version), null, false)
                             .getVisitor()
                             .visitNonNull(d, ctx);
-                    HashMap<String, String> updatedProps = new HashMap<>(currentProperties);
+                    HashMap<String, String> updatedProps = new HashMap<>(mrr.getPom().getRequested().getProperties());
                     updatedProps.put("maven.compiler.release", version.toString());
                     mrr = mrr.withPom(mrr.getPom().withRequested(mrr.getPom().getRequested().withProperties(updatedProps)));
```
DidierLoiseau, line 125: I don’t understand why this is needed, It’s not done for the other `AddProperty` executions.

DidierLoiseau:
I think maybe @sambsnyd should have a look since he is the one who implemented 12c8f372d77f2b.

timtebeek:
Thanks a lot for the work already done here @DidierLoiseau ! I've tried for a bit to get the new failures in `org.openrewrite.java.migrate.UpgradeToJava17Test` going, but can't immediately work it out. They fail with

> java.lang.AssertionError: Expected recipe to complete in 1 cycle, but took 2 cycles. This usually indicates the recipe is making changes after it should have stabilized.

We usually see that if a recipe unconditionally makes a change, as opposed to only when the existing value does not match the target value.

Could be good to have @sambsnyd weigh in indeed, as I've seen reports of these excess properties over the past week or so.

DidierLoiseau:
@timtebeek I fixed the test. The problem was related to the second comment I made: previously, the code was working with `mrr.getPom().getRequested().getProperties()`, so updating that was enough to prevent a second cycle of the recipe. Now that it works with `mrr.getPom().getProperties()`, that’s what needs to be updated.

It does not look like there is a convenient way to do that, so I just used `maybeUpdateModel()` in boh cases.

src/main/java/org/openrewrite/java/migrate/maven/UpdateMavenProjectPropertyJavaVersion.java
DidierLoiseau: This was to prevent a second cycle of the loop: after adding the property to the `mrr`, it would prevent entering this `if` again.

DidierLoiseau:
Actually, there seems to be a bug in `AddPropertyVisitor.visitTag()` as it does not call `maybeUpdateModel()` when updating a property. It’s only done when adding one.

Previously it was working because `UpdateMavenProjectPropertyJavaVersion` was never using `AddProperty` for updating properties… except in the `maven.compiler.release` case, where it was changing the `mrr.getPom().getRequested().getProperties()` directly to prevent entering that block again.

Side note: when visitors instantiate and call other visitors which call `maybeUpdateModel()` (like it should have been the case here, actually), I think the `UpdateMavenModel` visitor will be added to each individual visitor’s `afterVisit`, so it will be executed several times.

timtebeek (approved):
I like the simplifications done here with increased focus, and pointing to `UseMavenCompilerPluginReleaseConfiguration` where a change would be out of scope for this particular recipe. The added tests for common cases in practice will help guard regression in the future. Thanks a lo!

Issue #551, opened 2024-09-12
`UseVarForObject` creates broken code when replacing array that is created with array literals


uhafner:
## What version of OpenRewrite are you using?

I am using

- <rewrite-maven-plugin.version>5.39.2</rewrite-maven-plugin.version>
- <rewrite-testing-frameworks.version>2.17.1</rewrite-testing-frameworks.version>
- <rewrite-static-analysis.version>1.15.0</rewrite-static-analysis.version>
- <rewrite-migrate-java.version>2.23.0</rewrite-migrate-java.version>
- <rewrite-recommendations.version>1.8.4</rewrite-recommendations.version>

## How are you running OpenRewrite?

I am using the Maven plugin, and my project is a single module project. I did run the command `mvn -X clean rewrite:run`.

See https://github.com/uhafner/codingstyle for details. The OpenRewrite configuration is visible in the [pom.xml](https://github.com/uhafner/codingstyle/blob/8826fcc351959f67906fe42bfb9e75f089d77a0f/pom.xml#L275-#L322).

## What is the smallest, simplest way to reproduce the problem?

```java
package edu.hm.hafner.util;

class TreeStringBuilderTest {
    void error() {
        String[] dictionary = {"aa", "b", "aba", "ba"};
    }
}

```

## What did you expect to see?

Type variable should not be replaced by var or the array literals initializer needs to be replaced with a `new` initializer.

```java
package edu.hm.hafner.util;

class TreeStringBuilderTest {
    void error() {
        var dictionary = new String[] {"aa", "b", "aba", "ba"};
    }
}

```
## What did you see instead?

```java
package edu.hm.hafner.util;

class TreeStringBuilderTest {
    void error() {
        var dictionary = {"aa", "b", "aba", "ba"};
    }
}

```
## What is the full stack trace of any errors you encountered?

No stack trace produced.

## Are you interested in [contributing a fix to OpenRewrite]

Not right now.

MBoegers:
Thanks for the report, I'll have a look into this.

MBoegers:
Recalled [JEP 286](https://openjdk.org/jeps/286) about local variable type inference, there is a section _Risks and Assumptions_ that explicitly states `var x = {1, 2}` as not inferable and therefore illegal. 

Options here are:
- Skipping array initializer in general, so `int[] x = {1, 2}` stays
[...]

PR #552, opened 2024-09-12
Skip adding var to local variables initialized by array initializer

MBoegers:
## What's changed?
`String[] dictionary = {"aa", "b", "aba", "ba"};` will no longer be transformed to `var dictionary = {"aa", "b", "aba", "ba"};`

## What's your motivation?
- Fixes https://github.com/openrewrite/rewrite-migrate-java/issues/551

Move type attribution like: `int[] x = {1, 2}` to `var x = new int[]{1, 2}` but that's way more complicated and adds nothing special. 

### Checklist


timtebeek (approved):
Thanks for the quick fix! Hope you're able to get some sleep. :)

Issue #453, opened 2024-04-04
Minor issue with StringFormatted recipe


motlin:
## What version of OpenRewrite are you using?

I am using

- Maven/Gradle plugin v5.26.0
- rewrite-migrate-java v2.11.0

## How are you running OpenRewrite?

I am using the Maven plugin, and [my project](https://github.com/motlin/liftwizard) is a multi module project.

## What is the smallest, simplest way to reproduce the problem?

Run org.openrewrite.java.migrate.lang.StringFormatted on a String.format() call that has a newline after the open paren:

```java
String error = String.format(
        "Overwrote system property {%s:%s} with %s.",
        key,
        oldValue,
        value);
```

## What did you expect to see?

Rewritten code like this:

```java
String error =
        "Overwrote system property {%s:%s} with %s.".formatted(
        key,
        oldValue,
        value);
```

Or like this:

```java
String error = "Overwrote system property {%s:%s} with %s.".formatted(
        key,
        oldValue,
        value);
```

## What did you see instead?

Code like the first example above, but there's a trailing space character after the equals sign.

```java
String error = // <-- extra, trailing space here
        "Overwrote system property {%s:%s} with %s.".formatted(
        key,
        oldValue,
        value);
```

This is a minor inconvenience, but Checkstyle immediately fails on the rewritten code.

timtebeek:
Thanks for the clear and detailed report @motlin ; Such a space is stored in a prefix of the next LST element, and it appears we have no explicit handling of the prefix in the current implementation.

https://github.com/openrewrite/rewrite-migrate-java/blob/cca0195e404553416301aa60064374266002a223/src/main/java/org/openrewrite/java/migrate/lang/StringFormatted.java#L77-L84

Typically you'd see a `.withPrefix` call that could remove any prefix if no longer needed.

If you want to help resolve this issue the best place to start is with a reproduction test similar to this one, and from there make the required change:
https://github.com/openrewrite/rewrite-migrate-java/blob/cca0195e404553416301aa60064374266002a223/src/test/java/org/openrewrite/java/migrate/lang/StringFormattedTest.java#L35-L64

Hope that helps!

bsmahi:
Hi @timtebeek have tried with the `mi = mi.withPrefix(Space.EMPTY);` written test case for the above scenarios, it is working fine however, we need to fix remaining text cases as well. 

```java
Expression select = wrapperNotNeeded ? arguments.get(0) :
[...]

PR #556, opened 2024-09-15
StringFormatted should not leave trailing whitespace after replacement

BramliAK:
## What's changed?
I updated the StringFormattedVisitor class to remove return line for the first argument

## What's your motivation?
Integrating openRewitre team

- #453

### Checklist


timtebeek:
Thanks for the fix here @BramliAK ! I've pushed up a few quick polishing commits; good to see this resolved!

PR #558, opened 2024-09-18
Drop unnecessary trailing escape in UseTextBlocks

BramliAK:
## What's changed?
i added some control to not create unnecessary trailing escape in some cases on the other hand I leave them in the case where the text ends with Quotation Mark, Equal, Curly Brackets and Semicolon because I saw that it is more readable like that

### Checklist


timtebeek:
Really appreciate the measured change here @BramliAK !

timtebeek (approved):

PR #563, opened 2024-09-25
Revert adding SerialVersionUID

melloware:
@DidierLoiseau @timtebeek has convinced me this is bad change and leave only the `@Serial` in

### DIFF
--- a/src/main/resources/META-INF/rewrite/java-version-17.yml
+++ b/src/main/resources/META-INF/rewrite/java-version-17.yml
@@ -30,7 +30,6 @@ recipeList:
   - org.openrewrite.java.migrate.UpgradeBuildToJava17
   - org.openrewrite.java.migrate.lang.StringFormatted
   - org.openrewrite.staticanalysis.InstanceOfPatternMatch
-  - org.openrewrite.staticanalysis.AddSerialVersionUidToSerializable
   - org.openrewrite.staticanalysis.AddSerialAnnotationToSerialVersionUID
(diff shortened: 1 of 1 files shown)


DidierLoiseau:
For reference, this follows a discussion in #560. Thanks @melloware for you reactivity.

@timtebeek, I don’t think it’s possible to downgrade `rewrite-migrate-java` when running the Maven plugin from the command-line without declaring the plugin in the pom, right? Would it be possible to make a new release of the plugin (with the updated bom) once this is fixed?

timtebeek:
Thanks both! It's definitely possible to override just one recipe module. I'll push a patch release of this one

DidierLoiseau:
@timtebeek 

> It's definitely possible to override just one recipe module.

without configuring the plugin in the pom?

timtebeek:
Absolutely! You can do so by passing in `-Drewrite.recipeArtifactCoordinates=org.openrewrite.recipe:rewrite-migrate-java:2.26.1`
https://openrewrite.github.io/rewrite-maven-plugin/run-mojo.html#recipeartifactcoordinates

timtebeek:
Hmm that's unfortunate; not quite sure how Maven merges conflicting plugin dependencies; perhaps there's ordering there too? So would `-Drewrite.recipeArtifactCoordinates=org.openrewrite.recipe:rewrite-migrate-java:2.26.1,you.library:second:1.0.1` work?

PR #570, opened 2024-10-05
`UseEnumSetOf` should rewrite `EnumSet.noneOf` from empty `Set.of()`

BramliAK:
## What's changed?
Rewriting with EnumSet.noneOf from empty Set.of() by UseEnumSetOf

### Checklist


timtebeek (approved):
Thanks for taking this on @BramliAK ! I've pushed up a small polishing commit to clarify that the `noneOf` and `of` cases are fully separate, reusing the components you already deduced we need for the new separate template. Great to see this additional case covered with minimal changes

PR #575, opened 2024-10-10
`Preconditions.checkNotNull` recipes should retain unrelated `String.valueOf(String)`

amishra-u:
## What's changed?
String.valueOf(String) and the String itself are not equivalent, especially when the value of the String is null. When the String is null, String.valueOf(String) returns the literal "null", whereas directly referencing the String would result in a null value.
Also the intended purpose of recipe (String.ofValue) was to only update the expressions which was previously updated by "PreconditionsCheckNotNullWithMessageToObjectsRequireNonNull", in such case adding a marker to the expression and then selectively updating the marked expression is desired but it seems difficult to do with refasterstyle templates.

So deleted the refaster template for this conversion.

### Checklist


timtebeek (approved):
Great to see this improve the specificity of the recipes. Thanks a lot @amishra-u

PR #577, opened 2024-10-11
Johnzon upgrade for Jakarta 9

BoykoAlex:
Current implementation moves up the version of all `org.apachae.johnzon` artifacts as well as adds `jakarta` classifier to each artifact.
I saw https://johnzon.apache.org/download.html and figured that this is exactly what needs to be done. However, after running the Boot 3.x upgrade recipe (part of which is Johnzon lib migration) I saw errors in the POM that the dependency with the classifier cannot be found (at least on the Maven Central perhaps?).

I looked at https://johnzon.apache.org/ and having dependencies without the classifier and adding:
```xml
    <dependency>
      <groupId>jakarta.json</groupId>
      <artifactId>jakarta.json-api</artifactId>
      <version>2.1.2</version>
      <scope>provided</scope> <!-- or compile if your environment doesn't provide it -->
    </dependency>
```
as the docs suggested makes everything compile.


timtebeek (approved):
Thanks for the fix!

PR #580, opened 2024-10-14
Temporarily remove Recipes Related to ChangeType until #4582 fixed

amishra-u:
## What's changed?
Removed three recipes based on ChangeType Recipe for nogauva recipe list as they incorrectly updates the type. more detail https://github.com/openrewrite/rewrite/issues/4582

## What's your motivation?
Make noguava recipe stable.

### Checklist


amishra-u:
CI Tests are failing because of this change
https://github.com/openrewrite/rewrite/commit/f50d4aee6586807d4aa5393e49906b0bb61cdc3c

which is fixed later in 
https://github.com/openrewrite/rewrite/commit/f9b638d085c8f9a2400602c2de9d115378dff183#diff-896f2e51b7ded561add43c5e439ad0bdc7384b26a6c47f4d9beee459eb21535eR149

knutwannheden:
The NoGuava recipe is indeed far from complete and it probably never will be complete. Some features just don't have any counterparts in the JDK.

I think if you want to exclude these three recipes, you can just create your own composite YAML recipe without them in it.

PR #587, opened 2024-10-20
Recipes to replace Guava `Immutable{Set|List|Map}.copyOf()` with Java `{Set|List|Map}.copyOf()`

### LINKED ISSUE #584: Recipes to replace Guava `Immutable{Set|List|Map}.copyOf()` calls with Java `{Set|List|Map}.copyOf()` calls
Since Java 10 the interfaces `Set`, `List`, and `Map` declare static `copyOf()` methods which are equivalent to the `copyOf()` counterparts in Guava's `ImmutableSet`, `ImmutableList`, and `ImmutableMap` counterparts.

Implement three imperative recipes which implement this migration. It would probably make sense to create an abstract supertype based on `AbstractNoGuavaImmutableOf`, so that the recipes can be declared as subtypes of that new abstract base class. As noted the precondition would be Java 10.

There should be some test cases for this (again similarly to as how it was done for the implementations of `AbstractNoGuavaImmutableOf`) and finally the three new recipes should be added to the recipe list of `org.openrewrite.java.migrate.guava.NoGuavaJava11` which is declared in `no-guava.yml`, where also the other three mentioned recipes are already declared.

knutwannheden:
- Fixes #584

Add recipes to replace Guava `ImmutableSet.copyOf()`, `ImmutableList.copyOf()`, and `ImmutableMap.copyOf()` calls with Java `Set.copyOf()`, `List.copyOf()`, and `Map.copyOf()` calls.

* Add `AbstractNoGuavaImmutableCopyOf` class to provide a base for the new recipes.
* Add `NoGuavaImmutableSetCopyOf`, `NoGuavaImmutableListCopyOf`, and `NoGuavaImmutableMapCopyOf` classes extending `AbstractNoGuavaImmutableCopyOf`.
* Update `no-guava.yml` to include the new recipes.
* Add test classes `NoGuavaImmutableSetCopyOfTest`, `NoGuavaImmutableListCopyOfTest`, and `NoGuavaImmutableMapCopyOfTest` with test cases for the new recipes.

---

For more details, open the [Copilot Workspace session](https://copilot-workspace.githubnext.com/openrewrite/rewrite-migrate-java/issues/584?shareId=7dd268d3-5b32-42c5-88cb-8d8345acccaa).

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "java.compile.nullAnalysis.mode": "automatic"
+}
--- a/src/main/java/org/openrewrite/java/migrate/guava/AbstractNoGuavaImmutableCopyOf.java
+++ b/src/main/java/org/openrewrite/java/migrate/guava/AbstractNoGuavaImmutableCopyOf.java
@@ -1,0 +1,144 @@
+/*
+ * Copyright 2023 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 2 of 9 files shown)


BramliAK:
In my opinion, a new abstract class known as AbstractNoGuavaImmutable can be created and the common methods can be added to it. We can further expand it with AbstractNoGuavaImmutableCopyOf.java and AbstractNoGuavaImmutableOf.java to eliminate the use of duplicated code.

timtebeek:
I was thinking along the same lines, especially after the improvements you added @BramliAK ; reuse between these two is likely preferred over duplication.

PR #594, opened 2024-10-28
Joda time to Java time: Add few more templates for migration

amishra-u:
## What's changed?
Added migration templates for few more joda-time methods,  including methods from the following classes:
1. AbstractDateTime
2. AbstractDuration
3. AbstractInstant
4. BaseDateTime
5. DateTimeFormatter
6. Instant

Additionally, introduced support for migrating methods implemented by two different classes that do not share the same class hierarchy.

**Note:** I had to increase heap size for test task. The tests were failing with OutOfMemory error.

### Checklist


build.gradle.kts
```diff
@@ -65,3 +65,7 @@ dependencies {
 tasks.withType(Javadoc::class.java) {
     exclude("**/PlanJavaMigration.java")
 }
+
+tasks.test {
+    maxHeapSize = "2g"  // Set max heap size to 2GB or adjust as necessary
```
timtebeek, line 70: Surprised we're having to set this here and not generally in for instance
https://github.com/openrewrite/rewrite-build-gradle-plugin/blob/ce39a207adb0bb0c657517f9eb32853b83152e18/src/main/java/org/openrewrite/gradle/RewriteJavaPlugin.java#L102-L131

sambsnyd: Yeah we should probably set a higher memory limit than the default 512MB there. I am a bit surprised that this project needs more than that, though. The unit tests shouldn't be particularly demanding on memory, but maybe something has increased our consumption from the last time I looked

amishra-u: I believe applyTemplate operation is very heavy on memory. And this JodaTime has too many apply  template calls. Everytime it OOMed on applyTemplate call.

PR #601, opened 2024-11-13
Do not convert return types in NoGuavaJava21 by default

BramliAK:
## What's changed?
Rollback  visit Variable Declarations

### Checklist


timtebeek (approved):
Thanks @BramliAK ! Good to be back to stable recipe outputs by default, while still being able to do the unsafe migrations if desired.

PR #603, opened 2024-11-13
Proof that `javax.jms` is migrated to `jakarta.jms` in spring.xml

Laurens-W:
## What's changed?

Added a testcase to prove `javax` is migrated to `jakarta` in a spring.xml beans config file

@timtebeek @knutwannheden 

## Any additional context

Uses the `Reference` trait
- https://github.com/openrewrite/rewrite/pull/4648

### Checklist

### DIFF
--- a/src/test/java/org/openrewrite/java/migrate/jakarta/JavaxToJakartaTest.java
+++ b/src/test/java/org/openrewrite/java/migrate/jakarta/JavaxToJakartaTest.java
@@ -25,6 +25,7 @@ import org.openrewrite.test.RewriteTest;
 
 import static org.openrewrite.java.Assertions.*;
 import static org.openrewrite.maven.Assertions.pomXml;
+import static org.openrewrite.xml.Assertions.xml;
 
 class JavaxToJakartaTest implements RewriteTest {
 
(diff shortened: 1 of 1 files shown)


src/test/java/org/openrewrite/java/migrate/jakarta/JavaxToJakartaTest.java
```diff
@@ -610,4 +611,36 @@ public class TestApplication {
           )
         );
     }
+
+    @Test
+    void shouldRefactorSpringBeanXml() {
+        //language=java
```
timtebeek, line 617:
```suggestion
```

timtebeek (approved):

timtebeek:
Any idea what's causing the failure here? I see 
https://github.com/openrewrite/rewrite-migrate-java/blob/20b8fffb0ea48ba279d12cca2e1d3a3d23d30419/src/main/java/org/openrewrite/java/migrate/javax/UseJoinColumnForMapping.java#L73

So perhaps then another case of a fix through:
- https://github.com/openrewrite/rewrite/pull/4670 ?

PR #604, opened 2024-11-15
Add a `ScopedVariable` trait

Laurens-W:
## What's changed?

Added a trait to pull together the scope and identifier of `NamedVariable`s and added testcases to showcase its usage

@sambsnyd @timtebeek 

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/trait/ScopedVariable.java
+++ b/src/main/java/org/openrewrite/java/migrate/trait/ScopedVariable.java
@@ -1,0 +1,29 @@
+package org.openrewrite.java.migrate.trait;
+
+import lombok.RequiredArgsConstructor;
+import lombok.Value;
+import org.jspecify.annotations.Nullable;
+import org.openrewrite.Cursor;
+import org.openrewrite.java.tree.J;
+import org.openrewrite.trait.SimpleTraitMatcher;
+import org.openrewrite.trait.Trait;
+
+@Value
+public class ScopedVariable implements Trait<J.VariableDeclarations.NamedVariable> {
+    Cursor cursor;
+    Cursor scope;
+    J.Identifier identifier;
+
+    @RequiredArgsConstructor
+    public static class Matcher extends SimpleTraitMatcher<ScopedVariable> {
+
+        @Override
+        protected @Nullable ScopedVariable test(Cursor cursor) {
+            if (cursor.getValue() instanceof J.VariableDeclarations.NamedVariable) {
+                J.VariableDeclarations.NamedVariable variable = cursor.getValue();
+                return new ScopedVariable(cursor, variable.getDeclaringScope(cursor), variable.getName());
+            }
+            return null;
+        }
+    }
+}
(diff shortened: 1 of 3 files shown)


src/main/java/org/openrewrite/java/migrate/trait/ScopedVariable.java
```diff
@@ -0,0 +1,44 @@
...
+import lombok.RequiredArgsConstructor;
+import lombok.Value;
+import org.jspecify.annotations.Nullable;
+import org.openrewrite.Cursor;
+import org.openrewrite.java.tree.J;
+import org.openrewrite.trait.SimpleTraitMatcher;
+import org.openrewrite.trait.Trait;
+
+@Value
+public class ScopedVariable implements Trait<J.VariableDeclarations.NamedVariable> {
```
knutwannheden, line 27: Shouldn't we instead create this trait in the rewrite repo?

src/main/java/org/openrewrite/java/migrate/trait/ScopedVariable.java
```diff
@@ -0,0 +1,44 @@
...
+    J.Identifier identifier;
+
+    @RequiredArgsConstructor
+    public static class Matcher extends SimpleTraitMatcher<ScopedVariable> {
+
+        @Override
+        protected @Nullable ScopedVariable test(Cursor cursor) {
+            if (cursor.getValue() instanceof J.VariableDeclarations.NamedVariable) {
+                J.VariableDeclarations.NamedVariable variable = cursor.getValue();
+                return new ScopedVariable(cursor, variable.getDeclaringScope(cursor), variable.getName());
```
knutwannheden, line 39: This `getDeclaringScope()` utility method is probably a good starting point. But to me it looks like it misses cases like try-with-resources and exception handlers, for example. Also note that the flow scoping of instanceof patterns is quite different and probably a bad fit. Maybe worth thinking about at least.

src/main/java/org/openrewrite/java/migrate/trait/ScopedVariable.java
```diff
@@ -0,0 +1,44 @@
...
+    Cursor cursor;
+    Cursor scope;
+    J.Identifier identifier;
+
+    @RequiredArgsConstructor
+    public static class Matcher extends SimpleTraitMatcher<ScopedVariable> {
+
+        @Override
+        protected @Nullable ScopedVariable test(Cursor cursor) {
+            if (cursor.getValue() instanceof J.VariableDeclarations.NamedVariable) {
```
amishra-u, line 37: We can likely just use variable.getDeclaringScope(cursor) to obtain the variable's scope; using a trait might be overkill.

A scenario where a trait could be useful, suppose I'm at an identifier and need to determine the declaring scope of the variable that this identifier references.

```
void foo() {
   int x = 5;
   if (true) {
      print(x) // find the declaring scope of x at this point,
   }
}
```

Laurens-W:
Continued here
- https://github.com/openrewrite/rewrite/pull/4682

PR #605, opened 2024-11-16
Joda-Time to Java time: Add support for Method Parameter Migration

amishra-u:
## What's changed?
- Added support for safe migration of method parameters.
- A method parameter is considered safe for migration if:
   - All expressions referencing the parameter within the method body can be migrated.
   - All arguments assigned to this parameter in method calls can be migrated.

### Key Changes:
1. Used data flow analysis to identify the referencing expressions for method parameter.
2. Added VarTable to track all method parameters across files.
3. Added new test cases and updated existing ones to reflect the changes.

### Not Implemented Yet:
1. Migration of method return statements.
2. Migration of class variables.

### Checklist


src/main/java/org/openrewrite/java/migrate/joda/JodaTimeRecipe.java
```diff
@@ -15,17 +15,19 @@
...
-import java.util.HashSet;
-import java.util.Set;
+import java.util.*;
 
-public class JodaTimeRecipe extends ScanningRecipe<Set<NamedVariable>> {
+public class JodaTimeRecipe extends ScanningRecipe<JodaTimeRecipe.Accumulator> {
     @Override
     public String getDisplayName() {
-        return "Migrate Joda Time to Java Time";
+        return "Migrate joda time to java time";
```
timtebeek, line 30: Some capitalization is actually fine here; just not every word.
```suggestion
        return "Migrate Joda-Time to Java time";
```

timtebeek (approved):
Thanks for the continued work here @amishra-u ! Really coming together nicely.

PR #612, opened 2024-11-21
fix: add java version property only to parent pom

dpozinen:
## What's changed?
Fix bug where java version property would be add to *all* submodules of a project instead of only to the root pom

## Any additional context
I was having issues with `org.openrewrite.java.migrate.UpgradeToJava21`

Specifically, it adds the `<java.version>` property to each submodule, even if the parent module (in the same repo) already contains this property. 
I believe this is because of [this logic](https://github.com/openrewrite/rewrite-migrate-java/blob/ead663e7da49ea521341d786d25a2a903070d7e5/src/main/java/org/openrewrite/java/migrate/maven/UpdateMavenProjectPropertyJavaVersion.java#L91-L95) in `UpdateMavenProjectPropertyJavaVersion` which checks if relativePath has a value. 

But the problem is that in most cases `relativePath` is omitted completely, because if not specified, it defaults to `../pom.xml `. 
Now I realise the use case of `<relativePath/>` so it was preserved.

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/maven/UpdateMavenProjectPropertyJavaVersion.java
+++ b/src/main/java/org/openrewrite/java/migrate/maven/UpdateMavenProjectPropertyJavaVersion.java
@@ -29,6 +29,7 @@ import org.openrewrite.xml.tree.Xml;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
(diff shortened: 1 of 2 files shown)


sambsnyd:
Merged this here: https://github.com/openrewrite/rewrite-migrate-java/commit/02c1a54a25545eff7764b64575456be92277f554

Thanks for reporting!

PR #614, opened 2024-12-02
Fix multiple issues in JodaTime to JavaTime migration recipe

amishra-u:
## What's changed?
Fixed these bugs in JodaTime to JavaTime migration recipe..
1. Updated the code to handle scenarios where variable declarations, such as lambda parameters, might lack a type expression.
2. Recipe was throwing NPE for any variable declaration of unimplemented type. 
4. Fixed an issue where the updated assignment expression was incorrectly passed instead of the original one to fetch the migration template.
5. Javadoc may not always have valid method reference. Skipped migration of reference in javadoc.

### Checklist


timtebeek (approved):

timtebeek:
Thanks again @amishra-u !

amishra-u:
> Thanks again @amishra-u !

Thanks for the review.

PR #617, opened 2024-12-03
Joda-Time to Java time: Add templates for Joda Interval to Threeten-extra Interval

amishra-u:
## What's changed?
Added migration templates to convert Joda-Time Interval classes to the [ThreeTen-extra](https://www.threeten.org/threeten-extra/) Interval.  The Interval class is not part of the JDK but is implemented in ThreeTen-extra, built on top of Java's java.time classes. This change include migration templates for methods from the following Joda-Time classes:

1. org.joda.time.Interval
2. org.joda.time.AbstractInterval

## What's your motivation?
Increase coverage for Joda-Time to Java Time migration recipe by adding support for additional classes.

### Checklist

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -50,6 +50,7 @@ dependencies {
 
     testImplementation("com.google.guava:guava:33.0.0-jre")
     testImplementation("joda-time:joda-time:2.12.3")
+    testImplementation("org.threeten:threeten-extra:1.8.0")
 
     testRuntimeOnly("com.fasterxml.jackson.datatype:jackson-datatype-jsr353")
     testRuntimeOnly("com.fasterxml.jackson.core:jackson-core")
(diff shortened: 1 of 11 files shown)


timtebeek (commented):
Great to see! the only thing I'm wondering is if there should be a declarative recipe to add that `threeten-extra` dependency if not already present (transitively). Just to ensure there's no compilation failures due to a missing dependency after the migration.

amishra-u:
> Great to see! the only thing I'm wondering is if there should be a declarative recipe to add that `threeten-extra` dependency if not already present (transitively). Just to ensure there's no compilation failures due to a missing dependency after the migration.

I’ve added a declarative recipe. (Hoping AddDependency will support Bazel someday 🤞)

timtebeek (approved):
Thanks a lot once again!

PR #622, opened 2024-12-07
Fix bugs in JodaTime to JavaTime migration recipe

amishra-u:
## What's changed?
Fixed following bugs in JodaTime to JavaTime migration recipe..

1. **MethodInvocation Migration**: When the select of a [MethodInvocation](https://github.com/openrewrite/rewrite/blob/e536ed273fda87ce6857b06f200488130b3ef5f3/rewrite-java/src/main/java/org/openrewrite/java/tree/J.java#L3825) is a MethodCall and the select has not been migrated, the migration template is no longer incorrectly applied to the MethodInvocation.
2. **Variable Initializer Handling:** Variables declared inside variable initializers of **non-joda type** were previously skipped by the [JodaTimeScanner](https://github.com/openrewrite/rewrite-migrate-java/blob/39507e5c647d8459ac6fd1917d1795649d94a568/src/main/java/org/openrewrite/java/migrate/joda/JodaTimeScanner.java#L66). Due to which scanner failed to mark unsafe variables in such contexts. Fixed it.

Note: During debugging also found [issue](https://github.com/openrewrite/rewrite-analysis/issues/72) with FlowPath Analysis added unit test for the same

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/joda/JodaTimeScanner.java
+++ b/src/main/java/org/openrewrite/java/migrate/joda/JodaTimeScanner.java
@@ -63,7 +63,7 @@ class JodaTimeScanner extends ScopeAwareVisitor {
     @Override
     public NamedVariable visitVariable(NamedVariable variable, ExecutionContext ctx) {
         if (!variable.getType().isAssignableFrom(JODA_CLASS_PATTERN)) {
-            return variable;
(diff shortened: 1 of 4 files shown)


timtebeek (approved):
Thanks for the continued work here and in rewrite-analysis @amishra-u !

PR #626, opened 2024-12-11
Joda-Time to Java time: Add support for Method Return Statement Migration

amishra-u:
## What's changed?
- Added support for safely migrating method return statements of the Joda type.
- Migration Safety Criteria: A method return statement is considered safe to migrate if:
   - The return statement itself can be safely migrated.
   - All invocations of the method are safe to migrate.

### Key Changes:
1. Used a map of methodReferencedVars to track the referenced variables during assignment and in return expression. (A method invocation is unsafe for migration if any of referenced var is unsafe)
2. Added a safeMethodMap to track the safe methods
3. Added new test cases and updated existing ones to reflect the changes.

### Not Implemented Yet:
1. Migration of class variables.

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/joda/JodaTimeRecipe.java
+++ b/src/main/java/org/openrewrite/java/migrate/joda/JodaTimeRecipe.java
@@ -54,6 +54,7 @@ public class JodaTimeRecipe extends ScanningRecipe<JodaTimeRecipe.Accumulator> {
     @Getter
     public static class Accumulator {
         private final Set<NamedVariable> unsafeVars = new HashSet<>();
+        private final Map<JavaType.Method, Boolean> safeMethodMap = new HashMap<>();
         private final VarTable varTable = new VarTable();
     }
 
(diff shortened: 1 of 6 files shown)


timtebeek (approved):
Great to see @amishra-u ; thanks for the continued work here. I was slightly surprised to see methods migrated that have a public modifier and Joda-Time return type, but I suppose at some point we're going to have to make some changes if folks are to migrate at all. Let's hope folks understand that this is a one time migration that is hard to make guaranteed safe.

amishra-u:
> Great to see @amishra-u ; thanks for the continued work here. I was slightly surprised to see methods migrated that have a public modifier and Joda-Time return type, but I suppose at some point we're going to have to make some changes if folks are to migrate at all. Let's hope folks understand that this is a one time migration that is hard to make guaranteed safe.

Yes, it needs to run on the entire repository at once to ensure a safe migration. I've noticed that many projects have utility methods around DateTime. If these utility methods aren't migrated, the coverage becomes minimal because Joda-Time types are tightly interconnected. For example, intervals depend on DateTime, DateTime relies on time zones, and this interconnectedness makes everything unsafe if those utility methods are migrated

PR #631, opened 2024-12-15
feat: normalize effective getter methods

timo-a:
## What's changed?

Adds recipe that renames "effective getter methods" to the way lombok would name them.

## What's your motivation?

When a team starts without lombok, their getter and setter methods will over time stop adhering to the way lombok names them. Maybe the team never followed the lombok naming, maybe fields got renamed but not their getter methods. Maybe previously more complex methods morph into just being just getters but keep their old name...
This recipe will normalize those methods names, so the converter can find them.  
This is a separate recipe on purpose. Reasons:
 - due to the edge cases listed in the recipe, it makes sense to split functionality to keep things simple.
 - maybe some recipes follow a different naming convention on purpose. This recipe will make it easy to spot these cases which might be overlooked in the many changes of one big fuzzy recipe
 - if a user only wants to convert lombok naming compliant getters and not rename any recipes, separate recipes make that possible

Please evaluate the edge cases. I've estimated that the effort needed to cover them is not worth it, but maybe you can see a way.
Also, I overloaded a method in `LombokUtils` with the original version. Maybe the methods can be merged?

### Checklist


timo-a:
Hi @timtebeek, can you please take another look at this? I've applied all changes I've seen you do on the other PRs.

timtebeek:
Thanks for making those adjustments! That's indeed been holding back a review because of the time investment needed to adjust the PRs. We'll have another look!

timtebeek (approved):
This one is looking good @timo-a , thanks a lot for your patience throughout. The only thing I'm wondering is whether `NormalizeGetter` is descriptive enough for folks, or if we should perhaps use something like `AdoptLombokGetterMethodNames`, as could be clearer in the intent.

timo-a:
Yeah, I guess normalize is a bit vague. I've implemented your suggestion.

timtebeek (approved):

timtebeek:
Thanks a lot! We'll merge as soon as the tests pass, and then this should go out with the release ~Wednesday

PR #632, opened 2024-12-15
New recipe to adopt Lombok setter method names

timo-a:
- analogous to https://github.com/openrewrite/rewrite-migrate-java/pull/631

probably best to work on that one first and then come back here.


timtebeek:
The rename suggested here, as well as the other fixes there can likely be applied here as well for consistency
- https://github.com/openrewrite/rewrite-migrate-java/pull/631

timtebeek (approved):
Great to see! Thanks for bringing parity and applying the feedback left on the Getter PR here as well.

PR #638, opened 2024-12-19
fix: Skip code references in comment

amishra-u:
## What's changed?
- fixes: #636 

Recipe doesn't have any handling for migrating code references in comments, so skip those.

### Checklist


timtebeek (approved):
Thanks for yet another quick fix.

PR #644, opened 2024-12-24
feat: use Lombok Log Annotations

timo-a:
## What's changed?

adds recipe that replaces explicit logger fields with the lombok annotation.

Naming of the declarative catch-all recipe.
The recipes have nullable arguments. Where should they be documented?

## Any additional context

More context here: https://timo-a.github.io/rewrite-recipe-starter/component-singleton/0/recipes/lombok/log/a_LogManual.html

Some annotations (like `@Flogger`) are missing. This is on purpose. I've done what I consider the most commonly used ones. People who need others can easily add them, based on what's already there.

The recipes allow for field names other than `log`, however they do not add `lombok.log.fieldName=customLoggerName` 
to `lombok.config`. Not sure how difficult that would be to implement. Would be nice to have but I think the effort of doing it manually manual is reasonable.

### Checklist


timtebeek (approved):
Thanks a lot for this work @timo-a ! I've made some changes to use Type and Method matchers when comparing current fields, pull up the majority of the logic into LogVisitor, and rename usages of any alternatively named logger fields. With that I think we're good to merge, thanks!

PR #646, opened 2024-12-24
feat: use @Data

timo-a:
### Checklist


timo-a:
Hi @timtebeek, could you have another look at this one as well?

PR #669, opened 2025-01-31
Java21 recipe deprecatedSecurityManager

ranuradh:
## What's changed?
Java 21 recipe: org.openrewrite.java.migrate.DeprecatedSecurityManager

## What's your motivation?
The default value of the java.security.manager system property has been changed to disallow since Java 18. Unless the system property is set to allow on the command line, any invocation of System.setSecurityManager(SecurityManager) with a non-null argument will throw an UnsupportedOperationException. You can set system property as Djava.security.manager=allow. 
This recipe calls 
``` System.setProperty("java.security.manager", "allow");```
whenever
 System.setSecurityManager(SecurityManager) is called

Reference : https://bugs.openjdk.org/browse/JDK-8271301 & https://bugs.openjdk.org/browse/JDK-8270380
@timtebeek 

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/DeprecatedSecurityManager.java
+++ b/src/main/java/org/openrewrite/java/migrate/DeprecatedSecurityManager.java
@@ -1,0 +1,83 @@
+/*
+ * Copyright 2025 the original author or authors.
+ * <p>
+ * Licensed under the Moderne Source Available License (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://docs.moderne.io/licensing/moderne-source-available-license
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.java.migrate;
+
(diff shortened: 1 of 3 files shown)


MBoegers:
Is adding a `System.setProperty()` call every time `System.setSecurityManager(SecurityManager)` is called, real the best general migration?
It more feels like silencing and would overwrite `-Djava.security.manager=false`settings with might be intended by users.
Can we find a better general migration?

ranuradh:
Hi @MBoegers  thanks for the feedback,  I based my recipe on https://bugs.openjdk.org/browse/JDK-8271301) where ```
 For most applications, the maintainer can simply add `-Djava.security.manager=allow` to the launcher command before they remove the `setSecurityManager` call. ``` Since I was not sure how to add `-Djava.security.manager=allow` to the launcher I figured calling ```System.setProperty()``` before the invocation of the call could work.

MBoegers:
I understand and your path. I have to gig a little deeper, what's the ultimate migration path.
I think adding via `setProperty` is not the same as adding the `-D` parameter. Let me please think for a night or two 🙏

timtebeek:
I"m not sure this approach would work; effectively you'd be trying to set a system property from a Java process that has already started, whereas I think this property should be set before the Java process is started. I don't think this is a change we can make a source-level adjustment for; this should change in the way the Java process is invoked after a migration to Java 21+. That's how I interpret the "Unless the system property is set to allow on the command line" above. I'd lean towards not having this as a recipe, as much as we appreciate the effort you've put in.

ranuradh:
Got it thanks will close the PR

PR #667, opened 2025-01-27
Java 11 recipe  for default keystore change 

ranuradh:
## What's changed?
recipe: org.openrewrite.java.migrate.ChangeDefaultKeyStore in Java11
image

## What's your motivation?
This recipe returns `jks` when Keystore.getDefaultType() is called

## Anyone you would like to review specifically?
@timtebeek 

### Checklist
- [X] I've added unit tests to cover both positive and negative cases
- [X] I've read and applied the [recipe conventions and best practices](https://docs.openrewrite.org/authoring-recipes/recipe-conventions-and-best-practices)
- [X] I've used the IntelliJ IDEA auto-formatter on affected files

PR #670, opened 2025-02-04
Polish ChangeDefaultKeyStore to only apply to Java 11

timtebeek:
## What's changed?
- Only apply to projects using Java 11 at the moment that we run the recipe.
- Ensure there are spaces between string concatenations in the description.
- Use the existing MethodMatcher in the Precondition as well.
- Extract a constant for a repeated test text block.
- Use more descriptive test method names.

## What's your motivation?
Projects using Java 17+ should not revert to an older default value.

Hope you agree with the changes here; I think it's best to get this in before a release.

## Any additional context
- https://github.com/openrewrite/rewrite-migrate-java/pull/667

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/ChangeDefaultKeyStore.java
+++ b/src/main/java/org/openrewrite/java/migrate/ChangeDefaultKeyStore.java
@@ -15,7 +15,6 @@
  */
 package org.openrewrite.java.migrate;
 
-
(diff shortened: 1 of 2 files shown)


MBoegers (approved):
Thanks for fixing..

ranuradh (approved):
thanks for updates Tim.

PR #672, opened 2025-02-11
Omit migration to @Setter annotated field of other annotations are present

MBoegers:
## What's changed?

- Reported in https://github.com/openrewrite/rewrite/issues/5015 the Lombok recipes don't migrate other annotations.

## What's your motivation?

prevent breaking changes
### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/lombok/LombokUtils.java
+++ b/src/main/java/org/openrewrite/java/migrate/lombok/LombokUtils.java
@@ -53,7 +53,7 @@ class LombokUtils {
             return false;
         }
         // Check there is no annotation except @Overwrite
-        if (noneAtAllOrOnlyOverrideAnnotated(cursor, service)) {
+        if (hasAnyAnnotatioOtherThanOverride(cursor, service)) {
             return false;
         }
         // Check field is declared on method type
@@ -120,7 +120,7 @@ class LombokUtils {
         }
 
         // Check there is no annotation except @Overwrite
-        if (noneAtAllOrOnlyOverrideAnnotated(cursor, service)) {
+        if (hasAnyAnnotatioOtherThanOverride(cursor, service)) {
             return false;
         }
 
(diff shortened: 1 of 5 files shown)


src/main/java/org/openrewrite/java/migrate/lombok/LombokUtils.java
```diff
@@ -141,4 +155,10 @@ static AccessLevel getAccessLevel(J.MethodDeclaration methodDeclaration) {
         }
         return PACKAGE;
     }
+
+    private static boolean noneAtAllOrOnlyOverwriteAnnotated(J.MethodDeclaration method) {
+        List<J.Annotation> annotations = method.getLeadingAnnotations();
+        return !annotations.isEmpty() &&
+                !(annotations.size() == 1 && overwriteMatcher.matches(annotations.get(0)));
```
timtebeek, line 162: Best to use the annotation service here as there might also be annotations on the return type for instance. Leading annotations only has a limited view on annotations.

MBoegers: We don't have a cursor here, it is a util outside a recipe.
2 options I see:

1. accept and note for final fix
2. move the check into recipe as additional check

knutwannheden: As LombokUtils is private it should be possible to add a Cursor parameter, right?

MBoegers: Refactored it in, I don't know if it is the right thing to do considering it's just a fix not a final solution ;)

MBoegers:
@timtebeek @knutwannheden The CI is failing because of missing type information at `@AllArgsConstructor` annotation. May this be a symtom of the changes at Lombok support? 
```log
MigrateCollectionsSingletonListTest > lombokAllArgsConstructor() FAILED
    java.lang.IllegalStateException: LST contains missing or invalid type information
    NewClass->EnumValue->EnumValueSet->Block->ClassDeclaration->Block->ClassDeclaration->CompilationUnit
    /*~~(NewClass type is missing or malformed)~~>*/new(singletonList(FOO))
```

timtebeek:
Indeed that's due to Lombok changes in openrewrite/rewrite. I hadn't yet made adjustments here as it seems we're close to a fix upstream. 🙏🏻

timtebeek (approved):
Great to see, thanks! I've added some slight polish: we tend to retrieve the AnnotationService through a service method defined on the visitor, to prevent repeated instantiation and to allow for instance the Kotlin visitor to swap in an alternative implementation.

PR #674, opened 2025-02-13
Align JakartaEE migrations with EE9 vs EE10 versions

tobli:
## What's changed?
This commit updates both the EE9 and the EE10 recipe to pin specification jars to the version corresponding to the umbrella EE version. The versions are aligned with https://github.com/jakartaee/jakartaee-api/blob/9.1.0/pom.xml and https://github.com/jakartaee/jakartaee-api/blob/10.0.0/pom.xml respectively.

Where runtime dependencies are added, e.g. `com.sun.xml.ws:jaxws-rt`, the implementation matching the spec version is selected: `com.sun.xml.ws:jaxws-rt:3.x` for `jakarta.xml.ws:jakarta.xml.ws-api:3.0.x` and `com.sun.xml.ws:jaxws-rt:4.x` for `jakarta.xml.ws:jakarta.xml.ws-api:4.0.x`

## What's your motivation?
The `org.openrewrite.java.migrate.jakarta.JavaxMigrationToJakarta` recipe states in the description that it will migrate to the EE9 version. However it updates both api and implementation jars to `latest.release`. As a consequence the EE9 migration will pick a version for EE10 or later. Since later Jakarta EE versions sometime remove or change api, this may lead to compile time or runtime errors.

### Checklist

### DIFF
--- a/src/main/resources/META-INF/rewrite/jakarta-ee-10.yml
+++ b/src/main/resources/META-INF/rewrite/jakarta-ee-10.yml
@@ -23,6 +23,7 @@ tags:
   - jakarta
 recipeList:
   - org.openrewrite.java.migrate.jakarta.JavaxMigrationToJakarta
+  - org.openrewrite.java.migrate.jakarta.MigrationToJakarta10Apis
   - org.openrewrite.java.migrate.jakarta.Faces2xMigrationToJakarta4x
   - org.openrewrite.java.migrate.jakarta.RemovedIsParmetersProvidedMethod
   - org.openrewrite.java.migrate.jakarta.RemovedSOAPElementFactory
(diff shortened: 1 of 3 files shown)


src/main/resources/META-INF/rewrite/jakarta-faces-4.yml
```diff
@@ -39,6 +40,16 @@ recipeList:
...
 ---
 type: specs.openrewrite.org/v1beta/recipe
+name: org.openrewrite.java.migrate.jakarta.UpdateJakartaFacesApi4
+displayName: Update Jakarta EE Java Faces Dependencies to 4.0.x.
+description: Update Jakarta EE Java Faces Dependencies to 4.0.x.
+recipeList:
+  - org.openrewrite.java.dependencies.UpgradeDependencyVersion:
+      groupId: jakarta.faces
+      artifactId: jakarta.faces-api
+      newVersion: 4.0.x
```
tobli, line 50: needed to add this, previously the EE9 migration included the Faces 4 (part of EE10) update

src/main/resources/META-INF/rewrite/jakarta-ee-9.yml
```diff
@@ -1006,11 +1008,11 @@ recipeList:
       oldArtifactId: javaee-api
       newGroupId: jakarta.platform
       newArtifactId: jakarta.jakartaee-api
-      newVersion: 9.0.0
+      newVersion: 9.1.0
   - org.openrewrite.java.dependencies.UpgradeDependencyVersion:
       groupId: jakarta.platform
       artifactId: "*"
-      newVersion: 9.0.0
+      newVersion: 9.1.0
```
tobli, line 1015: There are no api changes with EE 9.1, just Java11 support

src/main/resources/META-INF/rewrite/jakarta-ee-9.yml
```diff
@@ -816,93 +818,93 @@ recipeList:
       oldArtifactId: jackson-module-jaxb-annotations
       newGroupId: com.fasterxml.jackson.module
       newArtifactId: jackson-module-jakarta-xmlbind-annotations
-      newVersion: 2.13.x
+      newVersion: 2.18.x
```
tobli, line 821: I'm open to moving this to a separate PR

src/main/resources/META-INF/rewrite/jakarta-ee-9.yml
```diff
@@ -104,21 +104,23 @@ recipeList:
...
   - org.openrewrite.java.ChangePackage:
       oldPackageName: javax.annotation
       newPackageName: jakarta.annotation
       recursive: true
   - org.openrewrite.java.ChangePackage:
       oldPackageName: jakarta.annotation.processing
       newPackageName: javax.annotation.processing
-
+  - org.openrewrite.java.ChangePackage:
+      oldPackageName: jakarta.annotation.concurrent
```
tobli, line 122: `jakarta.annotation.concurrent` is not part of the EE9 api

src/main/resources/META-INF/rewrite/jakarta-ee-10.yml
```diff
@@ -37,7 +38,80 @@ recipeList:
...
+description: Jakarta EE 10 updates some apis compared to Jakarta EE 9.
+tags:
+  - jakarta
+recipeList:
   - org.openrewrite.java.migrate.jakarta.UpdateJakartaAnnotations2
+  - org.openrewrite.java.migrate.jakarta.UpdateJakartaXmlWsEE10
+  - org.openrewrite.java.dependencies.UpgradeDependencyVersion:
+      groupId: jakarta.activation
+      artifactId: jakarta.activation-api
+      newVersion: 2.1.x
```
tobli, line 54: I started with a "catch all" api version recipe, but I could break it out to a separate recipe per spec

src/main/resources/META-INF/rewrite/jakarta-ee-9.yml
timtebeek: Might be best indeed not to make this change here; I think ideally we go for a low but good enough default value, such that folks are likely not to see changes applied because they are already on a newer version. If we bump this as seen here we might then get complaints about unnecessary bumps to Jackson. Thanks for pointing it out!

timtebeek (commented):
Thanks a lot for the changes here! Might not feel very exciting in the moment, but as you've found it's good to be deliberate about the versions used such that there's minimal surprises for folks targeting an intermediate version of EE, or see that included in their composites.

I've added one comment on the Jackson migration thread above; other than that I think this looks good already.

src/main/resources/META-INF/rewrite/jakarta-ee-10.yml
timtebeek: Thanks for picking deliberate versions here as well, such that we don't see the same problem repeated in the future. It might help to add a comment just above here (and below for EE9) to show where you've picked these version numbers from. That way for future recipes we know where to look.

src/main/resources/META-INF/rewrite/jakarta-ee-9.yml
tobli: ✅ I'll take that out and repush

tobli:
> Thanks a lot for the changes here! Might not feel very exciting in the moment, but as you've found it's good to be deliberate about the versions used such that there's minimal surprises for folks targeting an intermediate version of EE, or see that included in their composites.
> 
> I've added one comment on the Jackson migration thread above; other than that I think this looks good already.

Thanks! Adding this would be very helpful, I've been bitten by the compatibility issues myself.

I took out the Jackson changes, as you noted they are not strictly necessary. Better to keep the scope small.

src/main/resources/META-INF/rewrite/jakarta-ee-10.yml
tobli: ✅ I can add a link to the respective BOM files (linked in the PR)

tobli: done

src/main/resources/META-INF/rewrite/jakarta-ee-10.yml
```diff
@@ -37,7 +38,81 @@ recipeList:
...
   - org.openrewrite.java.migrate.jakarta.JavaxToJakartaCdiExtensions
   - org.openrewrite.java.migrate.jakarta.UpdateJakartaPlatform10
+---
+type: specs.openrewrite.org/v1beta/recipe
+name: org.openrewrite.java.migrate.jakarta.MigrationToJakarta10Apis
+displayName: Migrate Jakarta EE 9 api dependencies to Jakarta EE 10 versions
+description: Jakarta EE 10 updates some apis compared to Jakarta EE 9.
+tags:
+  - jakarta
+# NOTE: The spec versions in this section comes from https://github.com/jakartaee/jakartaee-api/blob/10.0.0/pom.xml
```
tobli, line 48: linking to the BOM where the spec versions are defined

src/main/resources/META-INF/rewrite/jakarta-ee-9.yml
```diff
@@ -25,6 +25,7 @@ tags:
   - jakarta
 # TODO: Update XML references if necessary.
 # TODO: Rename bootstrapping files if necessary.
+# NOTE: The spec versions in this section comes from https://github.com/jakartaee/jakartaee-api/blob/9.1.0/pom.xml
```
tobli, line 28: linking to the BOM where the spec versions are defined

timtebeek (approved):
Thanks a lot for the quick turnaround here @tobli !

PR #691, opened 2025-03-20
Fix #688: Update @ActivationConfigPropery/@JMSDestination

### LINKED ISSUE #688: JMS is not migrated properly when using ...migrate.jakarta.JakartaEE10
I ran
```sh
mvn -U org.openrewrite.maven:rewrite-maven-plugin:run \
    -Drewrite.recipeArtifactCoordinates=org.openrewrite.recipe:rewrite-migrate-java:RELEASE \
    -Drewrite.activeRecipes=org.openrewrite.java.migrate.jakarta.JakartaEE10
```
And found some issues when running the application after migrating:
```java
@JMSDestinationDefinition(name = xxx,
                          interfaceName = "javax.jms.Topic",
                          destinationName = "xxx")
```
```java
@ActivationConfigProperty(propertyName = "destinationType", propertyValue = "javax.jms.Topic")
```
I'm not sure how the rewrite rules work for Java code. It seems that it's not a full replace of `javax.` by `jakarta.` for the general migration (as we are left with some `javax.` instances). In this case I think a full replace of `javax.jms.` by `jakarta.jms.` would do.

CC @melloware

melloware:
- Fix #688: Update @ActivationConfigPropery/@JMSDestination


timtebeek (approved):
Great to see! I've adopted the annotated trait here, and merged the two recipes into one. I've also changes the signature we look for, as where included in the composite recipe we're already expecting jakarta anntations at that point, which otherwise would be missed.

I had half expected these changes to be picked up by ChangePackage when combined with the type references we added a couple months ago. Do we not support type references on annotation attributes @jevanlingen ?

Fine to merge this as is for now; if we do support this through type references at some point we can remove this recipe but keep the tests.

melloware:
If you do refactor we have something similar already for `BeanValidationMessages.java` and `BeanValidationMessagesTest.java`  I think all these recipes could be simplified.

melloware (commented):
Awesome

timtebeek:
Thanks for the quick hint! Nice to see these additional cases covered with less code to boot.

PR #699, opened 2025-04-02
Add Getter method to `useMimeCoder` field in `UseJavaUtilBase64` recipe

wapkch:
…cipe

## What's changed?

Add Getter method to `useMimeCoder` field in `UseJavaUtilBase64` recipe

## What's your motivation?

It seems that when running a recipe with configuration parameters, there must be a Getter method to the parameter field in recipe.

For example:

```
---
name: UseJavaUtilBase64
displayName: UseJavaUtilBase64
description: UseJavaUtilBase64.
recipeList:
  - org.openrewrite.java.migrate.UseJavaUtilBase64:
      useMimeCoder: true

```

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/UseJavaUtilBase64.java
+++ b/src/main/java/org/openrewrite/java/migrate/UseJavaUtilBase64.java
@@ -16,6 +16,7 @@
 package org.openrewrite.java.migrate;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
+import lombok.Getter;
 import org.openrewrite.*;
 import org.openrewrite.java.ChangeType;
 import org.openrewrite.java.JavaTemplate;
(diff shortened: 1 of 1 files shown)


greg-at-moderne (approved):
For reference, @wapkch can you also paste what error happens when this is absent?

wapkch:
> For reference, @wapkch can you also paste what error happens when this is absent?

When this is absent, `useMimeCoder` field will always be `fasle` even if configured to true like this :

```
---
name: UseJavaUtilBase64
displayName: UseJavaUtilBase64
description: UseJavaUtilBase64.
recipeList:
  - org.openrewrite.java.migrate.UseJavaUtilBase64:
      useMimeCoder: true

```

greg-at-moderne:
OK. Understood. Thanks for the fix!

I've checked other pieces of code with `@Option` and most of them don't use the `@Getter`. But they use `@Value` on the whole class. But `@Getter` makes sense too.

PR #700, opened 2025-04-07
Adapt to new arg of AddDependencyVisitor

greg-at-moderne:
**What**:
Adding a new arg to the call of `AddDependencyVisitor`

**Why**:
Fixing CI after https://github.com/openrewrite/rewrite/pull/5252


timtebeek:
Thank you! Really helpful not to discover these as we run the release. :)

PR #707, opened 2025-04-18
Improved bouncy castle implementation

JohannisK:
## What's changed?
- Removed upgrade to java 8
- Improved tests

### DIFF
--- a/src/test/java/org/openrewrite/java/migrate/BouncyCastleTest.java
+++ b/src/test/java/org/openrewrite/java/migrate/BouncyCastleTest.java
@@ -20,13 +20,11 @@ import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.ValueSource;
 import org.openrewrite.DocumentExample;
 import org.openrewrite.maven.tree.MavenResolutionResult;
-import org.openrewrite.maven.tree.ResolvedDependency;
 import org.openrewrite.maven.tree.Scope;
 import org.openrewrite.test.RecipeSpec;
 import org.openrewrite.test.RewriteTest;
 
-import java.util.List;
-
+import static java.util.function.UnaryOperator.identity;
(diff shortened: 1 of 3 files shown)


src/test/java/org/openrewrite/java/migrate/BouncyCastleTest.java
```diff
@@ -67,21 +69,34 @@ void updateBouncyCastle() {
...
                 </project>
                 """,
-              spec -> spec.after(str -> assertThat(str)
-                .doesNotContainPattern("\\h*<groupId>org\\.bouncycastle<\\/groupId>\\s+<artifactId>bcprov-jdk15on<\\/artifactId>\\s+<version>.*<\\/version>")
-                .doesNotContainPattern("\\h*<groupId>org\\.bouncycastle<\\/groupId>\\s+<artifactId>bcpkix-jdk15on<\\/artifactId>\\s+<version>.*<\\/version>")
-                .containsPattern("\\h*<groupId>org\\.bouncycastle<\\/groupId>\\s+<artifactId>bcprov-jdk18on<\\/artifactId>\\s+<version>.*<\\/version>")
-                .containsPattern("\\h*<groupId>org\\.bouncycastle<\\/groupId>\\s+<artifactId>bcpkix-jdk18on<\\/artifactId>\\s+<version>.*<\\/version>")
-                .actual())
+              spec -> {
+                  spec.after(e -> e);
```
JohannisK, line 73: It only works when I include the `spec.after`, jus `spec.afterRecipe` is not enough as shown by the test below.

timtebeek (approved):

src/test/java/org/openrewrite/java/migrate/BouncyCastleTest.java
timtebeek: Ah yes, without that `after` you'd be asserting no change is made, but that it's parser correctly. Added that now.

PR #708, opened 2025-04-22
Replace the jacoco supporting Gradle 8 deprecated properties

Jenson3210:
In order to migrate jacoco to newer version / gradle version, we have to replace the properties as mentioned in [Gradle docs](https://docs.gradle.org/current/userguide/upgrading_version_7.html#report_and_testreport_api_cleanup). 

This should leak also in the spring recipes as there we also bump jacoco.

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/jacoco/Gradle8JacocoReportDeprecations.java
+++ b/src/main/java/org/openrewrite/java/migrate/jacoco/Gradle8JacocoReportDeprecations.java
@@ -1,0 +1,114 @@
+/*
+ * Copyright 2024 the original author or authors.
+ * <p>
+ * Licensed under the Moderne Source Available License (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://docs.moderne.io/licensing/moderne-source-available-license
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.java.migrate.jacoco;
+
+import lombok.EqualsAndHashCode;
+import lombok.Value;
+import org.jetbrains.annotations.NotNull;
+import org.openrewrite.ExecutionContext;
+import org.openrewrite.Preconditions;
+import org.openrewrite.Recipe;
+import org.openrewrite.TreeVisitor;
+import org.openrewrite.gradle.IsBuildGradle;
+import org.openrewrite.groovy.GroovyIsoVisitor;
+import org.openrewrite.java.tree.Expression;
+import org.openrewrite.java.tree.J;
+
+@Value
+@EqualsAndHashCode(callSuper = false)
+public class Gradle8JacocoReportDeprecations extends Recipe {
+
+    @Override
+    public @NotNull String getDisplayName() {
+        return "Replace gradle 8 introduced deprecations in jacoco report task";
+    }
+
+    @Override
+    public @NotNull String getDescription() {
+        return "Set the `enabled` to `required` and the `destination` to `outputLocation` for Reports deprecations that were removed in gradle 8. " +
+                "See [the gradle docs on this topic](https://docs.gradle.org/current/userguide/upgrading_version_7.html#report_and_testreport_api_cleanup).";
+    }
+
+    @Override
+    public @NotNull TreeVisitor<?, ExecutionContext> getVisitor() {
+        return Preconditions.check(new IsBuildGradle<>(), new GroovyIsoVisitor<ExecutionContext>() {
+            @Override
+            public J.@NotNull Assignment visitAssignment(J.@NotNull Assignment assignment, @NotNull ExecutionContext ctx) {
+                String prefix = getCursor().getNearestMessage("path");
+                String prefixOrEmpty = prefix == null ? "" : prefix + ".";
+                if (assignment.getVariable() instanceof J.FieldAccess) {
+                    J.FieldAccess fieldAccess = (J.FieldAccess) assignment.getVariable();
+                    String fieldName = getFieldName(fieldAccess);
+                    return replaceDeprecations(assignment, prefixOrEmpty + fieldName);
+                } else if (assignment.getVariable() instanceof J.Identifier) {
+                    J.Identifier identifier = (J.Identifier) assignment.getVariable();
+                    return replaceDeprecations(assignment, prefixOrEmpty + identifier.getSimpleName());
+                }
+                return assignment;
+            }
+
+            @Override
+            public J.@NotNull MethodInvocation visitMethodInvocation(J.@NotNull MethodInvocation method, @NotNull ExecutionContext executionContext) {
+                String parent = getCursor().getNearestMessage("path");
+                if (parent == null) {
+                    getCursor().putMessage("path", method.getSimpleName());
+                } else {
+                    getCursor().putMessage("path", parent + "." + method.getSimpleName());
+                }
+                return super.visitMethodInvocation(method, executionContext);
+            }
+
+            private @NotNull J.Assignment replaceDeprecations(J.@NotNull Assignment assignment, String path) {
+                String[] parts = path.split("\\.");
+                if (parts.length == 4 && "jacocoTestReport".equalsIgnoreCase(parts[0]) && "reports".equalsIgnoreCase(parts[1])) {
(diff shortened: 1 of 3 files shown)


src/main/java/org/openrewrite/java/migrate/jacoco/Gradle8JacocoReportDeprecations.java
```diff
@@ -0,0 +1,113 @@
...
+import org.openrewrite.java.tree.Expression;
+import org.openrewrite.java.tree.J;
+
+@Value
+@EqualsAndHashCode(callSuper = false)
+public class Gradle8JacocoReportDeprecations extends Recipe {
+
+    @Override
+    public String getDisplayName() {
+        return "Replace gradle 8 introduced deprecations in jacoco report task";
```
timtebeek, line 35:
```suggestion
        return "Replace Gradle 8 introduced deprecations in JaCoCo report task";
```

src/test/java/org/openrewrite/java/migrate/jacoco/Gradle8JacocoReportDeprecationsTest.java
```diff
@@ -0,0 +1,465 @@
...
+
+              """
+          )
+        );
+    }
+
+    @Test
+    void destinationInAnotherExtensionNotTouched() {
+        rewriteRun(
+          spec -> spec.recipe(new Gradle8JacocoReportDeprecations()),
```
timtebeek, line 449: Since all tests run the same recipe, can we configure that once by overriding the defaults method? That would allow us to more easily spot when some cases are different.

src/test/java/org/openrewrite/java/migrate/jacoco/Gradle8JacocoReportDeprecationsTest.java
```diff
@@ -0,0 +1,465 @@
...
+              jacocoTestReport.reports.xml.outputLocation = layout.buildDirectory.dir('jacocoXml')
+              jacocoTestReport.reports.csv.outputLocation = layout.buildDirectory.dir('jacocoCsv')
+              jacocoTestReport.reports.html.outputLocation = layout.buildDirectory.dir('jacocoHtml')
+
+              """
+          )
+        );
+    }
+
+    @DocumentExample
```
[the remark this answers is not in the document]
Jenson3210: One example shows what will happen with `enabled`, another what will happen with `destination`. 
Could also select only 1

src/main/resources/META-INF/rewrite/jacoco.yml
```diff
@@ -34,3 +34,4 @@ recipeList:
       groupId: org.jacoco
       artifactId: jacoco-maven-plugin
       newVersion: 0.8.x
+  - org.openrewrite.java.migrate.jacoco.Gradle8JacocoReportDeprecations
```
timtebeek, line 37: If this is required for a gradle 8 upgrade, should we then move this to our gradle upgrades recipes instead of having it here? Mostly wondering, as while I understand you placed it here, folks are more likely to expect this perhaps when upgrading gradle, not Java, even if a Java upgrade updates both gradle and jacoco. What are your thoughts here?

src/main/java/org/openrewrite/java/migrate/jacoco/Gradle8JacocoReportDeprecations.java
```diff
@@ -0,0 +1,113 @@
...
+
+    @Override
+    public String getDescription() {
+        return "Set the `enabled` to `required` and the `destination` to `outputLocation` for Reports deprecations that were removed in gradle 8. " +
+                "See [the gradle docs on this topic](https://docs.gradle.org/current/userguide/upgrading_version_7.html#report_and_testreport_api_cleanup).";
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return Preconditions.check(new IsBuildGradle<>(), new GroovyIsoVisitor<ExecutionContext>() {
```
timtebeek, line 46: Should we use the Java visitor here to support Kotlin too?

src/main/java/org/openrewrite/java/migrate/jacoco/Gradle8JacocoReportDeprecations.java
```diff
@@ -0,0 +1,113 @@
...
+            @Override
+            public J. Assignment visitAssignment(J. Assignment assignment, ExecutionContext ctx) {
+                String prefix = getCursor().getNearestMessage("path");
```
timtebeek, line 49: Now you have me wondering if cursor messaging is unique to the visitor, as `path`is very generic of course. And funnily enough I don't have the answer (reviewing from the train).

src/main/resources/META-INF/rewrite/jacoco.yml
Jenson3210: I was thinking the same... But, I put it here as it belongs together with the jacoco bump, not with the java upgrade. 
However, looking back at this + considering jacoco is default built-in on gradle, there would not be any trigger for me to run this recipe if running jacoco using gradle (as I do not need to update any plugin or anything. 

We can move it to the gradle8 migration indeed and then link to it from this recipe if you think that's fine? 
Personally, I would then create new package under [the main folder](https://github.com/openrewrite/rewrite/tree/main/rewrite-gradle/src/main/java/org/openrewrite/gradle) per version where we can put deprecated properties/replacements? and not per tool like here. As we can have multiple such properties per release.

src/main/java/org/openrewrite/java/migrate/jacoco/Gradle8JacocoReportDeprecations.java
```diff
@@ -0,0 +1,113 @@
...
+                return assignment;
+            }
+
+            @Override
+            public J. MethodInvocation visitMethodInvocation(J. MethodInvocation method, ExecutionContext ctx) {
+                String parent = getCursor().getNearestMessage("path");
+                if (parent == null) {
+                    getCursor().putMessage("path", method.getSimpleName());
+                } else {
+                    getCursor().putMessage("path", parent + "." + method.getSimpleName());
```
timtebeek, line 68: I've not seen this pattern of usage before. Elsewhere we tend to use the cursor stack to look up from any element on the tail end of what we're looking for. Using this approach I think we would be building up these string concatenations for any method invocations right? That might be a tad expensive. 

If we do wish to keep using this pattern, might it be possible to only traverse down the stack when we see, or have seen `jacocoTestReport`? That would remove all of the excess string concatenations for unrelated cases.

src/main/java/org/openrewrite/java/migrate/jacoco/Gradle8JacocoReportDeprecations.java
Jenson3210: No, tested locally with a kts based gradle smaple project. There went back till gradle 6 -> no enabled property existed. It was always this one already it seems. 
Project does not compile with gradle 6 + enabled in kts version

src/main/java/org/openrewrite/java/migrate/jacoco/Gradle8JacocoReportDeprecations.java
Jenson3210: changing it to "JACOCO_SETTINGS_PATH"

src/main/resources/META-INF/rewrite/jacoco.yml
timtebeek: Yes that could work, thanks

src/main/java/org/openrewrite/java/migrate/jacoco/Gradle8JacocoReportDeprecations.java
Jenson3210: @timtebeek changed this one

src/test/java/org/openrewrite/java/migrate/jacoco/Gradle8JacocoReportDeprecationsTest.java
Jenson3210: @timtebeek changed this one

src/main/resources/META-INF/rewrite/jacoco.yml
Jenson3210: I put some more thought in it: as the original thought was to fix the spring boot upgrade recipe, I would just add the deprecations of this property in the upgradeGradleVersion recipes and not in this one. One can perfectly use jacoco newer version in gradle 7 with deprecated properties. As of 8, they are removed... So it is really 100% linked with gradle version and not with jacoco version. 

Hence, I will add the recipe to the MigrateToGradle8 recipe for now. (which is also called in spring boot upgrade recipe -> also fixes the original issue)

Jenson3210:
Moved to https://github.com/openrewrite/rewrite/pull/5336

PR #713, opened 2025-04-30
Security Manager API Removal

MBoegers:
## What's changed?

With Java 24 the Security Manager API pushes users from using itself. We can help by removing the methods.

### Checklist

### DIFF
--- a/src/main/resources/META-INF/rewrite/java-version-25.yml
+++ b/src/main/resources/META-INF/rewrite/java-version-25.yml
@@ -1,5 +1,5 @@
 #
-# Copyright 2024 the original author or authors.
+# Copyright 2025 the original author or authors.
 # <p>
 # Licensed under the Moderne Source Available License (the "License");
 # you may not use this file except in compliance with the License.
--- a/src/main/resources/META-INF/rewrite/java-version-25.yml
+++ b/src/main/resources/META-INF/rewrite/java-version-25.yml
@@ -41,8 +41,8 @@ recipeList:
   - org.openrewrite.java.RemoveMethodInvocations:
       methodPattern: java.security.AccessController#checkPermission(..)
   - org.openrewrite.java.RemoveMethodInvocations:
-      methodPattern: import java.security.Policy#setPolicy(..)
+      methodPattern: java.security.Policy#setPolicy(..)
   - org.openrewrite.java.RemoveMethodInvocations:
-      methodPattern: import java.lang.SecurityManager#check*(..)
+      methodPattern: java.lang.SecurityManager#check*(..)
   - org.openrewrite.staticanalysis.UnnecessaryCatch
   - org.openrewrite.staticanalysis.UnnecessaryThrows
--- a/src/main/resources/META-INF/rewrite/java-version-25.yml
+++ b/src/main/resources/META-INF/rewrite/java-version-25.yml
@@ -1,0 +1,48 @@
+#
+# Copyright 2024 the original author or authors.
+# <p>
+# Licensed under the Moderne Source Available License (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+# <p>
+# https://docs.moderne.io/licensing/moderne-source-available-license
+# <p>
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+---
+type: specs.openrewrite.org/v1beta/recipe
+name: org.openrewrite.java.migrate.UpgradeToJava25
+displayName: Migrate to Java 25
+description: >-
+  This recipe will apply changes commonly needed when migrating to Java 25. This recipe will also replace deprecated API
+  with equivalents when there is a clear migration strategy. Build files will also be updated to use Java 25 as the
+  target/source and plugins will be also be upgraded to versions that are compatible with Java 25.
+tags:
+  - java25
+recipeList:
+  - org.openrewrite.java.migrate.RemoveSecurityManagerAPI
+
+---
+type: specs.openrewrite.org/v1beta/recipe
+name: org.openrewrite.java.migrate.RemoveSecurityManagerAPI
+displayName: Remove Security Manager API
+description: The Security Manager API is unsupported in Java 24. This recipe will remove the API and any usages of it.
+tags:
+  - java25
+preconditions:
+  - org.openrewrite.java.migrate.search.FindJavaVersion:
+      version: 24
+recipeList:
+  - org.openrewrite.java.RemoveMethodInvocations:
+      methodPattern: java.security.AccessController#checkPermission(*)
+  - org.openrewrite.java.RemoveMethodInvocations:
+      methodPattern: import java.security.Policy#setPolicy(*)
+  - org.openrewrite.java.RemoveMethodInvocations:
+      methodPattern: import java.lang.SecurityManager#check*(*)
+  - org.openrewrite.staticanalysis.UnnecessaryCatch
+  - org.openrewrite.staticanalysis.UnnecessaryThrows
--- a/src/main/resources/META-INF/rewrite/java-version-25.yml
+++ b/src/main/resources/META-INF/rewrite/java-version-25.yml
@@ -34,14 +34,71 @@ displayName: Remove Security Manager API
 description: The Security Manager API is unsupported in Java 24. This recipe will remove the API and any usages of it.
 tags:
   - java25
+  - security
+  - deprecation
 preconditions:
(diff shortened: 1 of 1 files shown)


MBoegers:
Other GitHub Examples: [full](https://github.com/Scott-Cooper/Drawbot_image_to_gcode_v2/blob/34fa3472e3e1093129032de62230beaa817b2084/GClip.java#L1-L119)
```java
SecurityManager security = System.getSecurityManager();
if (security != null) {
    try {
        security.checkSystemClipboardAccess();
        clipboard = Toolkit.getDefaultToolkit().getSystemClipboard();
    } catch (SecurityException e) {
        clipboard = new Clipboard("Application Clipboard");
    }
}
```
Should become
```java
clipboard = new Clipboard("Application Clipboard");
```

 and ([original](https://github.com/zxiaofan/JDK/blob/19a6c71e52f3ecd74e4a66be5d0d552ce7175531/jdk-11.0.2/src/java.base/java/util/Locale.java#L1021-L1102))
``java
SecurityManager sm = System.getSecurityManager();
if (sm != null) sm.checkPermission(new PropertyPermission("user.language", "write"));
```

Should be removed completely. (bearbeitet)

File: src/main/resources/META-INF/rewrite/java-version-25.yml [hunk 1]
```diff
@@ -0,0 +1,48 @@
...
+preconditions:
+  - org.openrewrite.java.migrate.search.FindJavaVersion:
+      version: 24
+recipeList:
+  - org.openrewrite.java.RemoveMethodInvocations:
+      methodPattern: java.security.AccessController#checkPermission(*)
+  - org.openrewrite.java.RemoveMethodInvocations:
+      methodPattern: import java.security.Policy#setPolicy(*)
+  - org.openrewrite.java.RemoveMethodInvocations:
+      methodPattern: import java.lang.SecurityManager#check*(*)
```
timtebeek, line 46: Method patterns do not allow for wildcards within parentheses; there you'll want to use `..`.
https://docs.openrewrite.org/reference/method-patterns
```suggestion
      methodPattern: java.security.AccessController#checkPermission(..)
  - org.openrewrite.java.RemoveMethodInvocations:
      methodPattern: import java.security.Policy#setPolicy(..)
  - org.openrewrite.java.RemoveMethodInvocations:
      methodPattern: import java.lang.SecurityManager#check*(..)
```

src/main/resources/META-INF/rewrite/java-version-25.yml
```diff
@@ -0,0 +1,48 @@
...
+preconditions:
+  - org.openrewrite.java.migrate.search.FindJavaVersion:
+      version: 24
+recipeList:
+  - org.openrewrite.java.RemoveMethodInvocations:
+      methodPattern: java.security.AccessController#checkPermission(..)
+  - org.openrewrite.java.RemoveMethodInvocations:
+      methodPattern: import java.security.Policy#setPolicy(..)
+  - org.openrewrite.java.RemoveMethodInvocations:
+      methodPattern: import java.lang.SecurityManager#check*(..)
```
timtebeek, line 46: These import statements are presumably coming from a copy paste error.
```suggestion
      methodPattern: java.security.Policy#setPolicy(..)
  - org.openrewrite.java.RemoveMethodInvocations:
      methodPattern: java.lang.SecurityManager#check*(..)
```

src/main/resources/META-INF/rewrite/java-version-25.yml
```diff
@@ -0,0 +1,48 @@
...
+      version: 24
+recipeList:
+  - org.openrewrite.java.RemoveMethodInvocations:
+      methodPattern: java.security.AccessController#checkPermission(..)
+  - org.openrewrite.java.RemoveMethodInvocations:
+      methodPattern: java.security.Policy#setPolicy(..)
+  - org.openrewrite.java.RemoveMethodInvocations:
+      methodPattern: java.lang.SecurityManager#check*(..)
+  - org.openrewrite.staticanalysis.UnnecessaryCatch
+  - org.openrewrite.staticanalysis.UnnecessaryThrows
```
timtebeek, line 48: Hmm; `UnnecessaryCatch` and `UnnecessaryThrows` will make changes anywhere, not just where it relates to the changes stemming from the removed methods. That might surprise folks that run a Java upgrade to see these changes, especially when they were not using anything related to `java.security` or `java.lang.SecurityManager`.  Perhaps we an add a precondition to `RemoveSecurityManagerAPI` at the very least, or move that into an imperative recipe that only runs the `Unnecessary` recipes when there are changes, and only on affected files.

timtebeek (changes requested):
Let's address the comment I posted.

src/main/resources/META-INF/rewrite/java-version-25.yml
MBoegers: Yes a precondition for the 3 types in use make sence

MBoegers: Does it make sence to make it 3 refcipes?

timtebeek:
I'm seeing [here](https://bugs.openjdk.org/browse/JDK-8338412)
> `getSecurityManager` used to return the system-wide Security Manager. It now always returns null.

So then can we do the simplest thing here and make that explicit? So going from
```java
SecurityManager security = System.getSecurityManager();
if (security != null) { ... }
```
To 
```java
SecurityManager security = null;
if (security != null) { ... }
```

That recipe should be very simple to write, and makes explicit what's now implicit. We do a very similar thing in rewrite-feature-flags where we replace the method invocation with a literal, and then [call out to other visitors to clean up](https://github.com/openrewrite/rewrite-feature-flags/blob/c448e2a4df5e8f42a60a172f3ebab828644609d5/src/main/java/org/openrewrite/featureflags/RemoveStringFlag.java#L111-L115). Removing that conditional should then happen by default when the variable is always `null`. Possibly best tackled as a separate PR from the other changes here.

MBoegers:
Yes, a decent option here. I will submit a second PR dealing with Securtity Manager. `System.getSecurtityManager()` will move to null and then apply existing cleanup recipes.

src/main/resources/META-INF/rewrite/java-version-25.yml
```diff
@@ -0,0 +1,105 @@
...
+name: org.openrewrite.java.migrate.AccessController
+displayName: Remove Security AccessController
+description: The Security Manager API is unsupported in Java 24. This recipe will remove the usage of `java.security.AccessController`.
+tags:
+  - java25
+  - security
+  - deprecation
+preconditions:
+  - org.openrewrite.java.migrate.search.FindJavaVersion:
+      version: 24
```
timtebeek, line 58: This is already checked above in `RemoveSecurityManagerAPI` right? Then we can either remove it here and below, or remove it above and potentially not have an intermediate `RemoveSecurityManagerAPI` at all.
```suggestion
```

src/main/resources/META-INF/rewrite/java-version-25.yml
```diff
@@ -0,0 +1,105 @@
...
+displayName: Remove Security Manager API
+description: The Security Manager API is unsupported in Java 24. This recipe will remove the API and any usages of it.
+tags:
+  - java25
+  - security
+  - deprecation
+preconditions:
+  - org.openrewrite.java.migrate.search.FindJavaVersion:
+      version: 24
+recipeList:
```
timtebeek, line 42: Alternative to the suggestion below; pick this one or the other one, but not both.
```suggestion
```

MBoegers: The Leaf recipes (`RemoveSecurityPolicy`, `RemoveSecurityManager` and `AccessController`) should guard that they are only executed against Java 24 as these changes are only needed upgrading to Java 24. Check on `RemoveSecurityManagerAPI` is not needed, agree. And `RemoveSecurityManagerAPI` is two hands full of clicks in Recipe Builder.

src/main/resources/META-INF/rewrite/java-version-25.yml
MBoegers: removed  `RemoveSecurityManagerAPI`

src/main/resources/META-INF/rewrite/java-version-25.yml
```diff
@@ -0,0 +1,90 @@
...
+description: The Security Manager API is unsupported in Java 24. This recipe will remove the use of `java.security.Policy`.
+tags:
+  - java25
+  - security
+  - deprecation
+preconditions:
+  - org.openrewrite.java.migrate.search.FindJavaVersion:
+      version: 24
+  - org.openrewrite.java.search.FindTypes:
+      fullyQualifiedTypeName: java.security.Policy
```
timtebeek, line 65: Might as well be most selective through find methods here, right? Just to avoid any accidental changes.
```suggestion
  - org.openrewrite.java.search.FindMethods:
      methodPattern: java.security.Policy#setPolicy(..)
```
Same can be applied below.

timtebeek (approved):
Thanks!

PR #714, opened 2025-05-01
Pattern Matching for switch in Java 21 / JEP 411

### LINKED ISSUE #306: Pattern Matching for switch in Java 21 / JEP 411
Add recipe for [JEP 441: Pattern Matching for switch](https://openjdk.org/jeps/441) to improve Java 21 migration (and possibly also for the related [JEP 440: Record Patterns](https://openjdk.org/jeps/440)).

The two examples copied from the [JEP 441: Pattern Matching for switch](https://openjdk.org/jeps/441) project (there are more):

**instanceof**
```java
// Prior to Java 21
    String formatted = "unknown";
    if (obj instanceof Integer i) {
        formatted = String.format("int %d", i);
        formatted = String.format("long %d", l);
    } else if (obj instanceof Double d) {
        formatted = String.format("double %f", d);
    } else if (obj instanceof String s) {
        formatted = String.format("String %s", s);
    }
    return formatted;
}
```

```java
// As of Java 21
static String formatterPatternSwitch(Object obj) {
    return switch (obj) {
        case Long l    -> String.format("long %d", l);
        case Double d  -> String.format("double %f", d);
        case String s  -> String.format("String %s", s);
        default        -> obj.toString();
    };
}
```

**Switches and null**
```java
// Prior to Java 21
static void testFooBarOld(String s) {
    if (s == null) {
        System.out.println("Oops!");
        return;
    }
    switch (s) {
        case "Foo", "Bar" -> System.out.println("Great");
        default           -> System.out.println("Ok");
    }
}
```

```java
// As of Java 21
static void testFooBarNew(String s) {
    switch (s) {
        case null         -> System.out.println("Oops");
        case "Foo", "Bar" -> System.out.println("Great");
        default           -> System.out.println("Ok");
    }
}
```

no

## Additional context
- #305

Not at the moment

Jenson3210:
## What's changed?
Added 5 recipes (+ added to the java 21 migration recipe):
- NullCheckAsSwitchCase
- IfElseIfConstructToSwitch
- RefineSwitchCases
- SwtichCaseEnumGuardtoLabel
- SwitchPatternMatching (combining the other 4 in a logical order to get most benefit from them)

## What's your motivation?
- Closes #306

### Checklist


src/test/java/org/openrewrite/java/migrate/lang/IfElseIfConstructToSwitchTest.java
```diff
@@ -0,0 +1,347 @@
...
+    }
+
+    @Test
+    @DocumentExample
+    void defaultSwitchBlockWithNullCheckAndFinalElseStatement() {
+        rewriteRun(
+          java(
+            """
+              package org.openrewrite.java.migrate.lang;
+
```
timtebeek, line 39: Ideally leave out any package declarations unless absolutely necessary; typically you can leave these out for most tests.

src/test/java/org/openrewrite/java/migrate/lang/IfElseIfConstructToSwitchTest.java
```diff
@@ -0,0 +1,347 @@
...
+                          case null -> formatted = "null";
+                          case Integer i -> formatted = String.format("int %d", i);
+                          case Long l -> formatted = String.format("long %d", l);
+                          case Double d -> formatted = String.format("double %f", d);
+                          case String s -> {
+                              String str = "String";
+                              formatted = String.format("%s %s", str, s);
+                          }
+                          default -> formatted = "unknown";
+                      }
```
timtebeek, line 79: This made me think of another recipe we could write: when every case body is an assignment to the same variable, and there is a default case, then we can turn the whole thing into a switch expression. Would you mind logging that as an issue?

Jenson3210: created

src/test/java/org/openrewrite/java/migrate/lang/IfElseIfConstructToSwitchTest.java
Jenson3210: removed all package declarations

src/main/java/org/openrewrite/java/migrate/lang/IfElseIfConstructToSwitch.java
```diff
@@ -0,0 +1,211 @@
...
+                SwitchCandidate switchCandidate = new SwitchCandidate(iff, getCursor());
+
+                if (switchCandidate.isValidCandidate()) {
+                    Object[] arguments = switchCandidate.buildTemplateArguments(getCursor());
+                    String switchBody = switchCandidate.buildTemplate();
+                    JavaTemplate switchTemplate = JavaTemplate.builder(switchBody).javaParser(JavaParser.fromJavaVersion().classpathFromResources(ctx)).contextSensitive().build();
+                    return super.visitSwitch(switchTemplate.apply(getCursor(), iff.getCoordinates().replace(), arguments).withPrefix(iff.getPrefix()), ctx);
+                } else {
+                    return super.visitIf(iff, ctx);
+                }
```
jevanlingen, line 72:
```suggestion
                }
                return super.visitIf(iff, ctx);
```

src/main/java/org/openrewrite/java/migrate/lang/IfElseIfConstructToSwitch.java
```diff
@@ -0,0 +1,211 @@
...
+
+            // we need at least 3 cases to use a switch
+            if (potentialCandidate && patternMatchers.keySet().size() + (nullCaseInSwitch ? 1 : 0) + (hasLastElseBlock ? 1 : 0) <= 2) {
+                this.potentialCandidate = false;
+            }
+            return potentialCandidate;
+        }
+
+        String switchOn() {
+            return ((J.Identifier) patternMatchers.keySet().stream().map(J.InstanceOf::getExpression).findAny().get()).getSimpleName();
```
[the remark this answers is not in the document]
Jenson3210: yes, as otherwise there would it would not be a candidate for replacement. 
We went over all if conditions and all of them are an instanceof check. If we encountered something else in the if, we do return false earlier on and we do not touch the iff block

src/test/java/org/openrewrite/java/migrate/lang/NullCheckAsSwitchCaseTest.java
```diff
@@ -0,0 +1,125 @@
...
+                          case "C" -> formatted = "Good";
+                          case "D" -> formatted = "Hmmm...";
+                          default -> formatted = "unknown";
+                      }
+                      return formatted;
+                  }
+              }
+              """,
+            """
+            class Test {
```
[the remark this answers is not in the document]
Jenson3210: There's another issue indeed for this created already, but well spotted. 
To me that's a different recipe completely (last one in the chain)
`SwitchStatementsAsExpressions` or something alike

See also

src/test/java/org/openrewrite/java/migrate/lang/IfElseIfConstructToSwitchTest.java
```diff
@@ -0,0 +1,320 @@
...
+package org.openrewrite.java.migrate.lang;
+
+import org.junit.jupiter.api.Test;
+import org.openrewrite.DocumentExample;
+import org.openrewrite.test.RecipeSpec;
+import org.openrewrite.test.RewriteTest;
+
+import static org.openrewrite.java.Assertions.java;
+
+class IfElseIfConstructToSwitchTest implements RewriteTest {
```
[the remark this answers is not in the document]
Jenson3210: Great! thanks, added!

src/main/java/org/openrewrite/java/migrate/lang/IfElseIfConstructToSwitch.java
```diff
@@ -0,0 +1,211 @@
...
+
+        SwitchCandidate(J.If iff, Cursor cursor) {
+            J.If ifPart = iff;
+            while (potentialCandidate && ifPart != null) {
+                if (ifPart.getIfCondition().getTree() instanceof J.Binary) {
+                    ifPart = handleNullCheck(ifPart, cursor);
+                } else if (ifPart.getIfCondition().getTree() instanceof J.InstanceOf) {
+                    ifPart = handleInstanceOfCheck(ifPart);
+                } else {
+                    noPotentialCandidate();
```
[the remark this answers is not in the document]
Jenson3210: that variable gets set from 2 locations, so i thought it was "stronger" to have it named with what it does.

JohannisK (approved):

src/main/java/org/openrewrite/java/migrate/lang/NullCheck.java
```diff
@@ -0,0 +1,97 @@
...
+            if (cursor.getValue() instanceof J.If) {
+                J.If iff = cursor.getValue();
+                if (iff.getIfCondition().getTree() instanceof J.Binary) {
+                    J.Binary binary = (J.Binary) iff.getIfCondition().getTree();
+                    if (J.Binary.Type.Equal == binary.getOperator()) {
+                        if (binary.getLeft() instanceof J.Literal && ((J.Literal) binary.getLeft()).getValue() == null && binary.getRight() instanceof J.Identifier) {
+                            return new NullCheck(cursor, (J.Identifier) binary.getRight());
+                        } else if (binary.getRight() instanceof J.Literal && ((J.Literal) binary.getRight()).getValue() == null && binary.getLeft() instanceof J.Identifier) {
+                            return new NullCheck(cursor, (J.Identifier) binary.getLeft());
+                        }
```
timtebeek, line 90: Minor change pending upstream to make this easier:
- https://github.com/openrewrite/rewrite/pull/5523

src/test/java/org/openrewrite/java/migrate/lang/RefineSwitchCasesTest.java
```diff
@@ -0,0 +1,160 @@
...
+              }
+              """,
+            """
+              class Test {
+                  static void score(Object obj) {
+                      switch (obj) {
+                          case Integer i when i >= 5 && i <= 10 -> System.out.println("You got it");
+                          case Integer i when i >= 0 && i < 5 -> System.out.println("Shame");
+                          case Integer i -> {
+                          }
```
timtebeek, line 152: This one jumped out at me as a bit odd; would it make sense to not add these, or have a `doAfterVisitor` that clears those out? Or am I misunderstanding the need for these?

src/main/resources/META-INF/rewrite/java-version-21.yml
```diff
@@ -130,3 +131,15 @@ recipeList:
       methodPattern: java.awt.image.ColorModel finalize()
   - org.openrewrite.java.RemoveMethodInvocations:
       methodPattern: java.awt.image.IndexColorModel finalize()
+---
+type: specs.openrewrite.org/v1beta/recipe
+name: org.openrewrite.java.migrate.SwitchPatternMatching
+displayName: Adopt switch pattern matching (JEP 441)
+description: "[JEP 441](https://openjdk.org/jeps/441) describes some improvements on how switch statements can be improved with better pattern matching."
```
timtebeek, line 138:
```suggestion
description: "[JEP 441](https://openjdk.org/jeps/441) describes how some switch statements can be improved with pattern matching."
```

src/main/java/org/openrewrite/java/migrate/lang/NullCheck.java
```diff
@@ -0,0 +1,100 @@
...
+        } else if (statement instanceof J.Assignment) {
+            return assigns((J.Assignment) statement, variable);
+        }
+        return false;
+    }
+
+    private static boolean assigns(J.Assignment assignment, J.Identifier variable) {
+        return assignment.getVariable() instanceof J.Identifier &&
+                // TODO Slight worry here about say A.field vs. B.field, when only comparing the simple name
+                ((J.Identifier) assignment.getVariable()).getSimpleName().equals(variable.getSimpleName());
```
timtebeek, line 73: I saw this pattern in a couple places: comparing based on `simpleName` only. There's a bit of a risk here that I'm not sure we adequately handle in all cases, in cases like `A.field` vs `B.field`. Would be good to double check.

timtebeek (commented):
Overall I have to say these have been great to see! Sorry it took a while to get to a review; I'm looking forward to seeing these in practice. Did you already run them through the CLI against some larger set of LSTs?

I've left a few comments that I think would be good to go over before we merge.

src/main/java/org/openrewrite/java/migrate/lang/IfElseIfConstructToSwitch.java
```diff
@@ -221,5 +219,48 @@ private String getStatementArgument(Statement statement, Cursor cursor) {
...
+                if (select.isEmpty()) {
+                    select = "this";
+                }
+                return select + "$" + ((J.MethodInvocation) expression).getSimpleName();
+            }
+            return expression.toString();
+        }
+    }
+
+    private static <T extends Expression> boolean semanticallySame(@Nullable T e1, @Nullable T e2) {
```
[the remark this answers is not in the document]
Jenson3210: could we add the null checks there? 
like null == null, nonNull != null and null != nonNull?

src/main/java/org/openrewrite/java/migrate/lang/NullCheck.java
```diff
@@ -0,0 +1,163 @@
...
+                }
+                return couldModifyNullCheckedValue(((J.MethodInvocation) assigns).getSelect(), ((J.MethodInvocation) nullChecked).getSelect());
+            }
+        }
+        if (assigns instanceof J.Assignment) {
+            return couldModifyNullCheckedValue(((J.Assignment) assigns).getVariable(), nullChecked);
+        }
+
+        return true;
+    }
```
timtebeek, line 137: I've been trying for a bit if a visitor based approach might be clearer. It's hard to get to the 100% guarantee though.
```suggestion
    public boolean couldModifyNullCheckedValue() {
        Statement statement = whenNull();
        if (statement instanceof J.Block || statement instanceof Expression) {
            return couldModifyNullCheckedValue(statement, nullCheckedParameter);
        }
        // Cautious by default
        return true;
    }

    private static boolean couldModifyNullCheckedValue(@Nullable J expression, Expression nullChecked) {
        if (nullChecked instanceof J.FieldAccess && couldModifyNullCheckedValue(expression, ((J.FieldAccess) nullChecked).getTarget())) {
            return true;
        }
        if (nullChecked instanceof J.MethodInvocation &&
                ((J.MethodInvocation) nullChecked).getSelect() != null &&
                couldModifyNullCheckedValue(expression, ((J.MethodInvocation) nullChecked).getSelect())) {
            return true;
        }
        return new JavaIsoVisitor<AtomicBoolean>() {
            @Override
            public J.Assignment visitAssignment(J.Assignment assignment, AtomicBoolean atomicBoolean) {
                J.Assignment as = super.visitAssignment(assignment, atomicBoolean);
                if (SemanticallyEqual.areEqual(as.getVariable(), nullChecked)) {
                    atomicBoolean.set(true);
                }
                return as;
            }

            @Override
            public J.FieldAccess visitFieldAccess(J.FieldAccess fieldAccess, AtomicBoolean atomicBoolean) {
                J.FieldAccess fa = super.visitFieldAccess(fieldAccess, atomicBoolean);
                if (SemanticallyEqual.areEqual(fa, nullChecked) ||
                        SemanticallyEqual.areEqual(fa.getTarget(), nullChecked)) {
                    atomicBoolean.set(true);
                }
                return fa;
            }

            @Override
            public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, AtomicBoolean atomicBoolean) {
                J.MethodInvocation mi = super.visitMethodInvocation(method, atomicBoolean);
                if (SemanticallyEqual.areEqual(mi, nullChecked) ||
                        ((mi.getSelect() != null) && SemanticallyEqual.areEqual(mi.getSelect(), nullChecked))) {
                    atomicBoolean.set(true);
                }
                return mi;
            }
        }.reduce(expression, new AtomicBoolean(false)).get();
    }
```

src/main/java/org/openrewrite/java/migrate/lang/IfElseIfConstructToSwitch.java
```diff
@@ -0,0 +1,221 @@
...
+        return Preconditions.check(Preconditions.not(new KotlinFileChecker<>()), new JavaVisitor<ExecutionContext>() {
+
+            @Override
+            public J visitIf(J.If iff, ExecutionContext ctx) {
+                SwitchCandidate switchCandidate = new SwitchCandidate(iff, getCursor());
+
+                if (switchCandidate.isValidCandidate()) {
+                    Object[] arguments = switchCandidate.buildTemplateArguments(getCursor());
+                    if (arguments.length == 0) {
+                        return iff;
```
jevanlingen: Are you sure you should just return the `iff` variable here? I would expected:

```suggestion
                        return super.visitIf(iff, ctx);
```

src/main/java/org/openrewrite/java/migrate/lang/IfElseIfConstructToSwitch.java
```diff
@@ -0,0 +1,221 @@
...
+            } else {
+                elze = elsePart != null ? elsePart.getBody() : null;
+                ifPart = null;
+            }
+            return ifPart;
+        }
+
+        void noPotentialCandidate() {
+            this.potentialCandidate = false;
+        }
```
jevanlingen: I think I would remove this method and just do

```java
potentialCandidate = false;
```

everywhere. I think that's clearer in the end, as it is on par with the rest of your code. For instance, check the `handleNullCheck` method. There you also use the `nullCheckedStatement` variable directly.

src/main/java/org/openrewrite/java/migrate/lang/IfElseIfConstructToSwitch.java
```diff
@@ -0,0 +1,221 @@
...
+                return super.visitIf(iff, ctx);
+            }
+        });
+    }
+
+    private static class SwitchCandidate {
+        private final Map<J.InstanceOf, Statement> patternMatchers = new LinkedHashMap<>();
+        private @Nullable Expression nullCheckedParameter = null;
+        private @Nullable Statement nullCheckedStatement = null;
+        private @Nullable Statement elze = null;
```
jevanlingen:
```suggestion
        private @Nullable Statement else_ = null;
```

src/main/java/org/openrewrite/java/migrate/lang/IfElseIfConstructToSwitch.java
```diff
@@ -0,0 +1,221 @@
...
+            this.potentialCandidate = false;
+        }
+
+        boolean isValidCandidate() {
+            Expression switchOn = switchOn();
+            // all ifs in the chain must be on the same variable in order to be a candidate for switch pattern matching
+            if (switchOn == null || potentialCandidate && !patternMatchers.keySet().stream()
+                    .map(J.InstanceOf::getExpression)
+                    .allMatch(it -> SemanticallyEqual.areEqual(switchOn, it))) {
+                this.potentialCandidate = false;
```
jevanlingen:
```suggestion
                potentialCandidate = false;
```

Or vice versa, use the `this.` prefix everywhere. At least, be consistent 😉

src/main/java/org/openrewrite/java/migrate/lang/IfElseIfConstructToSwitch.java
```diff
@@ -0,0 +1,221 @@
...
+                ifPart = null;
+            }
+            return ifPart;
+        }
+
+        void noPotentialCandidate() {
+            this.potentialCandidate = false;
+        }
+
+        boolean isValidCandidate() {
```
jevanlingen:
```suggestion
        boolean validate() {
```
This method does not "just" return whether it is a valid candidate, but also calculates it. So it contains side effects. You could debate this could be written in a more functional style, but let's not get there. But at least show what your function does (incl side effects) by giving it a proper name 💪🏻!

src/main/java/org/openrewrite/java/migrate/lang/IfElseIfConstructToSwitch.java
```diff
@@ -0,0 +1,221 @@
...
+            }
+            return potentialCandidate;
+        }
+
+        @Nullable Expression switchOn() {
+            return patternMatchers.keySet().stream()
+                    .map(J.InstanceOf::getExpression)
+                    .findAny()
+                    .orElse(null);
+        }
```
jevanlingen:
```suggestion
        Optional<Expression> switchOn() {
            return patternMatchers.keySet().stream()
                    .map(J.InstanceOf::getExpression)
                    .findAny();
        }
```

To be honest, I like that the null pointer starts behaving like an optional with JSpecify (just like Kotlin's null type system). Still I wouldn't go this far, as there is already an Optional reference in the heap anyways. Just stay in the Optional space, so you can use the `map().orElse` in the `buildTemplateArguments` method.

src/main/java/org/openrewrite/java/migrate/lang/NullCheck.java
```diff
@@ -0,0 +1,163 @@
...
+public class NullCheck implements Trait<J.If> {
+    Cursor cursor;
+    Expression nullCheckedParameter;
+
+    public Statement whenNull() {
+        return getTree().getThenPart();
+    }
+
+    public @Nullable Statement whenNotNull() {
+        J.If.Else anElse = getTree().getElsePart();
```
jevanlingen, line 42:
```suggestion
        J.If.Else else_ = getTree().getElsePart();
```

I think that's clearer; it also follows the JDK kind of code standard: https://docs.oracle.com/en/java/javase/23/docs/api/java.base/java/lang/classfile/CodeBuilder.html#return_().

src/main/java/org/openrewrite/java/migrate/lang/NullCheck.java
jevanlingen: I don't know if you can get it right, but I think the visitor pattern would be more easy to read here (instead by going for the recursive route). So I would give it a shot @Jenson3210!

Jenson3210: just replacing it works, but it missed one case as I had a missing test case (Identifier used sending it to some method that might update the variable behind that identifier if it is not a immutableVariable). 
As a rookie, I do not like these inline visitors that much **yet**. But that will come and be more me than the code, so adapted to visitor and fixed that case.

src/test/java/org/openrewrite/java/migrate/lang/RefineSwitchCasesTest.java
Jenson3210: Going to readd this as I recalled the reason for this: 

```java
// Primitive Pattern Matching for Switch
String format ="NULL";
Object test = "A";

switch (test) {
    case String s -> {
      if (s.equalsIgnoreCase("B")) {
        format = "B";
      }
    }
    default -> format = "A";
}

System.out.println(format);
```
prints "NULL"

so does: 
```java
// Primitive Pattern Matching for Switch
String format ="NULL";
Object test = "A";

switch (test) {
    case String s when s.equalsIgnoreCase("B") -> {
      format = "B";
    }
    case String s -> {}
    default -> format = "A";
}

System.out.println(format);
```

But not okay:
```java
// Primitive Pattern Matching for Switch
String format ="NULL";
Object test = "A";

switch (test) {
    case String s when s.equalsIgnoreCase("B") -> {
      format = "B";
    }
    default -> format = "A";
}

System.out.println(format);
```

PR #727, opened 2025-05-16
New recipe to only add JAXB API dependencies without runtime dependencies

rlsanders4:
## What's changed?
A new recipe was added to update the JAXB API dependency when migrating to Jakarta EE 9 and remove any JAXB runtime dependencies. 

## What's your motivation?
This is necessary when migrating to an app server where the JAXB implementation is provided by the runtime.

@timtebeek

### Checklist


src/main/resources/META-INF/rewrite/java-version-11.yml
```diff
@@ -180,6 +180,74 @@ recipeList:
       newVersion: 2.5.x
 ---
 type: specs.openrewrite.org/v1beta/recipe
+name: org.openrewrite.java.migrate.javax.AddJaxbAPIDependencies
```
timtebeek, line 183: Should we reference this new `org.openrewrite.java.migrate.javax.AddJaxbAPIDependencies` from anywhere?
Without that folks would need to seek it out and run it specifically, whereas perhaps it needs to hook into Java 11 upgrade?

rlsanders4: The current Java 11 upgrade list adds some jakarta runtime dependencies to the application. That makes sense for some use cases, but it will cause errors for users migrating to an app server which already provides these runtime dependencies. Currently, we are calling some the Java 11 recipes piecemeal in order to support that managed scenario instead of using the full Java 11 list.

For this recipe, we will use it instead of the existing `AddJaxbDependencies` which is already on the Java 11 list.

One potential long term solution is to break this in into a few lists:

- A list for users migrating to a managed environment
- A list for users migrating to an unmanaged environment (this could remain the default suggestion)
- A list for common recipes to be referenced by both

This would add some additional complexity though.

evie-lau (approved):

evie-lau (changes requested):
Could we move the common recipe components into another group, and have the runtime/no-runtime recipe groups call that along with their own runtime addition/removal recipes?

timtebeek (approved):
Thanks both for iterating on the right approach here!

PR #729, opened 2025-05-16
Upgrade existing jakarta.annotation-api to 2.0.x for EE9

evie-lau:
## What's changed?

In the Jakarta EE 9 recipe list, add an extra recipe step to ensure `jakarta.annotation-api` is upgraded to the correct version for EE9 in case an older version was used.

#### QUESTION: Is it even a valid case to run `AddCommonAnnotationsDependencies` first, before running `JavaxMigrationToJakarta` (Jakarta EE 9 migration)? Should the EE 9 migration supercede the former recipe?

If [`AddCommonAnnotationsDependencies`](https://github.com/openrewrite/rewrite-migrate-java/blob/main/src/main/resources/META-INF/rewrite/add-common-annotations-dependencies.yml) is run beforehand, it updates the original `javax.annotation-api` to use `jakarta.annotation-api:1.3.x`. The EE9 recipe then fails to upgrade to `2.0.x` because the original `javax.annotation-api` no longer exists.

When this recipe runs afterwards, it should update the version if the jakarta version of the artifact already existed.

### Checklist

### DIFF
--- a/src/main/resources/META-INF/rewrite/jakarta-ee-9.yml
+++ b/src/main/resources/META-INF/rewrite/jakarta-ee-9.yml
@@ -112,6 +112,10 @@ recipeList:
       newGroupId: jakarta.annotation
       newArtifactId: jakarta.annotation-api
       newVersion: 2.0.x
+  - org.openrewrite.java.dependencies.UpgradeDependencyVersion:
+      groupId: jakarta.annotation
+      artifactId: jakarta.annotation-api
+      newVersion: 2.0.x
   - org.openrewrite.java.ChangePackage:
       oldPackageName: javax.annotation
       newPackageName: jakarta.annotation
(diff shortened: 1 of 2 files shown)


evie-lau:
I noticed this fails this test
https://github.com/openrewrite/rewrite-migrate-java/blob/5adb91d9e1edf16e0ccb67837169c25a33b01860/src/test/java/org/openrewrite/java/migrate/jakarta/JavaxToJakartaTest.java#L577-L613

I'm trying to figure out if this change makes sense, or whether the tests need to be updated.

timtebeek (approved):

evie-lau:
I think this might be good to go. It's also now in line with what some of the other EE9 recipes do, where they all run `ChangeDependency` followed by `UpgradeDependencyVersion` to the same version specified.

Issue #733, opened 2025-05-07
Move statements before `super(..)` in Java constructor


JohannisK:
## What problem are you trying to solve?
[Jep 513](https://openjdk.org/jeps/513) calls for allowing statements before the call to `super()` in Java constructors, allowing for simple assignments and validations before doing expensive construction.

## What precondition(s) should be checked before applying this recipe?
Java version >= 25
(Also double check Jep 513 lands in Java 25)

## Describe the situation before applying the recipe

```java
class A {
    public A(String bar) {
        super();
        if(bar.equals("test"))
            throw new RuntimeException();
    }
}
```

## Describe the situation after applying the recipe

```java
class A {
    public A(String bar) {
        if(bar.equals("test"))
            throw new RuntimeException();
        super();
    }
}
```

## Any additional context
https://openjdk.org/jeps/513 has several examples and considerations on what statements should be supported before calls to `super(..)`

MohamedFazil1406:
### Before applying the Recipe

The constructor calls super() first and performs validation afterwards. This follows the old Java restriction that required super() to be the first statement in a constructor.

### After applying the recipe

The recipe moves safe validation statements from after super() to before super() in constructors, leveraging JEP 513's flexible constructor bodies feature in Java 25. This allows invalid arguments to be detected before superclass construction occurs.

timtebeek:
- We've handled the cases that are safe to do in https://github.com/openrewrite/rewrite-migrate-java/pull/1135
- Reordering statements _after_ super/this has the potential to do harm so we'll not automate those changes for now.

Thanks all!

PR #736, opened 2025-05-19
Move statements before super(..) in Java constructor #733

rkampani:
## What's changed?
Relates to [#5397 ](https://github.com/openrewrite/rewrite-migrate-java/issues/733)
[Jep 513](https://openjdk.org/jeps/513) calls for allowing statements before the call to super() in Java constructors, allowing for simple assignments and validations before doing expensive construction.

Java version >= 25
(Also double check Jep 513 lands in Java 25)

**Before Applying Recipe**
```
class A {
    public A(String bar) {
        super();
        if(bar.equals("test"))
            throw new RuntimeException();
    }
}

```
**After Applying Recipe**

```
class A {
    public A(String bar) {
        if(bar.equals("test"))
            throw new RuntimeException();
        super();
    }
}
```
### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/util/RelocateSuperCall.java
+++ b/src/main/java/org/openrewrite/java/migrate/util/RelocateSuperCall.java
@@ -1,0 +1,68 @@
+package org.openrewrite.java.migrate.util;
+
+
+import lombok.EqualsAndHashCode;
+import lombok.Value;
+import org.openrewrite.ExecutionContext;
+import org.openrewrite.Preconditions;
+import org.openrewrite.Recipe;
+import org.openrewrite.TreeVisitor;
+import org.openrewrite.java.JavaIsoVisitor;
+import org.openrewrite.java.search.UsesJavaVersion;
+import org.openrewrite.java.tree.J;
+import org.openrewrite.java.tree.Statement;
+
+import java.util.List;
+
+@EqualsAndHashCode(callSuper = false)
+@Value
+public class RelocateSuperCall extends Recipe {
+
+    @Override
+    public String getDisplayName() {
+        return "Move `super()` after conditionals (Java 25+)";
+    }
+
+    @Override
+    public String getDescription() {
+        return "Relocates `super()` calls to take advantage of the early construction context introduced by JEP 513 in Java 25+, allowing statements before constructor calls.";
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return Preconditions.check(
+                new UsesJavaVersion<>(25),
+                new RelocateSuperCallVisitor());
+    }
+
+    private static class RelocateSuperCallVisitor extends JavaIsoVisitor<ExecutionContext> {
+
+        @Override
+        public J.MethodDeclaration visitMethodDeclaration(J.MethodDeclaration method, ExecutionContext ctx) {
+            if (!method.isConstructor() || method.getBody() == null) {
+                return method;
+            }
+
+            List<Statement> statements = method.getBody().getStatements();
+            if (statements.size() < 2) {
+                return method;
+            }
+
+            Statement first = statements.get(0);
+            if (!(first instanceof J.MethodInvocation)) {
+                return method;
+            }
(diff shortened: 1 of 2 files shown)


src/main/java/org/openrewrite/java/migrate/util/RelocateSuperCall.java
```diff
@@ -0,0 +1,68 @@
...
+import org.openrewrite.java.JavaIsoVisitor;
+import org.openrewrite.java.search.UsesJavaVersion;
+import org.openrewrite.java.tree.J;
+import org.openrewrite.java.tree.Statement;
+
+import java.util.List;
+
+@EqualsAndHashCode(callSuper = false)
+@Value
+public class RelocateSuperCall extends Recipe {
```
timtebeek, line 19: Might make sense to reference this immediately in the Java 25 upgrade recipe:
https://github.com/openrewrite/rewrite-migrate-java/blob/cb161e6fb27a5b5c1a5374aa750fe21baefa9f5f/src/main/resources/META-INF/rewrite/java-version-25.yml#L18-L32

src/main/java/org/openrewrite/java/migrate/RelocateSuperCall.java
```diff
@@ -0,0 +1,83 @@
...
+            if (!(first instanceof J.MethodInvocation)) {
+                return method;
+            }
+            J.MethodInvocation methodInvocation = (J.MethodInvocation) first;
+            if (!"super".equals(methodInvocation.getSimpleName())) {
+                return method;
+            }
+
+            // Move super() to the end
+            List<Statement> updated = new java.util.ArrayList<>(statements);
```
MBoegers, line 76: ATM you move the `super()` call to the very end of a constructor call. If i recall correctly there are limitations regarding the reference to instance fields. Could you please revisit [JEP 513](https://openjdk.org/jeps/513#Early-construction-contexts)s section about _Early-construction-contexts_ and verify that we are producing JLS compliant code?

MBoegers (changes requested):

MBoegers:
Thanks for your contribution, I love seeing new Language features been picked up :) I think 1 or 2 tests should be enough to make sure we are JLS compliant in the context of the Early construction contexts.

timtebeek:
We've just merged the Java 25 parser support, so we can have another look here now. Would you want to continue this work @rkampani , or would you like us to take over?
- https://github.com/openrewrite/rewrite/pull/5985

src/main/java/org/openrewrite/java/migrate/RelocateSuperCall.java
rkampani: updated , pls re-review !

rkampani:
https://github.com/openrewrite/rewrite-migrate-java/pull/736#issuecomment-3244881477
Thanks for the reminder — I had completely forgotten about it. I've just pushed the changes. Could you please re-review when you get a chance

src/test/java/org/openrewrite/java/migrate/RelocateSuperCallTest.java
```diff
@@ -0,0 +1,207 @@
...
+                    if (age < 0)
+                        throw new IllegalArgumentException("Age cannot be negative");
+                    this.age = age;
+                }
+            }
+            class Employee extends Person {
+                Employee(int age) {
+                    if (age < 18 || age > 67)
+                        throw new IllegalArgumentException("Age must be between 18 and 67");
+                    super(age);
```
tfeuerba, line 70: Isn't this changing semantics? Supplying -4 will now raise "Age must be between 18 and 67" instead of "Age cannot be negative".

rkampani:
> We've just merged the Java 25 parser support, so we can have another look here now. Would you want to continue this work @rkampani , or would you like us to take over?
> 
> * [Init of Java 25 parser rewrite#5985](https://github.com/openrewrite/rewrite/pull/5985)

Hi @timtebeek : Yeah, if you can take it over the work, that’d be great — I won’t be able to wrap it up this week. Since Java 25 is already out, we can probably add this feature as needed.

PR #739, opened 2025-05-22
Add Java 21 migration to Java 25

MBoegers:
### Checklist


timtebeek (approved):
Thanks for the quick fix!

PR #742, opened 2025-05-23
Create Faces 3.0 migration

evie-lau:
!! WORK IN PROGRESS
Most of the migrations currently in the Faces 4 migration should actually happen during the EE9 Faces 3 migration.

## What's changed?
Created new Faces 3.0 migration, moved the javax to jakarta changes from Faces 4 (EE10) into here.

### Checklist

### DIFF
--- a/src/main/resources/META-INF/rewrite/jakarta-ee-9.yml
+++ b/src/main/resources/META-INF/rewrite/jakarta-ee-9.yml
@@ -333,29 +333,6 @@ recipeList:
       recursive: true
 ---
 type: specs.openrewrite.org/v1beta/recipe
-name: org.openrewrite.java.migrate.jakarta.JavaxFacesToJakartaFaces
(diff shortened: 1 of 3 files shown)


melloware:
I am almost thinking we dump the EE9 recipe as no one should really be migrating to 9.  10 is the official release

evie-lau:
I feel like since EE9 is still a valid released spec, it should still be allowed as a migration target, and should be fixed to migrate correctly.

melloware:
I will leave it up to Moderne but my peace has been said that EE9 is not really a release anyone should be migrating to.

timtebeek (approved):
Thanks both! Very helpful context to understanding the changes here, and the wider context about EE9 vs EE10. I think we're fine to keep the EE9 recipe around, even if the goal ought to be EE10. Helps keep the recipes organized, and ensures folks that already used some of the existing recipes can continue to reference those.

Issue #718, opened 2025-05-09
Convert assigning Switch statements to Switch expressions


Jenson3210:
When every case body is an assignment to the same variable, and there is a default case, then we can turn the whole thing into a switch expression in java 21.

**BEFORE**

```java
        String formatted = "initialValue";
        switch (obj) {
            case null -> formatted = "null";
            case Integer i -> formatted = String.format("int %d", i);
            case Long l -> formatted = String.format("long %d", l);
            case Double d -> formatted = String.format("double %f", d);
            case String s -> {
                String str = "String";
                formatted = String.format("%s %s", str, s);
            }
            default -> formatted = "unknown";
        }
```

**AFTER**

```java
        String formatted = switch (obj) {
            case null -> "null";
            case Integer i -> String.format("int %d", i);
            case Long l -> String.format("long %d", l);
            case Double d -> String.format("double %f", d);
            case String s -> {
                String str = "String";
                yield String.format("%s %s", str, s);
            }
            default -> "unknown";
        }
```

PR #747, opened 2025-06-02
If Else-If construct to switch.

### LINKED ISSUE #306: Pattern Matching for switch in Java 21 / JEP 411

Jenson3210:
## What's changed?
Cutting up the task at hand in multiple smaller PR's for better insights, risk assessment... 
- IfElseIfConstructToSwitch

## What's your motivation?
- Fixes #306

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/lang/IfElseIfConstructToSwitch.java
+++ b/src/main/java/org/openrewrite/java/migrate/lang/IfElseIfConstructToSwitch.java
@@ -60,26 +60,27 @@ public class IfElseIfConstructToSwitch extends Recipe {
                 Preconditions.not(new KotlinFileChecker<>()),
                 Preconditions.not(new GroovyFileChecker<>())
         );
-
         return Preconditions.check(preconditions, new JavaVisitor<ExecutionContext>() {
 
             @Override
             public J visitIf(J.If if_, ExecutionContext ctx) {
                 J.Switch switch_ = new SwitchCandidate(if_, getCursor()).buildSwitchTemplate();
                 if (switch_ != null) {
-                    return super.visitSwitch(
-                            new JavaIsoVisitor<ExecutionContext>() {
-                                @Override
-                                public J.Case visitCase(J.Case case_, ExecutionContext ctx) {
-                                    if (case_.getBody() == null) {
-                                        return case_;
-                                    }
-                                    if (case_.getBody() instanceof J.Block && ((J.Block) case_.getBody()).getStatements().isEmpty() && !((J.Block) case_.getBody()).getEnd().isEmpty()) {
-                                        return case_.withBody(((J.Block) case_.getBody()).withEnd(Space.EMPTY));
-                                    }
-                                    return case_.withBody(case_.getBody().withPrefix(Space.SINGLE_SPACE));
-                                }
-                            }.visitSwitch(switch_, ctx), ctx);
+                    switch_ = new JavaIsoVisitor<ExecutionContext>() {
+                        @Override
+                        public J.Case visitCase(J.Case case_, ExecutionContext ctx) {
+                            if (case_.getBody() == null) {
+                                return case_;
+                            }
+                            if (case_.getBody() instanceof J.Block &&
+                                    ((J.Block) case_.getBody()).getStatements().isEmpty() &&
+                                    !((J.Block) case_.getBody()).getEnd().isEmpty()) {
+                                return case_.withBody(((J.Block) case_.getBody()).withEnd(Space.EMPTY));
+                            }
+                            return case_.withBody(case_.getBody().withPrefix(Space.SINGLE_SPACE));
+                        }
+                    }.visitSwitch(switch_, ctx);
+                    return super.visitSwitch(switch_, ctx);
                 }
                 return super.visitIf(if_, ctx);
             }
@@ -107,9 +108,10 @@ public class IfElseIfConstructToSwitch extends Recipe {
                     ifPart = handleInstanceOfCheck(ifPart);
                 } else {
                     potentialCandidate = false;
+                    return;
                 }
             }
-            validatePotentialCandidate();
+            potentialCandidate = validatePotentialCandidate();
         }
 
         private J.@Nullable If handleNullCheck(J.If ifPart, Cursor cursor) {
@@ -142,27 +144,26 @@ public class IfElseIfConstructToSwitch extends Recipe {
             return ifPart;
         }
 
-        private void validatePotentialCandidate() {
+        private boolean validatePotentialCandidate() {
             Optional<Expression> switchOn = switchOn();
             // all ifs in the chain must be on the same variable in order to be a candidate for switch pattern matching
-            if (!switchOn.isPresent() || potentialCandidate && !patternMatchers.keySet().stream()
+            if (!switchOn.isPresent() || !patternMatchers.keySet().stream()
                     .map(J.InstanceOf::getExpression)
                     .allMatch(it -> SemanticallyEqual.areEqual(switchOn.get(), it))) {
-                potentialCandidate = false;
-                return;
+                return false;
             }
-            // All InstanceOf checks must have a pattern, otherwise we can't use switch pattern matching (consider calling org.openrewrite.staticanalysis.InstanceOfPatternMatch - or java 17 upgrade - first)
+            // All InstanceOf checks must have a pattern, otherwise we can't use switch pattern matching
+            // (consider calling org.openrewrite.staticanalysis.InstanceOfPatternMatch - or java 17 upgrade - first)
             if (patternMatchers.keySet().stream().anyMatch(instanceOf -> instanceOf.getPattern() == null)) {
-                potentialCandidate = false;
-                return;
+                return false;
             }
             boolean nullCaseInSwitch = nullCheckedParameter != null && SemanticallyEqual.areEqual(nullCheckedParameter, switchOn.get());
             boolean hasLastElseBlock = else_ != null;
 
             // we need at least 3 cases to use a switch
-            if (potentialCandidate && patternMatchers.keySet().size() + (nullCaseInSwitch ? 1 : 0) + (hasLastElseBlock ? 1 : 0) <= 2) {
-                potentialCandidate = false;
-            }
+            return 3 <= patternMatchers.size() +
+                    (nullCaseInSwitch ? 1 : 0) +
+                    (hasLastElseBlock ? 1 : 0);
         }
 
         public J.@Nullable Switch buildSwitchTemplate() {
@@ -223,7 +224,7 @@ public class IfElseIfConstructToSwitch extends Recipe {
             if (statement.getClazz() instanceof J.Identifier) {
                 return ((J.Identifier) statement.getClazz()).getSimpleName();
             } else if (statement.getClazz() instanceof J.FieldAccess) {
-                return  ((J.FieldAccess) statement.getClazz()).toString();
+                return ((J.FieldAccess) statement.getClazz()).toString();
             }
             throw new IllegalStateException("Found unsupported statement where clazz is " + statement.getClazz());
         }
--- a/src/test/java/org/openrewrite/java/migrate/lang/IfElseIfConstructToSwitchTest.java
+++ b/src/test/java/org/openrewrite/java/migrate/lang/IfElseIfConstructToSwitchTest.java
@@ -30,8 +30,8 @@ class IfElseIfConstructToSwitchTest implements RewriteTest {
     @Override
     public void defaults(RecipeSpec spec) {
         spec
-                .recipe(new IfElseIfConstructToSwitch())
-                .allSources(source -> version(source, 21));
+          .recipe(new IfElseIfConstructToSwitch())
+          .allSources(source -> version(source, 21));
     }
 
     @Test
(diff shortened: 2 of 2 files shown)


src/main/java/org/openrewrite/java/migrate/lang/IfElseIfConstructToSwitch.java
```diff
@@ -0,0 +1,216 @@
...
+    public @Nullable Duration getEstimatedEffortPerOccurrence() {
+        return Duration.ofMinutes(3);
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return Preconditions.check(Preconditions.not(new KotlinFileChecker<>()), new JavaVisitor<ExecutionContext>() {
+
+            @Override
+            public J visitIf(J.If iff, ExecutionContext ctx) {
```
jevanlingen:
```suggestion
            public J visitIf(J.If if_, ExecutionContext ctx) {
```

The underscore convention also applies to the `if` now I think about it 😄.

_You can also apply this to all the other `J.If` instances._

jevanlingen (approved):

src/main/java/org/openrewrite/java/migrate/lang/IfElseIfConstructToSwitch.java
```diff
@@ -0,0 +1,220 @@
...
+            if (else_ != null) {
+                // default -> statement
+                arguments[i] = getStatementArgument(else_, cursor);
+            }
+
+            return arguments;
+        }
+
+        String buildTemplate() {
+            StringBuilder switchBody = new StringBuilder("switch (#{any(org.openrewrite.java.tree.Expression)}) {\n");
```
jevanlingen: I think:
```suggestion
            StringBuilder switchBody = new StringBuilder("switch (#{any()}) {\n");
```
will probably work as well.

src/main/java/org/openrewrite/java/migrate/lang/NullCheck.java
```diff
@@ -0,0 +1,149 @@
...
+                ((J.MethodInvocation) nullChecked).getSelect() != null &&
+                couldModifyNullCheckedValue(expression, ((J.MethodInvocation) nullChecked).getSelect())) {
+            return true;
+        }
+        return new JavaIsoVisitor<AtomicBoolean>() {
+
+            private final boolean isCertainlyImmutable = nullChecked.getType() != null && JavaType.Primitive.fromClassName(nullChecked.getType().toString()) != null;
+
+            @Override
+            public J.Identifier visitIdentifier(J.Identifier identifier, AtomicBoolean atomicBoolean) {
```
jevanlingen:
```suggestion
            public J.Identifier visitIdentifier(J.Identifier identifier, AtomicBoolean couldModifyNull) {
```

Giving a variable the name of the type is a bad practice. Use a name that represents what it intention it represents.

--

Notice this applies to all the `atomicBoolean` variables in this visitor class.

src/main/java/org/openrewrite/java/migrate/lang/NullCheck.java
```diff
@@ -0,0 +1,149 @@
...
+            return new Matcher();
+        }
+
+        @Override
+        protected @Nullable NullCheck test(Cursor cursor) {
+            if (cursor.getValue() instanceof J.If) {
+                J.If if_ = cursor.getValue();
+                if (if_.getIfCondition().getTree() instanceof J.Binary) {
+                    J.Binary binary = (J.Binary) if_.getIfCondition().getTree();
+                    if (J.Binary.Type.Equal == binary.getOperator()) {
```
jevanlingen: Some (like me 😎) do advocate static imports, others don't. But to me the following is far more readable

```suggestion
                    if (binary.getOperator() == Equal) {
```

as it turns the code closer to regular English.

_I don't agree with the statement that it's harder to read, because you don't know what this `Equal` thing actually is (because you can't see the type). When reading the code, types don't matter, it's all about intention. When writing the code, types give the guarantee that you wrote it well._

But this is much debatable, so I'll leave it up to you if you apply or ignore this comment!

Jenson3210: Some (like me 😎) do not really have strong feelings about this IF there are no global standards agreed. So if there are others (like you 😎) that do have feelings, these (like me 😎) will gladly commit a change so that the others (like you 😎) can sleep at night 😉

src/test/java/org/openrewrite/java/migrate/lang/IfElseIfConstructToSwitchTest.java
```diff
@@ -0,0 +1,496 @@
...
+                          case Integer i -> formatted = String.format("int %d", i);
+                          case Long l -> formatted = String.format("long %d", l);
+                          case Double d -> formatted = String.format("double %f", d);
+                          case String s -> {
+                              String str = "String";
+                              formatted = String.format("%s %s", str, s);
+                          }
+                          default -> formatted = "unknown";
+                      }
+                      return formatted;
```
jevanlingen, line 80: O boy, I should have seen this earlier. The outcome is still not really what I would have expected as a developer. It would be:

```suggestion
                      String formatted = switch (obj) {
                          case null -> "null";
                          case Integer i -> String.format("int %d", i);
                          case Long l -> String.format("long %d", l);
                          case Double d -> String.format("double %f", d);
                          case String s -> {
                              String str = "String";
                              yield String.format("%s %s", str, s);
                          }
                          default -> "unknown";
                      };
                      return formatted;
```

(or even using the `return` statement directly) This changes the recipe quite much 😢.

Jenson3210: We have another ticket for this as a separate recipe. If you look at the Jep that's another topic/purpose. 
Not all of the if-else-if instance of's are assigning formatted. This is just for this test's sake. 
So in cases where the if else if construct just does stuff and not assigns the same variable, this switch statement would not have to converted to a switch expression. 
But indeed, the tests are working towards this goal as I can then reuse them all the time building on top of the outcome of a previous recipe. 
The goal would be that if you start with java 8 if-else-if construct you in the end can end up the switch expressions indeed, but by doing it over multiple recipes we guarantee that we also clean up other cases as closely as possible to this state. 

src/test/java/org/openrewrite/java/migrate/lang/IfElseIfConstructToSwitchTest.java
```diff
@@ -0,0 +1,496 @@
...
+                          String str = "String";
+                          formatted = String.format("%s %s", str, s);
+                      }
+                      return formatted;
+                  }
+              }
+              """,
+            """
+              class Test {
+                  static String formatter(Object obj) {
```
jevanlingen, line 168: This output is incorrect, as the code will no longer compile:

image

Same thing does also apply to some test below this one.

Jenson3210: well spotted! fix is in the making

src/main/java/org/openrewrite/java/migrate/lang/NullCheck.java
```diff
@@ -0,0 +1,151 @@
...
+    public Statement whenNull() {
+        return getTree().getThenPart();
+    }
+
+    public @Nullable Statement whenNotNull() {
+        J.If.Else anElse = getTree().getElsePart();
+        if (anElse == null) {
+            return null;
+        }
+        return anElse.getBody();
```
jevanlingen:
```suggestion
        return anElse == null ? null : anElse.getBody();
```

src/main/java/org/openrewrite/java/migrate/lang/IfElseIfConstructToSwitch.java
```diff
@@ -0,0 +1,233 @@
...
+                    J.Switch switch_ = JavaTemplate.builder(switchBody)
+                            .contextSensitive()
+                            .build()
+                            .apply(getCursor(), if_.getCoordinates().replace(), arguments)
+                            .withPrefix(if_.getPrefix());
+                    return super.visitSwitch(
+                            new JavaIsoVisitor<ExecutionContext>() {
+                                @Override
+                                public J.Case visitCase(J.Case case_, ExecutionContext ctx) {
+                                    if (!(case_.getBody() instanceof J.Block) || !((J.Block) case_.getBody()).getStatements().isEmpty()) {
```
Jenson3210:
```suggestion
                                    if (!(case_.getBody() instanceof J.Block) || !((J.Block) case_.getBody()).getStatements().isEmpty() ||((J.Block) case_.getBody()).getEnd().isEmpty()) {
```

src/test/java/org/openrewrite/java/migrate/lang/IfElseIfConstructToSwitchTest.java
```diff
@@ -0,0 +1,499 @@
...
+                              formatted = String.format("%s %s", str, s);
+                          }
+                          default -> {}
+                      }
+                      return formatted;
+                  }
+
+                  static String methodInvocation(Tester test) {
+                      String formatted = "initialValue";
+                      switch (test.getObj()) {
```
timtebeek: I'm not sure we should be converting this case, as method invocations could have side effects; consider for instance a case like:
```java
Iterator it = ...;
if (it.next() ...) {}
else if (it.next() ...) {}
else if (it.next() ...) {}
```
I'm going to agree that it's unlikely, especially with a chain of `if` statements, but still perhaps best not converted to be safe.

src/test/java/org/openrewrite/java/migrate/lang/IfElseIfConstructToSwitchTest.java
```diff
@@ -0,0 +1,499 @@
...
+                      switch (obj) {
+                          case null -> formatted = "null";
+                          case Integer i -> formatted = String.format("int %d", i);
+                          case Long l -> formatted = String.format("long %d", l);
+                          case Double d -> formatted = String.format("double %f", d);
+                          case String s -> {
+                              String str = "String";
+                              formatted = String.format("%s %s", str, s);
+                          }
+                          default -> {}
```
timtebeek, line 179: What's the purpose of the `default -> {}` here? I don't really see why that's needed at the moment.

src/test/java/org/openrewrite/java/migrate/lang/IfElseIfConstructToSwitchTest.java
```diff
@@ -0,0 +1,499 @@
...
+              """,
+            """
+              class Test {
+                  static String formatter(Object obj) {
+                      String formatted = "initialValue";
+                      switch (obj) {
+                          case null -> formatted = "null";
+                          case Integer -> formatted = String.format("int %d", (Integer) obj);
+                          case Long -> formatted = String.format("long %d", (Long) obj);
+                          case Double -> formatted = String.format("double %f", (Double) obj);
```
timtebeek: Should we run some separate visitor first to pull these casts into the instanceof before we convert to guards?

src/main/java/org/openrewrite/java/migrate/lang/IfElseIfConstructToSwitch.java
```diff
@@ -0,0 +1,230 @@
...
+            for (Map.Entry<J.InstanceOf, Statement> entry : patternMatchers.entrySet()) {
+                J.InstanceOf instanceOf = entry.getKey();
+                // case class (pattern) -> statement
+                if (instanceOf.getClazz() instanceof J.Identifier) {
+                    arguments[i++] = ((J.Identifier) instanceOf.getClazz()).getSimpleName();
+                } else if (instanceOf.getClazz() instanceof J.FieldAccess) {
+                    arguments[i++] = ((J.FieldAccess) instanceOf.getClazz()).toString();
+                }
+                arguments[i++] = instanceOf.getPattern() == null ? "" : instanceOf.getPattern().withPrefix(Space.SINGLE_SPACE).print(cursor);
+                arguments[i++] = getStatementArgument(entry.getValue(), cursor);
```
timtebeek: I see we're using `print` here to build up the template arguments; that means type information could be lost, and requires the template to be `.contextSensitive()`. Would it be possible to instead use placeholders and pass in the original arguments?

src/test/java/org/openrewrite/java/migrate/lang/IfElseIfConstructToSwitchTest.java
Jenson3210: That would imo be indeed a different recipe, but a valid case! 
I can create a separate issue for [JEP 394](https://javaalmanac.io/features/instanceof-patterns/)? This can be in the java 17 recipe already i guess... 

If we run that one before this one, we would "freely" get the guards added also in this one

timtebeek: Sound like a perfect little isolated recipe, yes please! Would also help us spread the complexity and workload some more.

src/main/java/org/openrewrite/java/migrate/lang/IfElseIfConstructToSwitch.java
Jenson3210: I remember trying this and just getting the toString printed. (will try again)
How would I do this for https://github.com/openrewrite/rewrite-migrate-java/pull/747/files#diff-e7b415092b3371d77c333e96ce2304428ca9a502cbd8c90c66fe2f72ca23ea47R227?
I that case I need to add `;` if it is not a block `{}` but a single statement.

Jenson3210: Problem here is also that if there is no pattern, I want an empty string. How can I send in one case a String, in another one an Expression/Statement? If that's possible, I think we can indeed remove the printing

src/test/java/org/openrewrite/java/migrate/lang/IfElseIfConstructToSwitchTest.java
Jenson3210: - https://github.com/openrewrite/rewrite-migrate-java/issues/763

src/test/java/org/openrewrite/java/migrate/lang/IfElseIfConstructToSwitchTest.java
Jenson3210: Screenshot 2025-06-17 at 09 23 41

Jenson3210: https://docs.oracle.com/javase/specs/jls/se17/preview/specs/patterns-switch-jls.html#jls-14.11.2
enhanced switch statements must be exhaustive

Jenson3210: This brought me to the following issue: it seems the label is mandatory also. IJ mentions to create an `ignored` label, but I think we should come up with the label addition first to avoid that people now introduce `ignored` when migrating and then not get the cast -> label transformation anymore... 
So we will need that other recipe also to make sense of the switch statements of older languages. 
We can then call that one to be sure our instanceOf checks are labeled

Jenson3210:
I just updated a test that now check that: if no pattern matching is present, we will not add it through this recipe. 
There is a recipe for this existing already and I would not combine calling that one from this recipe as it might just add Pattern matching without doing the conversion to switch -> different responsibilities. 

Mostly these 2 recipes will be executed together through the upgrade java recipes. `InstanceOfPatternMatching` is present in java 17 upgrade, and this one will be part of java 21 upgrade.

timtebeek:
Could be nice to add a test that runs two recipes back to back : `.recipe(new InstanceOfPatternMatching(), new IfElseIfConstructToSwitch())`, and then see that is works well.

timtebeek (approved):
Great! Thanks for reworking this to no longer needing a `print`. Good to finally close out the set. 🚀

PR #748, opened 2025-06-02
Null check as switch case

Jenson3210:
## What's changed?
Cutting up the task at hand in multiple smaller PR's for better insights, risk assessment... 
- NullCheckAsSwitchCase

## What's your motivation?
- #306

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/lang/NullCheckAsSwitchCase.java
+++ b/src/main/java/org/openrewrite/java/migrate/lang/NullCheckAsSwitchCase.java
@@ -1,0 +1,115 @@
+/*
+ * Copyright 2025 the original author or authors.
+ * <p>
+ * Licensed under the Moderne Source Available License (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://docs.moderne.io/licensing/moderne-source-available-license
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.java.migrate.lang;
+
+import lombok.EqualsAndHashCode;
+import lombok.Value;
+import org.jspecify.annotations.Nullable;
+import org.openrewrite.*;
+import org.openrewrite.internal.ListUtils;
+import org.openrewrite.java.JavaTemplate;
+import org.openrewrite.java.JavaVisitor;
+import org.openrewrite.java.search.SemanticallyEqual;
+import org.openrewrite.java.tree.Expression;
+import org.openrewrite.java.tree.J;
+import org.openrewrite.java.tree.Space;
+import org.openrewrite.java.tree.Statement;
+import org.openrewrite.staticanalysis.kotlin.KotlinFileChecker;
+
+import java.time.Duration;
+import java.util.Optional;
+import java.util.concurrent.atomic.AtomicReference;
+
+import static java.util.Objects.requireNonNull;
+import static org.openrewrite.java.migrate.lang.NullCheck.Matcher.nullCheck;
+
+@Value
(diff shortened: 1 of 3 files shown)


src/test/java/org/openrewrite/java/migrate/lang/NullCheckAsSwitchCaseTest.java
```diff
@@ -0,0 +1,560 @@
...
+                      }
+                  }
+              }
+              """
+          )
+        );
+    }
+
+    @Test
+    void doNotMergeWhenNullBlockReturnsSomething() {
```
jevanlingen: We can leave the if-check there, it's fine by me, but you can also return from out a switch expression if you want to:

image

_Notice the `yield` keyword was added in Java 13 to return from the expression._

---

**Edit**: Now I think a little bit more about this, I think it's better to **leave** it as you **implemented** it. Because once you refine the switch case, it's starts to become ugly. As this does not compile:

image

But it's up to you if you want to change this or not 😉.

src/main/java/org/openrewrite/java/migrate/lang/NullCheckAsSwitchCase.java
```diff
@@ -0,0 +1,115 @@
...
+                            return statement;
+                        }
+                        // The switch must switch on the same expression as the nullCheck
+                        if (!SemanticallyEqual.areEqual(((J.Switch) nextStatement).getSelector().getTree(), check.getNullCheckedParameter())) {
+                            return statement;
+                        }
+                        // If the null checked block returns early or potentially reassigns the value of the check object (touching it with a methodInvocation or accessing a field or directly assigning the null-checked identifier)
+                        if (check.returns() || check.couldModifyNullCheckedValue()) {
+                            return statement;
+                        }
```
jevanlingen: I think doing it in one go is actually easier to read:

```suggestion
                        if (!(nextStatement instanceof J.Switch) ||
                                !SemanticallyEqual.areEqual(((J.Switch) nextStatement).getSelector().getTree(), check.getNullCheckedParameter()) ||
                                check.returns() || 
                                check.couldModifyNullCheckedValue()) {
                            return statement;
                        }
```

You can even argue if the comment is really needed if you format it like this. I think it kinda easy to read without any explanation 😄.

src/test/java/org/openrewrite/java/migrate/lang/NullCheckAsSwitchCaseTest.java
Jenson3210: I was thinking not to do it so that indeed we can have later a transformation towards switch expressions (which is not possible if 1, but not all branches return something. 

```
if (o == null) {
    return null;
}
switch (o) {
    case String s -> formatted = "Text: " + s;
    case Double d -> formatted = "Double: " + s.toString();
    default -> formatted = Objects.toString(o);
}
return formatted.toUppercase();
```
would not allow for further conversion to switch expressions afterwards due to the risk of NPE. 
I thought it was best to: if the body of the null check alters the code flow in a way that it does something different then the switch, we keep the null check separate. 
```
switch (o) {
    case null -> return null;
    case String s -> formatted = "Text: " + s;
    case Double d -> formatted = "Double: " + s.toString();
    default -> formatted = Objects.toString(o);
}
```
would indeed be possible but will not allow conversion anymore to expressions and I rather end up with 
```
if (o == null) {
    return null;
}
formatted = switch (o) {
    case String s -> "Text: " + s;
    case Double d -> "Double: " + s.toString();
    default -> Objects.toString(o);
}
return formatted.toUppercase();
```

as 
```
formatted = switch (o) {
    case null -> null
    case String s -> "Text: " + s;
    case Double d -> "Double: " + s.toString();
    default -> Objects.toString(o);
}
return formatted.toUppercase(); // NPE
```
would be invalid

jevanlingen (approved):

Jenson3210:
@timtebeek would you want any further review? Or can I go ahead and do the necessary to merge (adapt the declarative yaml)

Jenson3210:
Added the Throws to the if statement. 

Now, this will also be the case for J.Return and any others that might not be extending Expression but that do extend Statement. 


Should we add a test for each of these? Or should we just pass all Statements (hence all J's?) to the method? Do we then need tests for all these cases?

PR #753, opened 2025-06-02
Add recipe to replace ManagedBean with Named annotation

evie-lau:
## What's changed?
New recipe to migrate `@ManagedBean` from JSF/Jakarta Faces to the recommended CDI-based approach using `@jakarta.inject.Named`

## What's your motivation?
ManagedBean is removed from EE10/Faces 4.0, and the existing [recipe](https://github.com/openrewrite/rewrite-migrate-java/blob/90713334d935cc8577dd9f7c2518480c89b0b6ac/src/main/resources/META-INF/rewrite/jakarta-faces-4.yml#L204) doesn't migrate the `@ManagedBean` annotation.
It seems using the CDI `@Named` is the general approach to replacing `@ManagedBean`.

## Any additional context

This current recipe converts both javax and jakarta ManagedBean to the jakarta CDI Named annotation, making this appropriate for EE9/10 migration, but not for EE8 users wanting to switch to CDI. 
### Checklist

### DIFF
--- a/src/main/resources/META-INF/rewrite/jakarta-faces-3.yml
+++ b/src/main/resources/META-INF/rewrite/jakarta-faces-3.yml
@@ -37,7 +37,6 @@ recipeList:
   - org.openrewrite.java.migrate.jakarta.RemovedJakartaFacesResourceResolver
   - org.openrewrite.java.migrate.jakarta.RemovedStateManagerMethods
   - org.openrewrite.java.migrate.jakarta.RemovedUIComponentConstant
-  - org.openrewrite.java.migrate.jakarta.FacesManagedBeansDeprecated
   - org.openrewrite.java.migrate.jakarta.UpgradeFaces3OpenSourceLibraries
 ---
 type: specs.openrewrite.org/v1beta/recipe
--- a/src/main/resources/META-INF/rewrite/jakarta-faces-4.yml
+++ b/src/main/resources/META-INF/rewrite/jakarta-faces-4.yml
@@ -199,7 +199,6 @@ recipeList:
       elementName: web-app
       newValue: https://jakarta.ee/xml/ns/jakartaee https://jakarta.ee/xml/ns/jakartaee/web-app_6_0.xsd
 ---
-# TODO: currently same as FacesManagedBeansDeprecated in Faces3 migration. Can this list be pared down?
 type: specs.openrewrite.org/v1beta/recipe
 name: org.openrewrite.java.migrate.jakarta.FacesManagedBeansRemoved
 displayName: Substitute removed Faces Managed Beans


src/main/java/org/openrewrite/java/migrate/jakarta/UpdateManagedBeanToNamed.java
```diff
@@ -0,0 +1,85 @@
...
+    public String getDescription() {
+        return "Faces ManagedBean was deprecated in JSF 2.3 (EE8) and removed in Jakarta Faces 4.0 (EE10). " +
+               "Replace `@ManagedBean` with `@Named` for CDI-based bean management.";
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return new JavaIsoVisitor<ExecutionContext>() {
+            private final AnnotationMatcher MANAGED_BEAN_MATCHER_JAVAX = new AnnotationMatcher("javax.faces.bean.ManagedBean");
+            private final AnnotationMatcher MANAGED_BEAN_MATCHER_JAKARTA = new AnnotationMatcher("jakarta.faces.bean.ManagedBean");
```
timtebeek: Would you mind adding a quick precondition here on the use of those types? That'd make for a cheap performance improvement as it avoid visiting any compilation units unnecessarily.

src/main/java/org/openrewrite/java/migrate/jakarta/UpdateManagedBeanToNamed.java
```diff
@@ -0,0 +1,85 @@
...
+                    // Get the name from the @ManagedBean annotation
+                    String beanName;
+                    beanName = annotation.getArguments() == null ? null :
+                            annotation.getArguments().stream()
+                                    .filter(J.Assignment.class::isInstance)
+                                    .map(J.Assignment.class::cast)
+                                    .filter(arg -> arg.getVariable().toString().equals("name"))
+                                    .findFirst()
+                                    .map(arg -> arg.getAssignment().toString())
+                                    .orElse(null);
```
timtebeek: You might be able to get the bean name more easily with a construct similar to what's seen here:
https://github.com/openrewrite/rewrite-spring/blob/16c63561b254f72a33ac1f70926cc62d233a8d70/src/main/java/org/openrewrite/java/spring/trait/SpringBean.java#L46-L50

src/main/java/org/openrewrite/java/migrate/jakarta/UpdateManagedBeanToNamed.java
```diff
@@ -0,0 +1,85 @@
...
+                                .javaParser(JavaParser.fromJavaVersion().classpathFromResources(ctx, "jakarta.inject-api-2.0.1"))
+                                .imports("jakarta.inject.Named")
+                                .build()
+                                .apply(getCursor(), annotation.getCoordinates().replace());
+                    }
+                    maybeAddImport("jakarta.inject.Named");
+                    maybeRemoveImport("javax.faces.bean.ManagedBean");
+                    maybeRemoveImport("jakarta.faces.bean.ManagedBean");
+                }
+                return annotation;
```
timtebeek: If there's any chance of these appearing in nested annotations, then it might make sense to call `super.visitAnnotation` here.

src/main/java/org/openrewrite/java/migrate/jakarta/UpdateManagedBeanToNamed.java
```diff
@@ -0,0 +1,85 @@
...
+                    } else {
+                        annotation = JavaTemplate.builder("@Named")
+                                .javaParser(JavaParser.fromJavaVersion().classpathFromResources(ctx, "jakarta.inject-api-2.0.1"))
+                                .imports("jakarta.inject.Named")
+                                .build()
+                                .apply(getCursor(), annotation.getCoordinates().replace());
+                    }
+                    maybeAddImport("jakarta.inject.Named");
+                    maybeRemoveImport("javax.faces.bean.ManagedBean");
+                    maybeRemoveImport("jakarta.faces.bean.ManagedBean");
```
timtebeek: Slight preference for immediate returns here, unless you want to call `super` with the modified annotation below.
```suggestion
                    maybeAddImport("jakarta.inject.Named");
                    maybeRemoveImport("javax.faces.bean.ManagedBean");
                    maybeRemoveImport("jakarta.faces.bean.ManagedBean");
                    // Replace the @ManagedBean annotation with @Named
                    if (beanName != null) {
                        return JavaTemplate.builder("@Named(\"#{}\")")
                                .javaParser(JavaParser.fromJavaVersion().classpathFromResources(ctx, "jakarta.inject-api-2.0.1"))
                                .imports("jakarta.inject.Named")
                                .build()
                                .apply(getCursor(), annotation.getCoordinates().replace(), beanName);
                    } else {
                        return JavaTemplate.builder("@Named")
                                .javaParser(JavaParser.fromJavaVersion().classpathFromResources(ctx, "jakarta.inject-api-2.0.1"))
                                .imports("jakarta.inject.Named")
                                .build()
                                .apply(getCursor(), annotation.getCoordinates().replace());
                    }
```

src/main/java/org/openrewrite/java/migrate/jakarta/UpdateManagedBeanToNamed.java
```diff
@@ -0,0 +1,85 @@
...
+        return new JavaIsoVisitor<ExecutionContext>() {
+            private final AnnotationMatcher MANAGED_BEAN_MATCHER_JAVAX = new AnnotationMatcher("javax.faces.bean.ManagedBean");
+            private final AnnotationMatcher MANAGED_BEAN_MATCHER_JAKARTA = new AnnotationMatcher("jakarta.faces.bean.ManagedBean");
+
+            @Override
+            public J.Annotation visitAnnotation(J.Annotation annotation, ExecutionContext ctx) {
+                if (MANAGED_BEAN_MATCHER_JAVAX.matches(annotation) || MANAGED_BEAN_MATCHER_JAKARTA.matches(annotation)) {
+                    // Get the name from the @ManagedBean annotation
+                    String beanName;
+                    beanName = annotation.getArguments() == null ? null :
```
timtebeek: Unrelated to the below trait we can immediately assign here.
```suggestion
                    String beanName = annotation.getArguments() == null ? null :
```

src/test/java/org/openrewrite/java/migrate/jakarta/UpdateManagedBeanToNamedTest.java
```diff
@@ -0,0 +1,84 @@
...
+
+class UpdateManagedBeanToNamedTest implements RewriteTest {
+    @Override
+    public void defaults(RecipeSpec spec) {
+        spec.parser(JavaParser.fromJavaVersion()
+            .classpathFromResources(new InMemoryExecutionContext(), "jsf-api-2.1.29-11", "jakarta.faces-api-3.0.0", "jakarta.inject-api-2.0.1"))
+          .recipe(new UpdateManagedBeanToNamed());
+    }
+
+    @DocumentExample
```
timtebeek: We don't yet support this annotation on parameterized tests. Fine to leave it out for now then.
```suggestion
```

timtebeek: Alternatively two separate tests could be added, one of which carries this annotation such that it shows up in the docs.

timtebeek (commented):
Couple quick comments; great to see already! Ought to be quick to address and then merge.

src/main/java/org/openrewrite/java/migrate/jakarta/UpdateManagedBeanToNamed.java
evie-lau: I don't think there's any chance of nested annotations here.

timtebeek (approved):
Thanks for the quick fixes!

Should `org.openrewrite.java.migrate.jakartaUpdateManagedBeanToNamed` be added to any larger declarative migration recipe before we merge?

evie-lau:
Just to clarify the final decision:
- added this recipe into the Faces 4.0 migration, where ManagedBean is removed and migration is required. 
- followed up on #742 and removed the Faces Bean -> CDI bean recipe list from Faces 3.0 migration

The managed bean migration recipe can still be invoked separately for anyone on EE9 wanting to get off the deprecated package.

PR #754, opened 2025-06-04
Pick better SDKMan versions

timtebeek:
## What's your motivation?
Right now we pick an incorrect distribution for Zulu upgrades to Java 17.
There's are the versions that are available:
```
17.0.13.crac-zulu
17.0.14-zulu
17.0.14.crac-zulu
17.0.14.fx-zulu
17.0.15-zulu
17.0.15.crac-zulu
17.0.15.fx-zulu
```

We ought to pick `17.0.15-zulu` as the highest version number, with a matching (non fx/crac) distribution.

In practice, we pick 17.0.13.crac-zulu as the first match here:
https://github.com/openrewrite/rewrite-migrate-java/blob/962b76c47d6115ba48e553189fb78181094fb07c/src/main/java/org/openrewrite/java/migrate/UpdateSdkMan.java#L96-L98

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/UpdateSdkMan.java
+++ b/src/main/java/org/openrewrite/java/migrate/UpdateSdkMan.java
@@ -20,8 +20,10 @@ import lombok.Value;
 import org.jspecify.annotations.Nullable;
 import org.openrewrite.*;
 import org.openrewrite.binary.Binary;
+import org.openrewrite.internal.ListUtils;
 import org.openrewrite.quark.Quark;
 import org.openrewrite.remote.Remote;
+import org.openrewrite.semver.LatestRelease;
 import org.openrewrite.text.PlainText;
 import org.openrewrite.text.PlainTextParser;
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/openrewrite/java/migrate/UpdateSdkMan.java
```diff
@@ -92,11 +94,18 @@ public Tree visit(@Nullable Tree tree, ExecutionContext ctx) {
...
-                            return plainText.withText(matcher.replaceFirst("java=" + candidate));
-                        }
+                    String dist = newDistribution == null ? matcher.group(2) : "-" + newDistribution;
+                    String oldFull = matcher.group(1) + matcher.group(2);
+                    String newBasis = ver + dist;
+                    LatestRelease releaseComparator = new LatestRelease(dist);
+                    Pattern majorPattern = Pattern.compile("^" + ver + "[.-].*");
+                    List<String> filteredCandidates = ListUtils.filter(
+                            readSdkmanJavaCandidates(),
+                            candidate -> majorPattern.matcher(candidate).matches() && (releaseComparator.isValid(oldFull, candidate) || releaseComparator.isValid(newBasis, candidate))
```
steve-aom-elliott: I realize the `||` part could potentially be incorrect when switching distributions seeing it now. I think I can just fully get rid of the comparison against `oldFull` and just use the comparison against `newBasis`.

timtebeek (commented):
Thanks a lot for the help here! Nice how you've figured out LatestRelease.

We might separately find that new JDK releases will break the tests in the future, but we can figure that out by then; possibly by adding a static src/test/resources/ variant of of sdkman versions.

PR #755, opened 2025-06-05
Make fields annotated with `@Produces` static

karthikNousher:
## What's changed?

- Added a recipe to modify session beans that have `@Produces`-annotated fields by making them `static`.
- Ensured the recipe only targets classes annotated with `@Stateless`, `@Stateful`, or `@Singleton`.
- Created a unit test to verify the behavior of the recipe using `JavaParser.fromJavaVersion().dependsOn(...)`.

## Any additional context
- Used mock annotations in tests (`@Stateless`, `@Produces`) via `dependsOn()` to avoid runtime dependencies.
- The recipe applies a JavaTemplate that rewrites `private SomeDependency someDependency;` to `private static SomeDependency someDependency;`.

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/AddStaticVariableOnProducerSessionBean.java
+++ b/src/main/java/org/openrewrite/java/migrate/AddStaticVariableOnProducerSessionBean.java
@@ -19,8 +19,6 @@ import org.openrewrite.*;
 import org.openrewrite.java.JavaVisitor;
 import org.openrewrite.java.search.UsesType;
 import org.openrewrite.java.tree.J;
-import org.openrewrite.java.tree.JavaType;
-import org.openrewrite.java.tree.Space;
 import org.openrewrite.marker.Markers;
 
 import java.util.ArrayList;
--- a/src/main/java/org/openrewrite/java/migrate/AddStaticVariableOnProducerSessionBean.java
+++ b/src/main/java/org/openrewrite/java/migrate/AddStaticVariableOnProducerSessionBean.java
@@ -1,4 +1,18 @@
-package org.openrewrite.java.migrate;
+/*
+ * Copyright 2025 the original author or authors.
+ * <p>
+ * Licensed under the Moderne Source Available License (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://docs.moderne.io/licensing/moderne-source-available-license
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 
 import org.openrewrite.*;
 import org.openrewrite.java.JavaVisitor;
@@ -28,7 +42,7 @@ public class AddStaticVariableOnProducerSessionBean extends Recipe {
         return Preconditions.check(new UsesType<>("jakarta.enterprise.inject.Produces", false),
                 new JavaVisitor<ExecutionContext>() {
                     @Override
-                    public J visitVariableDeclarations(J.VariableDeclarations multiVariable, ExecutionContext executionContext) {
+                    public J visitVariableDeclarations(J.VariableDeclarations multiVariable, ExecutionContext ctx) {
 
                         J.ClassDeclaration visitingClass = getCursor().firstEnclosing(J.ClassDeclaration.class);
                         boolean isSessionBean = visitingClass.getLeadingAnnotations().stream()
@@ -48,7 +62,7 @@ public class AddStaticVariableOnProducerSessionBean extends Recipe {
                                 return multiVariable.withModifiers(updatedModifiers);
                             }
                         }
-                        return super.visitVariableDeclarations(multiVariable, executionContext);
+                        return super.visitVariableDeclarations(multiVariable, ctx);
                     }
                 });
     }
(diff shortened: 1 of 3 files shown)


src/main/java/org/openrewrite/java/migrate/AddStaticVariableOnProducerSessionBean.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    @Override
+    public String getDescription() {
+        return "Ensures that the fields annotated with @Produces which is inside the session bean (@Stateless, @Stateful, or @Singleton) are declared static.";
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return new JavaVisitor<ExecutionContext>() {
```
timtebeek: Can we add a Precondition.check here to verify UsesType `javax.enterprise.inject.Produces`?

src/main/java/org/openrewrite/java/migrate/AddStaticVariableOnProducerSessionBean.java
```diff
@@ -0,0 +1,81 @@
...
+
+            @Override
+            public J visitClassDeclaration(J.ClassDeclaration cs, ExecutionContext ctx) {
+                boolean isSessionBean = cs.getLeadingAnnotations().stream()
+                        .map(J.Annotation::getSimpleName)
+                        .anyMatch(name -> name.equals("Stateless") || name.equals("Stateful") || name.equals("Singleton"));
+
+                if (isSessionBean) {
```
evie-lau: The logic inside here can be moved into a `visitVariableDeclarations` and just call `super.visitClassDeclaration` here.

src/main/java/org/openrewrite/java/migrate/AddStaticVariableOnProducerSessionBean.java
karthikNousher: Thanks @timtebeek for the comment, have added the pre-condition check as you suggested.

src/main/java/org/openrewrite/java/migrate/AddStaticVariableOnProducerSessionBean.java
karthikNousher: Have used visitvariabledeclaration instead of classdeclaration

src/main/java/org/openrewrite/java/migrate/AddStaticVariableOnProducerSessionBean.java
```diff
@@ -0,0 +1,69 @@
...
+ * <p>
+ * https://docs.moderne.io/licensing/moderne-source-available-license
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
```
timtebeek:
```suggestion
 */
package org.openrewrite.java.migrate;

```

src/main/java/org/openrewrite/java/migrate/AddStaticVariableOnProducerSessionBean.java
```diff
@@ -0,0 +1,69 @@
...
+import org.openrewrite.java.search.UsesType;
+import org.openrewrite.java.tree.J;
+import org.openrewrite.java.tree.Space;
+import org.openrewrite.marker.Markers;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+public class AddStaticVariableOnProducerSessionBean extends Recipe {
```
timtebeek, line 29: Should `org.openrewrite.java.migrate.AddStaticVariableOnProducerSessionBean` be added to any larger composite recipe?

timtebeek (commented):
Added some polish and a final question before we merge.

src/main/java/org/openrewrite/java/migrate/AddStaticVariableOnProducerSessionBean.java
```diff
@@ -0,0 +1,70 @@
...
+                                hasAnnotation(multiVariable, "@jakarta.enterprise.inject.Produces") &&
+                                isSessionBean()) {
+                            return multiVariable.withModifiers(ListUtils.concat(multiVariable.getModifiers(),
+                                    new J.Modifier(Tree.randomId(), Space.SINGLE_SPACE, Markers.EMPTY, null, J.Modifier.Type.Static, Collections.emptyList())));
+                        }
+                        return super.visitVariableDeclarations(multiVariable, ctx);
+                    }
+
+                    private boolean isSessionBean() {
+                        J.ClassDeclaration j = getCursor().firstEnclosing(J.ClassDeclaration.class);
```
evie-lau: @timtebeek Would it be better to have this check as a ClassDeclaration visitor to reduce visits in the case it doesn't have any of the 3 annotations?

timtebeek: There's a slight risk here with nested classes; if the outer most class lacks those annotations, then any inner classes are not visited; not sure if that would apply here, but figured point it out.

Also seems like those `jakarta.ejb` classes should be part of the precondition to avoid unnecessary visits. Thanks for point this out!

evie-lau: Hmm, that's a good point. I don't think it's common, but it may still be a possibility. May want to revert commit [95537c9](https://github.com/openrewrite/rewrite-migrate-java/pull/755/commits/95537c902866db1f31d63a60987b6bfc9f533929) switching to ClassDeclaration visitor

timtebeek:
Thanks both!

PR #761, opened 2025-06-12
Set `acceptTransitive` to true for JSpecify recipes

### LINKED ISSUE #757: Avoid adding `org.jspecify:jspecify` if already available transitively

When running the [Migrate from Spring Framework annotations to JSpecify](https://docs.openrewrite.org/recipes/java/jspecify/migratefromspringframeworkannotations) recipe on a Spring-based project, OpenRewrite adds the `org.jspecify:jspecify` dependency.

As [this is already included transitively](https://github.com/spring-projects/spring-framework/blob/be02d961fcbe6dc583b9ba7c46c74de465e6721d/spring-core/spring-core.gradle#L75) by the Spring Core module, the explicit declaration is not required in the project running the recipe.

Ideally, after the recipe runs, there should be no new JSpecify dependency added to the build configuration, regardless of whether it's Maven or Gradle.

Yes: inspect all build configurations manually and delete the newly introduced dependency.

## Additional context

This was spotted during the work at https://github.com/spring-projects/spring-batch/pull/4864/commits/d5198494bd49be332350d7edf9d611cf046f7e8a.

If the project agrees to this change, I'd be happy to contribute! 🙂

scordio:
## What's changed?
Prior to this change, the JSpecify recipes add `org.jspecify:jspecify` as a dependency, even for cases where the dependency is already available transitively.

Now, `acceptTransitive` is set to `true`, so that the dependency isn't added when it's already available transitively.

## What's your motivation?
* Fixes #757

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/AddStaticVariableOnProducerSessionBean.java
+++ b/src/main/java/org/openrewrite/java/migrate/AddStaticVariableOnProducerSessionBean.java
@@ -1,0 +1,78 @@
+/*
+ * Copyright 2025 the original author or authors.
+ * <p>
+ * Licensed under the Moderne Source Available License (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://docs.moderne.io/licensing/moderne-source-available-license
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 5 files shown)


scordio:
> * [ ]  I've added unit tests to cover both positive and negative cases

Is there any test case I could use as an example to verify the behavior with transitive dependencies?

My best guess so far is to duplicate the existing methods in `JSpecifyBestPracticesTest` and add something to the POM that brings JSpecify transitively. To be precise, `spring-core` would be what I'd add to the `MigrateFromSpringFrameworkAnnotations` test case but I'm unsure about the other cases.

timtebeek:
Only spotted this now! Sorry about that, I don't get notified about drafts. Looks perfect to me as is; no need to add tests beyond what we already have for this one in particular. Thanks a lot!

Issue #760, opened 2025-06-12
Applying AddJaxwsRuntime adds invalid dependency


BrendanHart:
## What version of OpenRewrite are you using?

I am using

- rewrite-migrate-java latest.release
- Maven/Gradle plugin latest.release

## How are you running OpenRewrite?

I am using a Groovy init script.
```groovy
initscript {
    repositories {
        maven { url "https://plugins.gradle.org/m2" }
    }
    dependencies { classpath("org.openrewrite:plugin:latest.release") }
}
rootProject {
    plugins.apply(org.openrewrite.gradle.RewritePlugin)
    dependencies {
        rewrite("org.openrewrite.recipe:rewrite-migrate-java:latest.release")

    }
    rewrite {
        activeRecipe("org.openrewrite.java.migrate.javax.AddJaxwsRuntime")
    }
    afterEvaluate {
        if (repositories.isEmpty()) {
            repositories {
                mavenCentral()
            }
        }
    }
}

```

## What is the smallest, simplest way to reproduce the problem?

```groovy
plugins {
    id 'java'
}
repositories {
    mavenCentral()
}
dependencies {
    compileOnly "com.sun.xml.ws:jaxws-rt:2.3.7"

    implementation "org.springframework.boot:spring-boot-starter-web-services:2.5.15"

    testImplementation "com.sun.xml.ws:jaxws-rt:2.3.7"
}
```

## What did you expect to see?

The output only contains valid dependencies.

## What did you see instead?

```groovy
plugins {
    id 'java'
}
repositories {
    mavenCentral()
}
dependencies {
    compileOnly "com.sun.xml.ws:jaxws-rt:2.3.7"

    implementation "org.springframework.boot:spring-boot-starter-web-services:2.5.15"

    runtimeClasspath "com.sun.xml.ws:jaxws-rt:2.3.7"

    testImplementation "com.sun.xml.ws:jaxws-rt:2.3.7"
}
```

## What is the full stack trace of any errors you encountered?

When we build this we see:
```
> Dependencies can not be declared against the `runtimeClasspath` configuration.
```

PR #762, opened 2025-06-13
Add only declarable Gradle jax-ws runtime dependencies

jevanlingen:
## What's changed?
For the `AddJaxwsRuntimeGradle` recipe the resolution-only dependencies are no longer added.

- Fixes https://github.com/openrewrite/rewrite-migrate-java/issues/760

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/javax/AddJaxwsRuntime.java
+++ b/src/main/java/org/openrewrite/java/migrate/javax/AddJaxwsRuntime.java
@@ -152,7 +152,7 @@ public class AddJaxwsRuntime extends Recipe {
                     for (GradleDependencyConfiguration configuration : tmpConfigurations) {
                         GradleDependencyConfiguration gdc = gp.getConfiguration(configuration.getName());
                         for (GradleDependencyConfiguration extendsFrom : gdc.allExtendsFrom()) {
-                            if (configurations.contains(extendsFrom.getName())) {
+                            if (configurations.contains(extendsFrom)) {
                                 configurations.remove(configuration);
                             }
                         }
--- a/src/main/java/org/openrewrite/java/migrate/javax/AddJaxwsRuntime.java
+++ b/src/main/java/org/openrewrite/java/migrate/javax/AddJaxwsRuntime.java
@@ -46,13 +46,8 @@ public class AddJaxwsRuntime extends Recipe {
     private static final String SUN_JAXWS_RUNTIME_GROUP = "com.sun.xml.ws";
     private static final String SUN_JAXWS_RUNTIME_ARTIFACT = "jaxws-rt";
 
-    private final AddJaxwsRuntimeGradle addJaxwsRuntimeGradle;
-    private final AddJaxwsRuntimeMaven addJaxwsRuntimeMaven;
-
-    public AddJaxwsRuntime() {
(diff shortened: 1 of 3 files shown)


src/main/java/org/openrewrite/java/migrate/javax/AddJaxwsRuntime.java
```diff
@@ -170,6 +167,14 @@ private Set<String> getTransitiveDependencyConfiguration(GradleProject gp, Strin
 
                     return configurations;
                 }
+
+                private G.CompilationUnit addJaxWsRuntimeDependency(String apiConfiguration, G.CompilationUnit g, ExecutionContext ctx) {
+                    if (apiConfiguration.endsWith("Classpath")) {
```
jevanlingen: 💡 Though we do not have a special unit test for this case, the same applies for the maven variant:

https://github.com/openrewrite/rewrite-migrate-java/blob/867b956d9c7dc8cff1f8b23aa3f85c09845017f8/src/main/java/org/openrewrite/java/migrate/javax/AddJaxwsRuntime.java#L208-L212

That piece of code has also some if-statement and a resolved-scope check. So personally I think we can move forward with this PR (even if we do not have a specific test to cover this is-it-a-resolution-only-config).

timtebeek:
@jevanlingen there was a perhaps related report in https://github.com/openrewrite/rewrite-migrate-java/issues/760 that might help you reproduce the issue; could be worth briefly seeing if that indeed helps, or is fixed now.

jevanlingen:
@timtebeek, yeah awesome, this is exactly the case this PR does fix!

src/main/java/org/openrewrite/java/migrate/javax/AddJaxwsRuntime.java
```diff
@@ -116,12 +111,10 @@ public G.CompilationUnit visitCompilationUnit(G.CompilationUnit cu, ExecutionCon
...
-                                        .visitNonNull(g, ctx);
+                                g = addJaxWsRuntimeDependency("compileOnly", g, ctx);
                             }
                             if (gp.getConfiguration("testImplementation") != null) {
-                                g = (G.CompilationUnit) new org.openrewrite.gradle.AddDependencyVisitor(SUN_JAXWS_RUNTIME_GROUP, SUN_JAXWS_RUNTIME_ARTIFACT, "2.3.x", null, "testImplementation", null, null, null, null, null)
-                                        .visitNonNull(g, ctx);
+                                g = addJaxWsRuntimeDependency("testImplementation", g, ctx);
                             }
                         } else {
                             for (String apiConfiguration : apiConfigurations) {
```
Jenson3210: I would expect to also see some filtering here on the newly introduced method from Sam that know if a configuration is `canBeDeclared()` perhaps?

jevanlingen: https://github.com/openrewrite/rewrite-migrate-java/pull/762/commits/6def62acbcdd7f28188a8e389bcf4467c372979b

Jenson3210 (approved):
Good to see that new method from Sam being used. This way we adhere to the configurations that are correctly created while we keep backwards compatible with these that end on classPath. 👍

timtebeek:
Thanks all!

PR #765, opened 2025-06-17
Update AddStaticVariableOnProducerSessionBean to modify XML defined EJBs

karthikNousher:
## What's changed?

- Added a xml scanning to detect the session beans and also apply the recipe to those classes.
- Recipe now updates the classes that are specified in the `ejb-jar.xml` file
- Added new test case to test the xml scanning functionality.

## What's your motivation?

- To create a recipe that can not only be used to add a static variable to session beans  annotated with `@Stateless`, `@Stateful`, or `@Singleton`, but also to those defined in the xml file.

## Any additional context

- Extended the recipe to scan `ejb-jar.xml` for session bean definitions and incorporate these beans in the recipe processing.
- This way it allows the recipe to apply changes to classes that does not have `@Stateless`, `@Stateful`, or `@Singleton`
annotation but are specified as session beans in the XML configuration.

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/AddStaticVariableOnProducerSessionBean.java
+++ b/src/main/java/org/openrewrite/java/migrate/AddStaticVariableOnProducerSessionBean.java
@@ -49,6 +49,7 @@ public class AddStaticVariableOnProducerSessionBean extends ScanningRecipe<Set<S
 
     @Override
     public Set<String> getInitialValue(ExecutionContext ctx) {
+        // Class names of session beans found in XML
         return new HashSet<>();
     }
 
@@ -57,7 +58,6 @@ public class AddStaticVariableOnProducerSessionBean extends ScanningRecipe<Set<S
         return Preconditions.check(
                 new FindSourceFiles("**/ejb-jar.xml"),
                 new XmlVisitor<ExecutionContext>() {
-
                     @Override
                     public Xml visitTag(Xml.Tag tag, ExecutionContext ctx) {
                         if (EJB_PATH.matches(getCursor())) {
@@ -107,10 +107,8 @@ public class AddStaticVariableOnProducerSessionBean extends ScanningRecipe<Set<S
 
                     private boolean isInXml() {
                         J.ClassDeclaration parentClass = getCursor().firstEnclosing(J.ClassDeclaration.class);
-                        if (parentClass.getType() instanceof JavaType.FullyQualified) {
-                            JavaType.FullyQualified fqType = parentClass.getType();
-                            String fqName = fqType.getFullyQualifiedName();
-                            return acc.contains(fqName);
+                        if (parentClass != null && parentClass.getType() != null) {
+                            return acc.contains(parentClass.getType().getFullyQualifiedName());
                         }
                         return false;
                     }
--- a/src/main/java/org/openrewrite/java/migrate/AddStaticVariableOnProducerSessionBean.java
+++ b/src/main/java/org/openrewrite/java/migrate/AddStaticVariableOnProducerSessionBean.java
@@ -20,10 +20,10 @@ import lombok.Setter;
 import org.openrewrite.*;
 import org.openrewrite.internal.ListUtils;
 import org.openrewrite.java.JavaVisitor;
-import org.openrewrite.java.migrate.javax.RemoveEmbeddableId;
 import org.openrewrite.java.search.FindAnnotations;
 import org.openrewrite.java.search.UsesType;
 import org.openrewrite.java.tree.J;
+import org.openrewrite.java.tree.JavaType;
 import org.openrewrite.java.tree.Space;
 import org.openrewrite.marker.Markers;
 import org.openrewrite.xml.XmlVisitor;
@@ -59,9 +59,9 @@ public class AddStaticVariableOnProducerSessionBean extends ScanningRecipe<AddSt
                 String sessionType = null;
                 for (Xml.Tag child : tag.getChildren()) {
                     if ("ejb-class".equals(child.getName())) {
-                        className = child.getValue().toString();
+                        className = child.getValue().orElse(null);
                     } else if ("session-type".equals(child.getName())) {
-                        sessionType = child.getValue().toString();
+                        sessionType = child.getValue().orElse(null);
                     }
                 }
                 if (className != null && sessionType != null) {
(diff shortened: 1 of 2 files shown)


rlsanders4:
@timtebeek Could we also get @karthikNousher added to the IBM team? Thanks!
https://github.com/orgs/openrewrite/teams/ibm

src/main/java/org/openrewrite/java/migrate/AddStaticVariableOnProducerSessionBean.java
```diff
@@ -38,41 +46,89 @@ public String getDescription() {
...
+    }
 
-                    private boolean hasAnnotation(J j, String annotationPattern) {
-                        return !FindAnnotations.find(j, annotationPattern).isEmpty();
-                    }
-                });
+    // @Data
+    @Setter
+    @Getter
+    public static class Accumulator {
```
rlsanders4: We can directly use the Collection object as the accumulator instead of defining a new custom object for this.

src/test/java/org/openrewrite/java/migrate/AddStaticVariableOnProducerSessionBeanTest.java
```diff
@@ -55,6 +56,65 @@ public class SomeDependency {}
...
+                  private static SomeDependency someDependency;
+                  void exampleMethod() {
+                     return;
+                  }
+              }
+              """
+          )
+        );
+    }
+
```
rlsanders4, line 148: We should also add a negative test case for this xml file use case.

src/main/java/org/openrewrite/java/migrate/AddStaticVariableOnProducerSessionBean.java
```diff
@@ -38,41 +46,89 @@ public String getDescription() {
...
+        };
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor(Accumulator acc) {
+        return
+                Preconditions.check(
+                        Preconditions.or(
+                                new UsesType<>("jakarta.enterprise.inject.Produces", false),
+                                Preconditions.or(
```
rlsanders4: This second `Preconditions.or()` statement is not necessary. Checking for the `Produces` type should be sufficient.

src/main/java/org/openrewrite/java/migrate/AddStaticVariableOnProducerSessionBean.java
```diff
@@ -38,41 +46,89 @@ public String getDescription() {
...
-                                    new J.Modifier(Tree.randomId(), Space.SINGLE_SPACE, Markers.EMPTY, null, J.Modifier.Type.Static, emptyList())));
-                        }
-                        return super.visitVariableDeclarations(multiVariable, ctx);
+    public Accumulator getInitialValue(ExecutionContext ctx) {
+        return new Accumulator();
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getScanner(Accumulator acc) {
+        return new XmlVisitor<ExecutionContext>() {
```
rlsanders4: You should add a precondition to this visitor to only visit `ejb-jar.xml` files.

src/main/java/org/openrewrite/java/migrate/AddStaticVariableOnProducerSessionBean.java
```diff
@@ -15,18 +15,26 @@
...
+import java.util.HashMap;
+import java.util.Map;
 
 import static java.util.Collections.emptyList;
 
-public class AddStaticVariableOnProducerSessionBean extends Recipe {
+public class AddStaticVariableOnProducerSessionBean extends ScanningRecipe<AddStaticVariableOnProducerSessionBean.Accumulator> {
     @Override
     public String getDisplayName() {
         return "Adds `static` modifier to `@Produces` field that are on session bean";
```
rlsanders4:
```suggestion
        return "Adds `static` modifier to `@Produces` fields that are in session bean";
```

src/main/java/org/openrewrite/java/migrate/AddStaticVariableOnProducerSessionBean.java
```diff
@@ -38,41 +46,89 @@ public String getDescription() {
...
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getScanner(Accumulator acc) {
+        return new XmlVisitor<ExecutionContext>() {
+            @Override
+            public Xml visitTag(Xml.Tag tag, ExecutionContext ctx) {
+                String className = null;
+                String sessionType = null;
+                for (Xml.Tag child : tag.getChildren()) {
```
rlsanders4: We should only be doing this check for tags matching the following path:
```
ejb-jar/enterprise-beans/session
```
Make sure you are filtering out other tags.

src/main/java/org/openrewrite/java/migrate/AddStaticVariableOnProducerSessionBean.java
```diff
@@ -38,41 +46,89 @@ public String getDescription() {
...
 
-                    private boolean hasAnnotation(J j, String annotationPattern) {
-                        return !FindAnnotations.find(j, annotationPattern).isEmpty();
-                    }
-                });
+    // @Data
+    @Setter
+    @Getter
+    public static class Accumulator {
+        private Map<String, String> sessionBeanClasses = new HashMap<>();
```
rlsanders4: It is unnecessary to use a `Map` for this. You just need to know which classes to update, and you can store that information in a `Set`. There is no need to save the session type.

src/main/java/org/openrewrite/java/migrate/AddStaticVariableOnProducerSessionBean.java
```diff
@@ -38,41 +46,89 @@ public String getDescription() {
...
+                String sessionType = null;
+                for (Xml.Tag child : tag.getChildren()) {
+                    if ("ejb-class".equals(child.getName())) {
+                        className = child.getValue().orElse(null);
+                    } else if ("session-type".equals(child.getName())) {
+                        sessionType = child.getValue().orElse(null);
                     }
+                }
+                if (className != null && sessionType != null) {
+                    acc.getSessionBeanClasses().put(className, sessionType);
```
rlsanders4: As mentioned in another comment, we should just store the classes that we wish to update in a Set here. This is simpler because you don’t have to store as much information.

src/main/java/org/openrewrite/java/migrate/AddStaticVariableOnProducerSessionBean.java
```diff
@@ -21,15 +21,21 @@
...
+import java.util.Map;
 
 import static java.util.Collections.emptyList;
 
-public class AddStaticVariableOnProducerSessionBean extends Recipe {
+public class AddStaticVariableOnProducerSessionBean extends ScanningRecipe<Map<String, String>> {
     @Override
     public String getDisplayName() {
-        return "Adds `static` modifier to `@Produces` field that are on session bean";
+        return "Adds `static` modifier to `@Produces` fields that are in session bean";
```
rlsanders4:
```suggestion
        return "Adds `static` modifier to `@Produces` fields that are in session beans";
```

rlsanders4 (commented):
Hey Karthik, could you look though my comments from the last review? Several did not get addressed, such as:
- Using a `Set` instead of a `Map` as the accumulator
- Checking for `ejb-jar.xml` in a precondition for the scanner (this is more efficient than `visitDocument`)
- Ensuring that the scanner is only executing on tags with this path `ejb-jar/enterprise-beans/session`

src/main/java/org/openrewrite/java/migrate/AddStaticVariableOnProducerSessionBean.java
```diff
@@ -38,22 +48,45 @@ public String getDescription() {
...
+                        String className = null;
+                        if (EJB_PATH.matches(getCursor())) {
+                            for (Xml.Tag child : tag.getChildren()) {
+                                if ("ejb-class".equals(child.getName())) {
+                                    className = child.getValue().orElse(null);
+                                    break;
+                                }
+                            }
+                            if (className != null) {
+                                acc.add(className);
```
rlsanders4: You will need to check the `session` tag  has a child `session-type` tag with one of the expected values before adding it to the accumulator.

src/main/java/org/openrewrite/java/migrate/AddStaticVariableOnProducerSessionBean.java
```diff
@@ -66,13 +99,24 @@ private boolean isInSessionBean() {
...
                         return hasAnnotation(parentClass, "@jakarta.ejb.Singleton") ||
-                                hasAnnotation(parentClass, "@jakarta.ejb.Stateful") ||
-                                hasAnnotation(parentClass, "@jakarta.ejb.Stateless");
+                               hasAnnotation(parentClass, "@jakarta.ejb.Stateful") ||
+                               hasAnnotation(parentClass, "@jakarta.ejb.Stateless");
+                    }
+
+                    private boolean isInXml() {
+                        J.ClassDeclaration parentClass = getCursor().firstEnclosing(J.ClassDeclaration.class);
+                        if (parentClass != null && parentClass.getType() instanceof JavaType.FullyQualified) {
```
rlsanders4: `instanceof` does a null check already.
```suggestion
                        if (parentClass.getType() instanceof JavaType.FullyQualified) {
```

src/main/java/org/openrewrite/java/migrate/AddStaticVariableOnProducerSessionBean.java
```diff
@@ -38,22 +48,45 @@ public String getDescription() {
...
-                        Preconditions.or(
-                                new UsesType<>("jakarta.ejb.Singleton", false),
-                                new UsesType<>("jakarta.ejb.Stateful", false),
-                                new UsesType<>("jakarta.ejb.Stateless", false)
-                        )
-                ),
+                new UsesType<>("jakarta.enterprise.inject.Produces", false),
                 new JavaVisitor<ExecutionContext>() {
                     @Override
                     public J visitVariableDeclarations(J.VariableDeclarations multiVariable, ExecutionContext ctx) {
```
rlsanders4, line 88: There is no need to return super at the end of this visitor if it is the only one.

src/main/java/org/openrewrite/java/migrate/AddStaticVariableOnProducerSessionBean.java
```diff
@@ -38,22 +48,45 @@ public String getDescription() {
...
+    public TreeVisitor<?, ExecutionContext> getScanner(Set<String> acc) {
+        return Preconditions.check(
+                new FindSourceFiles("**/ejb-jar.xml"),
+                new XmlVisitor<ExecutionContext>() {
+
+                    @Override
+                    public Xml visitTag(Xml.Tag tag, ExecutionContext ctx) {
+                        String className = null;
+                        if (EJB_PATH.matches(getCursor())) {
+                            for (Xml.Tag child : tag.getChildren()) {
```
rlsanders4: This will work. But instead of iterating through all the child tags, it is easier to just use `tag.getChild(“tagName”)` to get the desired child tag.

rlsanders4 (approved):
Looks good to me. I think you are ready to have someone from the OpenRewrite team take a look.

PR #768, opened 2025-06-18
Fix `LombokValToFinalVar` in loops

MBoegers:
## What's changed?
`LombokValToFinalVar` now avoids useing `JavaTemplate` in for each loops were type attribution is not secure enough and parsing might fail.

## What's your motivation?
fix handling in for each loops

We assumed that `for(var s : strings)` is better migrated to `for(val s : strings)`than `for(final var s : strings)` as it more comply with the common usage.

Fixing parser issue in template logic connected to `val` in loops, see 
- https://github.com/openrewrite/rewrite/pull/5637

### Checklist

### DIFF
--- a/src/test/java/org/openrewrite/java/migrate/lombok/LombokValToFinalVarTest.java
+++ b/src/test/java/org/openrewrite/java/migrate/lombok/LombokValToFinalVarTest.java
@@ -125,23 +125,19 @@ class LombokValToFinalVarTest implements RewriteTest {
                   """
                     import lombok.val;
 
-                    import java.util.ArrayList;
                     import java.util.List;
 
                     class A {
-                        void bar() {
-                            List<String> lst = new ArrayList<>();
+                        void bar(List<String> lst) {
                             for (val s : lst) {}
                         }
                     }
                     """,
                   """
-                    import java.util.ArrayList;
                     import java.util.List;
 
                     class A {
-                        void bar() {
-                            List<String> lst = new ArrayList<>();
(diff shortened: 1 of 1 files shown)


src/test/java/org/openrewrite/java/migrate/lombok/LombokValToFinalVarTest.java
```diff
@@ -15,8 +15,11 @@
  */
 package org.openrewrite.java.migrate.lombok;
 
+import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
+import org.junitpioneer.jupiter.ExpectedToFail;
```
timtebeek:
```suggestion
```

src/test/java/org/openrewrite/java/migrate/lombok/LombokValToFinalVarTest.java
```diff
@@ -15,8 +15,10 @@
  */
 package org.openrewrite.java.migrate.lombok;
 
+import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
 import org.openrewrite.DocumentExample;
+import org.openrewrite.Issue;
```
timtebeek:
```suggestion
```

timtebeek:
This is now fully disconnected from the potential underlying issue we could separately explore in
- https://github.com/openrewrite/rewrite/pull/5637

As it stands we're not changing the shape of things in this recipe now; merely fixing an issue that was seen at a customer site.

PR #769, opened 2025-06-19
Only add null case when all values covered

Jenson3210:
## What's changed?
We now only add the null case if all possible input values are covered to avoid compilation issues

## What's your motivation?
Missed this in the original PR casuign compilation to break.

@knutwannheden 

### Checklist

### DIFF
--- a/src/test/java/org/openrewrite/java/migrate/lang/NullCheckAsSwitchCaseTest.java
+++ b/src/test/java/org/openrewrite/java/migrate/lang/NullCheckAsSwitchCaseTest.java
@@ -428,6 +428,61 @@ class NullCheckAsSwitchCaseTest implements RewriteTest {
         );
     }
 
+    @Test
+    void mergeWhenExistingSwitchIsCoveringAllPossibleEnumValues() {
+        rewriteRun(
+          java(
+            """
+              package suits;
+              public enum Suit {
+                  CLUBS, DIAMONDS, HEARTS, SPADES, JOKER
+              }
+              """
+          ),
+          //language=java
+          java(
+            """
+              import suits.Suit;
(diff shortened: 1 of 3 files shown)


src/test/java/org/openrewrite/java/migrate/lang/NullCheckAsSwitchCaseTest.java
```diff
@@ -72,6 +75,64 @@ static String score(String obj) {
         );
     }
 
+    @Test
+    @DocumentExample
```
timtebeek: Try to use just one representative example in the docs, otherwise those pages will grow long quickly.
```suggestion
```

src/test/java/org/openrewrite/java/migrate/lang/NullCheckAsSwitchCaseTest.java
```diff
@@ -672,5 +788,61 @@ public String getObj() {
             );
         }
 
+        @Test
+        void doNotMergeWhenExistingSwitchIsNotCoveringAllPossibleInputValuesWithDefault() {
```
timtebeek, line 873: I don't see a default here? Or do I misunderstand the test method name?

Jenson3210: meant to say notCoveredWithDefault as it is not covering all input values as there is no default.

Jenson3210:
Validated with cli that the null branch is no longer (falsely) being created with the metacat repo

pdelagrave:
Will fix #785 and #786

timtebeek (approved):

Issue #571, opened 2024-10-02
`org.openrewrite.java.migrate.util.MigrateCollectionsSingletonList` should not convert known null


blipper:
## What is the smallest, simplest way to reproduce the problem?

```java
import java.util.Collections;

class A {
    void foo( {
     Collections.singletonList(null);
    }
}
```

## What did you expect to see?

```java
import java.util.Collections;

class A {
    void foo( {
     Collections.singletonList(null);
    }
}
```

## What did you see instead?

```java

class A {
    void foo( {
     List.of(null);
    }
}

```

Really this should be any provably null or if conservative only provably non-null.  

This is a problem because singleList allows null element.
https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/List.html#of(E)
https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/Collections.html#singletonList(T)

PR #772, opened 2025-06-20
feat: Prevent migration of singletonMap with null value or null key

ZzetT:
## What's changed?
The rule `org.openrewrite.java.migrate.util.MigrateCollectionsSingletonMap` also migrated the following cases:
```java
Collections.singletonMap(null, "foo")
Collections.singletonMap("bar", null)
```
to
```java
Map.of(null, "foo")
Map.of("bar", null)
```
which leads to a NullPointerException

## What's your motivation?
There is a closed issue https://github.com/openrewrite/rewrite-migrate-java/issues/571 where it was even mentioned in the comments to apply the same to maps, but the changes were not done in the previous PR.

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/util/MigrateCollectionsSingletonMap.java
+++ b/src/main/java/org/openrewrite/java/migrate/util/MigrateCollectionsSingletonMap.java
@@ -70,8 +70,8 @@ public class MigrateCollectionsSingletonMap extends Recipe {
             }
 
             private boolean isNotLiteralNull(J.MethodInvocation m) {
-                return !J.Literal.isLiteralValue(m.getArguments().get(0), null) &&
-                       !J.Literal.isLiteralValue(m.getArguments().get(1), null);
+                return !(J.Literal.isLiteralValue(m.getArguments().get(0), null) ||
+                         J.Literal.isLiteralValue(m.getArguments().get(1), null));
             }            
         });
     }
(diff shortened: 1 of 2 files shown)


src/main/java/org/openrewrite/java/migrate/util/MigrateCollectionsSingletonMap.java
```diff
@@ -68,6 +68,13 @@ public J visitMethodInvocation(J.MethodInvocation method, ExecutionContext ctx)
 
                 return m;
             }
+
+            private boolean isNotLiteralNull(J.MethodInvocation m) {
+                return !(m.getArguments().get(0) instanceof J.Literal &&
+                         ((J.Literal) m.getArguments().get(0)).getValue() == null ||
+                         m.getArguments().get(1) instanceof J.Literal &&
+                         ((J.Literal) m.getArguments().get(1)).getValue() == null);
```
timtebeek: I think we can make this slightly shorter and more expressive using
https://github.com/openrewrite/rewrite/blob/9abac8d9b5bd5bd4953569290b6591e91ed130cf/rewrite-java/src/main/java/org/openrewrite/java/tree/J.java#L3721-L3727

```suggestion
                return !J.Literal.isLiteralValue(m.getArguments().get(0), null) &&
                       !J.Literal.isLiteralValue(m.getArguments().get(1), null);
```

timtebeek:
All good, thanks! I've applied those with a recipe, and that'll be in our PR checks going forward. :)

PR #775, opened 2025-06-23
Some compilation issues are occuring after running switch pattern matching recipe

Jenson3210:
## What's changed?
If any of the branches returns a value, we do (for now) not convert it to a switch as it is invalid. 
Besides this change: If any of the values throws, we do allow this. 
We used to unwrap if-statements if they were the only one -> no longer happening as only expressions and Throws are allowed to be written in that syntax.

I know we can enhance to do return from switch but I first want to stabilize the existing recipe before working on: 
- https://github.com/openrewrite/rewrite-migrate-java/issues/776

## What's your motivation?
Clean up compilation issues occuring during run of the recipe.

@timtebeek 

### Checklist


timtebeek (approved):
Thanks for the isolated fixes here! Makes reviewing a lot easier.

PR #782, opened 2025-07-04
Keep `jakarta.annotation-api` dependency when moving to Jakarta with Spring Boot project and null annotations are used

jevanlingen:
## What's changed?
For Spring Boot projects the `jakarta.annotation-api` dependency will not be removed when @Nonnull or/and @Nullable is used.

## What's your motivation?
We remove that dependency for Spring Boot, because with the dependencies Spring bring in, there is no need for the `jakarta.annotation-api` dependency. But when the nullable annotations this dependency provides are used, we cannot clean them up as that would result into uncompilable code.

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/jakarta/HasNoJakartaNullAnnotations.java
+++ b/src/main/java/org/openrewrite/java/migrate/jakarta/HasNoJakartaNullAnnotations.java
@@ -1,0 +1,61 @@
+package org.openrewrite.java.migrate.jakarta;
+
+import org.jspecify.annotations.Nullable;
+import org.openrewrite.ExecutionContext;
+import org.openrewrite.ScanningRecipe;
+import org.openrewrite.Tree;
+import org.openrewrite.TreeVisitor;
+import org.openrewrite.java.JavaIsoVisitor;
+import org.openrewrite.java.search.FindAnnotations;
+import org.openrewrite.java.tree.J;
+import org.openrewrite.marker.SearchResult;
+
+import java.util.concurrent.atomic.AtomicBoolean;
+
+public class HasNoJakartaNullAnnotations extends ScanningRecipe<AtomicBoolean> {
+    @Override
+    public String getDisplayName() {
+        return "Project has no Jakarta null annotations";
+    }
+
+    @Override
+    public String getDescription() {
+        return "Search for @Nonnull and @Nullable annotations, mark all source as found if no annotations are found.";
+    }
+
+    @Override
+    public AtomicBoolean getInitialValue(ExecutionContext ctx) {
+        return new AtomicBoolean();
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getScanner(AtomicBoolean acc) {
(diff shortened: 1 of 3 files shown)


src/test/java/org/openrewrite/java/migrate/jakarta/JavaxToJakartaTest.java
```diff
@@ -574,6 +584,170 @@ public class TestApplication {
...
+          spec -> spec.parser(JavaParser.fromJavaVersion().dependsOn(javaxServlet, jakartaAnnotation)),
+          mavenProject(
+            "Sample",
+            //language=xml
+            pomXml(
+              """
+                <?xml version="1.0" encoding="UTF-8"?>
+                <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+                    <modelVersion>4.0.0</modelVersion>
```
timtebeek:
```suggestion
                <project>
```

src/main/java/org/openrewrite/java/migrate/jakarta/HasNoJakartaNullAnnotations.java
```diff
@@ -0,0 +1,76 @@
...
+
+    @Override
+    public String getDescription() {
+        return "Search for @Nonnull and @Nullable annotations, mark all source as found if no annotations are found.";
+    }
+
+    @Override
+    public AtomicBoolean getInitialValue(ExecutionContext ctx) {
+        return new AtomicBoolean();
+    }
```
timtebeek: Would this be another case where we'd want ``Map<JavaProject, Boolean>` or even `Set<JavaProject>` to account for multi module projects? Now it's a boolean across all modules, whereas I think we'd want to evaluate per JavaProject.

jevanlingen: Yes you are right. Let me reimplement it exactly like [ModuleHasPlugin](https://github.com/openrewrite/rewrite/blob/facbcfb3da944b3b08dc6bab41c5d5a478254c29/rewrite-maven/src/main/java/org/openrewrite/maven/search/ModuleHasPlugin.java)! As it's nearly the same recipe, but with a different condition.

jevanlingen: https://github.com/openrewrite/rewrite-migrate-java/pull/782/commits/5819247fd23f340570a857c83b4f9d75ce92898a

timtebeek (approved):
As discussed this should already see some improvement for the specific case reported, but perhaps even better when we only remove the Jakarta annotations if available transitively, as that's likely what was intended earlier and should avoid needlessly lingering explicit dependencies.

We'll merge as is though to get this fix out sooner rather than wait for larger changes to some core recipes.

PR #787, opened 2025-07-06
feat: add recipe that prefers guava ranges

timo-a:
## What's changed?

adds recipe that converts `from.compareTo(candidate) <= 0 && candidate.compareTo(to) <= 0` into more idiomatic guava ranges.
Covers opened, closed and half-open intervals of `Comparable`s and primitives.

## What's your motivation?

Make code more readable

1) I chose `candidate` as a name for the object that is tested for its membership. Maybe there is a better name.
2) `from <= candidate && candidate <= to` is fairly easy to read. For uniformity, I chose to have it converted to `Range.closed(from, to).contains(candidate)` anyway.

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/guava/UseRanges.java
+++ b/src/main/java/org/openrewrite/java/migrate/guava/UseRanges.java
@@ -27,13 +27,11 @@ import org.openrewrite.java.template.RecipeDescriptor;
         description = "Simplifies hand crafted range checks."
 )
 public class UseRanges {
-
     @RecipeDescriptor(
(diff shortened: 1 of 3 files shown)


timtebeek (commented):
Perfect use case for Refaster here @timo-a ; thanks a lot for adding these. Did you already try them out in detail on your projects?

I'll need to take a closer look still at the various open/closed vs `<=` operators used, but should be fairly straightforward to compare and get merged after that.

timtebeek:
Another option would be to push these refaster recipes to ErrorProne Support, and then we can pick those up here. You'll notice I follow that pattern from time to time myself as well:
- https://github.com/PicnicSupermarket/error-prone-support/releases/tag/v0.24.0

timo-a:
Thanks, I'll look into that!

PR #788, opened 2025-07-07
Create replace for `@InlineMe` annotations

timtebeek:
### Checklist


timtebeek:
Replicated with 
- https://github.com/openrewrite/rewrite/pull/5923

knutwannheden (approved):
Nice! I am not sure if we can change the TypesInUse like that, but perhaps we also find a different solution. For now I think wrapping in a declarative recipe with a precondition is quite sufficient.

PR #789, opened 2025-07-08
Replace JavaTemplate by a `var` identifier element for UseVarForGenericsConstructors recipe

jevanlingen:
## What's changed?
No functional change, but a refactoring / bugfix. The JavaTemplate code has been replaced by a `var` identifier element.

## What's your motivation?
In rare cases, the JavaTemplate did not have the proper types available, and thus would throw errors like:

```diff
- Set<Long> x = new HashSet<>();
+ {{80000000-0000-0d0d-0000-000000000000}}Set<Long> x = new HashSet<>(){{80000000-0000-0d0d-0000-000000000000}};
```

## Any additional context
Working around the JavaTemplate by not using it all could be frowned upon. In this case it's an improvement, as the modifier and whitespace restoration are not needed anymore.

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/lang/var/UseVarForGenericsConstructors.java
+++ b/src/main/java/org/openrewrite/java/migrate/lang/var/UseVarForGenericsConstructors.java
@@ -17,17 +17,21 @@ package org.openrewrite.java.migrate.lang.var;
 
 import org.jspecify.annotations.Nullable;
 import org.openrewrite.*;
+import org.openrewrite.internal.ListUtils;
 import org.openrewrite.java.JavaIsoVisitor;
-import org.openrewrite.java.JavaParser;
-import org.openrewrite.java.JavaTemplate;
 import org.openrewrite.java.search.UsesJavaVersion;
 import org.openrewrite.java.tree.*;
 import org.openrewrite.marker.Markers;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Objects;
 
 import static java.util.Collections.emptyList;
+import static java.util.Collections.singleton;
+import static java.util.stream.Collectors.toList;
+import static org.openrewrite.Tree.randomId;
+import static org.openrewrite.java.tree.Space.EMPTY;
 
 public class UseVarForGenericsConstructors extends Recipe {
     @Override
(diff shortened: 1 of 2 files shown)


src/main/java/org/openrewrite/java/migrate/lang/var/UseVarForGenericsConstructors.java
```diff
@@ -145,108 +135,36 @@ private static boolean hasBounds(JavaType type) {
...
-                for (JavaType curType : typeParameters) {
-                    typeParamsExpression.add(JRightPadded.build(typeToExpression(curType)));
-                }
-
-                NameTree clazz = new J.Identifier(Tree.randomId(), Space.EMPTY, Markers.EMPTY, emptyList(), ((JavaType.Parameterized) type).getClassName(), null, null);
-                return new J.ParameterizedType(Tree.randomId(), Space.EMPTY, Markers.EMPTY, clazz, JContainer.build(typeParamsExpression), type);
-            }
-
-            throw new IllegalArgumentException(String.format("Unable to parse expression from JavaType %s", type));
+            J.NewClass initializer = Objects.requireNonNull((J.NewClass) vd.getVariables().get(0).getInitializer());
```
MBoegers, line 146: Do we have evidence that the other recipes around local variable type inference are effected by this as well? If so we should align here

src/test/java/org/openrewrite/java/migrate/lang/var/UseVarForGenericsConstructorsTest.java
MBoegers, line 1: Why remove the Java version 10 constraint in the tests?

jevanlingen: Because it was already done generally: 

https://github.com/openrewrite/rewrite-migrate-java/blob/01384fb73fedff082b775f32163f1891c94022ec/src/test/java/org/openrewrite/java/migrate/lang/var/UseVarForGenericsConstructorsTest.java#L29-L32

No need to this twice 😉. The `version` function is specially meant to set you source code to java x, if you do _not_ set all the source files to a specific version generally.

src/main/java/org/openrewrite/java/migrate/lang/var/UseVarForGenericsConstructors.java
jevanlingen: @MBoegers and me talked about this internally, and decided the other recipes should also be looked upon 😄.

src/test/java/org/openrewrite/java/migrate/lang/var/UseVarForGenericsConstructorsTest.java
MBoegers: we still have a test that assures we skip Java 7 f.e. right?

jevanlingen: Good one, there wasn't actually a test that proved this. So I added a test quickly: https://github.com/openrewrite/rewrite-migrate-java/pull/789/commits/5ef73d3d52a3762cd8f3fa3835d478f9d4bd53d2!

_Besides the missing test, you can see the beauty of the version function now; so all java files are java 10, except this one 😝._

Issue #286, opened 2023-09-08
Migrate from `*jdk15on` bouncycastle packages to `*jdk18on` or `*jdk15to18` 


yeikel:
## What problem are you trying to solve?

> 
> From : https://www.bouncycastle.org/latest_releases.html
> 
> With the arrival of Java 15. jdk15 is not quite as unambiguous as it was. The jdk18on jars are compiled to work with anything from Java 1.8 up. They are also multi-release jars so do support some features that were introduced in Java 9, Java 11, and Java 15. If you have issues with multi-release jars see the jdk15to18 release jars below.
> 
> Also
> 
> BC 1.71 changed the jdk15on jars to jdk18on so the base has now moved to Java 8. For earlier JVMs, or containers/applications that cannot cope with multi-release jars, you should now use the jdk15to18 jars.

> Further Note (users of Oracle JVM 1.7 or earlier, users of "pre-Java 9" toolkits): As of 1.63 we have started including signed jars for "jdk15to18", if you run into issues with either signature validation in the JCE or the presence of the multi-release versions directory in the regular "jdk18on" jar files try the "jdk15to18" jars instead. Please also note the JCE certificate in the public access versions of Oracle Java 6 (6u45) and Oracle Java 7 (7u80) is expired on the 20th April 2021. We still counter sign the jdk15to18 jars with this certificate for compatibility reasons, but Oracle does distribute JVMs for Java 6 and Java 7 with a newer, and stronger, certificate to holders of Java Support Contracts.

## What precondition(s) should be checked before applying this recipe?

- groupId `org.bouncycastle`  
- artifactId `*-jdk15on`
- Project is using Java 8 and up

## Describe the situation before applying the recipe
```xml

             <dependency>
            <groupId>org.bouncycastle</groupId>
            <artifactId>bcprov-jdk15on</artifactId>
           <version>1.70</version>
        </dependency>
        <dependency>
            <groupId>org.bouncycastle</groupId>
            <artifactId>bcpkix-jdk15on</artifactId>
           <version>1.70</version>
        </dependency>

```

## Describe the situation after applying the recipe

```diff

diff --git a/pom.xml
index 1579251e4..2519c4a40 100644
--- a/b/pom.xml
+++ b/c/pom.xml
@@ -215,12 +215,12 @@
         <dependency>
             <groupId>org.bouncycastle</groupId>
-            <artifactId>bcprov-jdk15on</artifactId>
[...]

PR #790, opened 2025-07-08
Add Bouncy Castle upgrade recipe for Java < 8

pdelagrave:
## What's changed?
A new recipe to replace the Bouncycastle artifacts from `jdk15on` to `jdk15to18`. `jdk15on` isn't maintained anymore and `jdk18on` is only for Java 8 and above. The `jdk15to18` artifact is the up-to-date replacement of the unmaintained `jdk15on` for java < 8

## What's your motivation?
- Fix https://github.com/openrewrite/rewrite-migrate-java/issues/286

### Existing recipe
The recipe to migrate from `-jdk15on` to `-jdk18on` [was added 3 months ago](https://github.com/openrewrite/rewrite-migrate-java/blob/main/src/main/resources/META-INF/rewrite/bouncycastle-jdk18on.yml). That existing recipe assumes the target code is >= java8.

I've copied over the existing recipe YAML and changed the versions and description, and copied the main test.

Because there won't be more of that issue with bouncycastle anytime soon if ever, I don't think it's worth investing to deduplicate that code, i.e. options support for declarative recipes.

### Precondition
I couldn't find a way to reuse existing precondition recipes to restrict this new recipe to only java < 8.

### Checklist

### DIFF
--- a/src/main/resources/META-INF/rewrite/bouncycastle-jdk15to18.yml
+++ b/src/main/resources/META-INF/rewrite/bouncycastle-jdk15to18.yml
@@ -17,9 +17,11 @@
 ---
 type: specs.openrewrite.org/v1beta/recipe
 name: org.openrewrite.java.migrate.BouncyCastleFromJdk15OnToJdk15to18
-displayName: Migrate BouncyCastle from jdk15on to jdk15to18 for Java < 8
+displayName: Migrate Bouncy Castle from `jdk15on` to `jdk15to18` for Java < 8
 description: >-
-  This recipe replaces the Bouncycastle artifacts from `jdk15on` to `jdk15to18`. `jdk15on` isn't maintained anymore and `jdk18on` is only for Java 8 and above. The `jdk15to18` artifact is the up-to-date replacement of the unmaintained `jdk15on` for java < 8.
+  This recipe replaces the Bouncy Castle artifacts from `jdk15on` to `jdk15to18`.
+  `jdk15on` isn't maintained anymore and `jdk18on` is only for Java 8 and above.
+  The `jdk15to18` artifact is the up-to-date replacement of the unmaintained `jdk15on` for Java < 8.
 tags:
   - bouncycastle
 recipeList:
--- a/src/main/resources/META-INF/rewrite/bouncycastle-jdk15to18.yml
+++ b/src/main/resources/META-INF/rewrite/bouncycastle-jdk15to18.yml
@@ -1,0 +1,60 @@
+#
+# Copyright 2025 the original author or authors.
+# <p>
+# Licensed under the Moderne Source Available License (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


src/main/resources/META-INF/rewrite/bouncycastle-jdk15to18.yml
```diff
@@ -0,0 +1,60 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+---
+type: specs.openrewrite.org/v1beta/recipe
+name: org.openrewrite.java.migrate.BouncyCastleFromJdk15OnToJdk15to18
```
timtebeek, line 19: I propose we include this by default in our Java 7 migration recipe:
https://github.com/openrewrite/rewrite-migrate-java/blob/2f005afaad1f70f2f37fdfc1a0cc9ec3fb6365f1/src/main/resources/META-INF/rewrite/java-version-7.yml#L19-L30

To mirror what we already do for version 8+:
https://github.com/openrewrite/rewrite-migrate-java/blob/d04102342d482a126de498c1549c17ab64edb191/src/main/resources/META-INF/rewrite/java-version-8.yml#L29

src/main/resources/META-INF/rewrite/bouncycastle-jdk15to18.yml
```diff
@@ -0,0 +1,60 @@
...
+displayName: Migrate BouncyCastle from jdk15on to jdk15to18 for Java < 8
+description: >-
+  This recipe replaces the Bouncycastle artifacts from jdk15on to jdk15to18. jdk15on isn't maintained anymore and jdk18on is only for Java 8 and above. The jdk15to18 artifact is the up-to-date replacement of the unmaintained jdk15on for java < 8.
+tags:
+  - bouncycastle
+recipeList:
+  - org.openrewrite.java.dependencies.ChangeDependency:
+      oldGroupId: org.bouncycastle
+      oldArtifactId: bcprov-jdk15on
+      newArtifactId: bcprov-jdk15to18
```
timtebeek, line 31: I think we should also update our recipes for Java 8+ to migrate from `jdk15to18` to `jdk18on`, where those now only migrate `jdk15on`.
https://github.com/openrewrite/rewrite-migrate-java/blob/ec56d0be592c4e95b5fc2eada26c55842b7b0c52/src/main/resources/META-INF/rewrite/bouncycastle-jdk18on.yml#L26-L30

timtebeek (changes requested):
Two suggestions to make these recipes usable from larger composite recipes, and not something that folks need to explicitly seek out and run separately.

timtebeek (approved):
Thanks for working through the feedback so quickly!

PR #793, opened 2025-07-09
Let all change-to-var recipes use a created var-Identifier instead of using a JavaTemplate

jevanlingen:
## What's changed?
Nothing, just refactoring.

## What's your motivation?
- https://github.com/openrewrite/rewrite-migrate-java/pull/789#discussion_r2194267438

## Any additional context
The `UseVarForGenericMethodInvocations` recipe is actually not complete of yet. I didn't want to change any logic, so I wrote a small incomplete implementation in comments.

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/lang/var/UseVarForGenericMethodInvocations.java
+++ b/src/main/java/org/openrewrite/java/migrate/lang/var/UseVarForGenericMethodInvocations.java
@@ -84,7 +84,8 @@ public class UseVarForGenericMethodInvocations extends Recipe {
 
             return DeclarationCheck.transformToVar(vd);
             // TODO implement to support cases like `var strs = List.<String>of();`
-            /*return DeclarationCheck.<J.MethodInvocation>transformToVar(vd, it -> {
+            /*J.VariableDeclarations finalVd = vd;
+            return DeclarationCheck.<J.MethodInvocation>transformToVar(vd, it -> {
                 // if left is defined but not right, copy types to initializer
                 if (finalVd.getTypeExpression() instanceof J.ParameterizedType && !((J.ParameterizedType) finalVd.getTypeExpression()).getTypeParameters().isEmpty() && it.getTypeParameters() == null) {
                     return it.withTypeParameters(((J.ParameterizedType) finalVd.getTypeExpression()).getPadding().getTypeParameters());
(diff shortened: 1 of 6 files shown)


src/main/java/org/openrewrite/java/migrate/lang/var/UseVarForGenericMethodInvocations.java
```diff
@@ -62,34 +56,42 @@ public J.VariableDeclarations visitVariableDeclarations(J.VariableDeclarations v
...
+            return DeclarationCheck.transformToVar(vd);
+            // TODO implement to support cases like `var strs = List.<String>of();`
+            /*J.VariableDeclarations finalVd = vd;
+            return DeclarationCheck.<J.MethodInvocation>transformToVar(vd, it -> {
+                // If left has generics but right has not, copy types parameters
+                if (finalVd.getTypeExpression() instanceof J.ParameterizedType && !((J.ParameterizedType) finalVd.getTypeExpression()).getTypeParameters().isEmpty() && it.getTypeParameters() == null) {
+                    return it.withTypeParameters(((J.ParameterizedType) finalVd.getTypeExpression()).getPadding().getTypeParameters());
+                }
+                return it;
+            });*/
```
Laurens-W, line 94: This regression is not captured by a unit test

jevanlingen: This is not a regression, but the implementation of:
> The UseVarForGenericMethodInvocations recipe is actually not complete of yet. I didn't want to change any logic, so I wrote a small incomplete implementation in comments.

There was some code which seem to do something:

https://github.com/openrewrite/rewrite-migrate-java/blob/775fcf23262a553993d6c0266e4c9600a9ca0610/src/main/java/org/openrewrite/java/migrate/lang/var/UseVarForGenericMethodInvocations.java#L108-L117

But it was never ever used, because the calling part always supplied two empty lists:

https://github.com/openrewrite/rewrite-migrate-java/blob/775fcf23262a553993d6c0266e4c9600a9ca0610/src/main/java/org/openrewrite/java/migrate/lang/var/UseVarForGenericMethodInvocations.java#L92

Thus there is no way it ever did anything. I created some code to resemble that "there was something". Then I thought, to make sure this is never called (that would be a regression 😉), let's put in comments.

_N.B. Normally I don't really like code in comments, but in this case I thought I made an exception, as it shows the migration from old->new a little bit better. Plus, if someone wants to step up and improve this recipe, the code in comments already points into the right direction._

PR #795, opened 2025-07-14
Convert assigning Switch statements to Switch expressions

pdelagrave:
## What's changed?
A new recipe to convert switch statements for which all cases are doing an assignment to the same variable into a switch expression that is assigned to that variable:

From:
```java
       String formatted = "initialValue";
       switch (obj) {
            case Integer i -> formatted = String.format("int %d", i);
            case Long l -> formatted = String.format("long %d", l);
            default -> formatted = "unknown";
       }
```
To:
```java
        String formatted = switch (obj) {
            case Long l -> String.format("long %d", l);
            default -> "unknown";
        }
```

## What's your motivation?
- Closes https://github.com/openrewrite/rewrite-migrate-java/issues/718

### Checklist


Laurens-W:
Appreciate the draft with extensive tests here!
Perhaps good to disable some of the more extensive tests for now and start work on converting the simple colon/arrow into a single statement/expression cases!

src/main/java/org/openrewrite/java/migrate/lang/SwitchUtils.java
```diff
@@ -0,0 +1,65 @@
...
+ */
+package org.openrewrite.java.migrate.lang;
+
+import org.openrewrite.java.tree.*;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+
+public class SwitchUtils {
```
timtebeek: Where possible let's try to mimize the publicly exposed API. In this case I'd expect all usages to be within this package.
```suggestion
class SwitchUtils {
```

src/main/resources/META-INF/rewrite/java-version-21.yml
```diff
@@ -145,3 +145,4 @@ recipeList:
   - org.openrewrite.java.migrate.lang.IfElseIfConstructToSwitch
   - org.openrewrite.java.migrate.lang.RefineSwitchCases
   - org.openrewrite.java.migrate.lang.SwitchCaseEnumGuardToLabel
+  - org.openrewrite.java.migrate.lang.SwitchCaseAssigningToSwitchExpression
```
timtebeek: This isn't necessarily related to pattern matching right? It might fit better under UpgradeToJava21 just below these lines:
https://github.com/openrewrite/rewrite-migrate-java/blob/d8830aeebb3e8b2887c77e4a2c144d9ec69cb173/src/main/resources/META-INF/rewrite/java-version-21.yml#L42-L43

pdelagrave:
Running the recipe on the Apache repos provided feedback leading to this list of fixes:
- [ ] Don't add a `default` case if the switch is already exhaustive.
- [ ] Ensure comments aren't lost around the modified code.
- [ ] The java version precondition can be set to java 17 and not 21.
- [ ] Do not apply the recipe if a case statement/body assignment is referencing the original variable. example: `case 55: foo = foo.bar()`. One goal of the recipe is to remove the original variable initialization so it would fail.
- [ ] Do not apply the recipe if the original variable initializer is anything else than a literal, variable reference, field access or an expression composed of them only. Anything else will execute code and can have side effects. One of the goal of that recipe is to either remove that initializer or move in as a `default:` case, in which case the code would end up not being executed ever or in the wrong order.
- [ ] Do not apply the recipe if a colon-switch label has an empty statement list AND is the last case of the switch. While that switch can be 'exhaustive', the last case of the switch being empty would mean there's no assignment and this isn't a valid scenario for the recipe.
- [ ] Do not apply the recipe if the original variable has no initializer and the switch isn't exhaustive and each case isn't assigning the original variable a value as expected. This is because we can't add a `default` case to make the switch exhaustive as we don't have the initializer from the original variable declaration. 
- [ ] Optionally add a feature where the migrated switch expression is inlined if the next statement is `return variable` where the variable is the same as the switch expression is being assigned to.

src/main/resources/META-INF/rewrite/java-version-21.yml
pdelagrave: Right, unsure why I've put it there in the first place.
I've actually moved it to the java17 migration recipe as all the feature this new recipe is about have been added before/by java 17.

src/test/java/org/openrewrite/java/migrate/lang/SwitchCaseAssigningToSwitchExpressionTest.java
```diff
@@ -0,0 +1,657 @@
...
+                      return;
+                  }
+              }
+              """
+          ));
+    }
+
+    @Test
+    @ExpectedToFail
+    void failsToFormatWithASpaceWhenOriginalVariableHasNoInitializer() {
```
pdelagrave: @timtebeek Would appreciate a pointer on how to fix that formatting bug, this is the result when `formatted` doesn't have an initializer:
```
String formatted= switch (1) {
```

timtebeek: The google review bot sometimes gets confused by small code changes; in those cases it's best to look at the underlying .patch file that is correct, and had wanted you to swap these annotation to sort them.

timtebeek:
Figured we could reduce the complexity here by spinning off a separate recipe to inline the return:
- https://github.com/openrewrite/rewrite-static-analysis/pull/647

src/test/java/org/openrewrite/java/migrate/lang/SwitchCaseAssigningToSwitchExpressionTest.java
```diff
@@ -0,0 +1,866 @@
...
+                  }
+
+                  void exhaustiveButCantAddDefaultAfterEmptyLabel(TrafficLight light) {
+                      String status = "initialValue";
+                      switch (light) {
+                          case RED: status = "stop"; break;
+                          case GREEN: status = "go"; break;
+                          case YELLOW:
+                      }
+                  }
```
pdelagrave: @timtebeek
While the switch is already exhaustive, wouldn't still adding `default: yield "initialValue";` at the end be safe and preserve the logic in any situation, given the last case is empty?
This would require a fix in the code but that new test made me wonder why not.

pdelagrave:
> Ran it against some 600 local LSTs; only saw results for a single repository: apache/maven
> 
> ```diff
> diff --git a/impl/maven-cli/src/main/java/org/apache/maven/cling/logging/impl/LogbackConfiguration.java b/impl/maven-cli/src/main/java/org/apache/maven/cling/logging/impl/LogbackConfiguration.java
> index 4481486..da4519e 100644
> --- a/impl/maven-cli/src/main/java/org/apache/maven/cling/logging/impl/LogbackConfiguration.java
> +++ b/impl/maven-cli/src/main/java/org/apache/maven/cling/logging/impl/LogbackConfiguration.java
> @@ -30,20 +30,13 @@ org.openrewrite.java.migrate.lang.SwitchCaseAssigningToSwitchExpression
>  public class LogbackConfiguration extends BaseSlf4jConfiguration {
>      @Override
>      public void setRootLoggerLevel(Level level) {
> -        ch.qos.logback.classic.Level value;
> -        switch (level) {
> -            case DEBUG:
> -                value = ch.qos.logback.classic.Level.DEBUG;
> -                break;
> +        ch.qos.logback.classic.Level value = switch (level) {
> +            case DEBUG: yield ch.qos.logback.classic.Level.DEBUG;
>  
> -            case INFO:
> -                value = ch.qos.logback.classic.Level.INFO;
> -                break;
> +            case INFO: yield ch.qos.logback.classic.Level.INFO;
>  
> -            default:
> -                value = ch.qos.logback.classic.Level.ERROR;
> -                break;
> -        }
> +            default: yield ch.qos.logback.classic.Level.ERROR;
> +        };
>          ((ch.qos.logback.classic.Logger) LoggerFactory.getLogger(Logger.ROOT_LOGGER_NAME)).setLevel(value);
>      }
>  
> diff --git a/impl/maven-cli/src/main/java/org/apache/maven/cling/logging/impl/Log4j2Configuration.java b/impl/maven-cli/src/main/java/org/apache/maven/cling/logging/impl/Log4j2Configuration.java
> index 1f02e57..07e84c8 100644
> --- a/impl/maven-cli/src/main/java/org/apache/maven/cling/logging/impl/Log4j2Configuration.java
> +++ b/impl/maven-cli/src/main/java/org/apache/maven/cling/logging/impl/Log4j2Configuration.java
> @@ -28,20 +28,13 @@ org.openrewrite.java.migrate.lang.SwitchCaseAssigningToSwitchExpression
>  public class Log4j2Configuration extends BaseSlf4jConfiguration {
>      @Override
>      public void setRootLoggerLevel(Level level) {
> -        String value;
> -        switch (level) {
> -            case DEBUG:
> -                value = "debug";
> -                break;
> +        String value = switch (level) {
> +            case DEBUG: yield "debug";
>  
> -            case INFO:
> -                value = "info";
> -                break;
> +            case INFO: yield "info";
>  
> -            default:
> -                value = "error";
> -                break;
> -        }
> +            default: yield "error";
> +        };
>          System.setProperty("maven.logging.root.level", value);
>      }
>  
> diff --git a/impl/maven-cli/src/main/java/org/apache/maven/cling/logging/impl/MavenSimpleConfiguration.java b/impl/maven-cli/src/main/java/org/apache/maven/cling/logging/impl/MavenSimpleConfiguration.java
> index 5147cd8..0c5f4e2 100644
> --- a/impl/maven-cli/src/main/java/org/apache/maven/cling/logging/impl/MavenSimpleConfiguration.java
> +++ b/impl/maven-cli/src/main/java/org/apache/maven/cling/logging/impl/MavenSimpleConfiguration.java
> @@ -31,20 +31,13 @@ org.openrewrite.java.migrate.lang.SwitchCaseAssigningToSwitchExpression
>  public class MavenSimpleConfiguration extends BaseSlf4jConfiguration {
>      @Override
>      public void setRootLoggerLevel(Level level) {
> -        String value;
> -        switch (level) {
> -            case DEBUG:
> -                value = "debug";
> -                break;
> +        String value = switch (level) {
> +            case DEBUG: yield "debug";
>  
> -            case INFO:
> -                value = "info";
> -                break;
> +            case INFO: yield "info";
>  
> -            default:
> -                value = "error";
> -                break;
> -        }
> +            default: yield "error";
> +        };
>          System.setProperty("org.slf4j.simpleLogger.defaultLogLevel", value);
>      }
> ```
> 
> Does that align with what you were seeing there?

I've ran it on apache repos and got a few repos responding, but by far `apache/fineract` had the most changes (~34)
There were more initially (48) but we're seeing less of them as the recipe is now stricter to be safe, for example:
```java
    EnumOptionData optionData = new EnumOptionData(AccountType.INVALID.getValue().longValue(), AccountType.INVALID.getCode(), "Invalid");
        switch (type) {
            case INVALID:
                optionData = new EnumOptionData(AccountType.INVALID.getValue().longValue(), AccountType.INVALID.getCode(), "Invalid");
...
```
doesn't get migrated anymore because the `optionData` initializer code could have a side effect and we don't want to remove it, even though in many circumstances (like this one) it would be fine.

src/main/java/org/openrewrite/java/migrate/lang/SwitchUtils.java
```diff
@@ -0,0 +1,69 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.java.migrate.lang;
+
+import org.openrewrite.java.tree.*;
+
+import java.util.*;
+
```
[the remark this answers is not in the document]
pdelagrave: @timtebeek The default IDEA formatter and the `OpenRewriteRecipeBestPractices` recipe both unfold the wildcard yet we end up with this situation.

A unified consistent standard formatting configuration applicable both in the CI and in the IDE would be really useful.

PR #801, opened 2025-07-25
Convert switch expressions to arrows where possible

### LINKED ISSUE #799: Convert switch expressions with colon to arrows
- Following on from https://github.com/openrewrite/rewrite-migrate-java/pull/795

```java
class A {
    void doFormat(String str) {
        String formatted = switch (str) {
            case "foo": yield "Foo";
            case "bar": yield "Bar";
            case null, default: yield "unknown";
        };
    }
}
```

```java
class A {
    void doFormat(String str) {
        String formatted = switch (str) {
            case "foo" -> "Foo";
            case "bar" -> "Bar";
            case null, default -> "Other";
        };
    }
}
```

Did not want to make this part of other recipes: it can stand alone for smaller complexity and isolated value.

## Any additional context
Ideally we start this off minimally: only where all cases directly contain `yield`. That way we have something quickly and can chain in this recipe visitor as a `doAfterVisit` for `SwitchCaseAssignmentsToSwitchExpression`.

timtebeek:
- Fixes #799


src/main/resources/META-INF/rewrite/examples.yml
```diff
@@ -6314,6 +6314,62 @@ examples:
...
+      class Test {
+          String doFormat(String str) {
+              return switch (str) {
+                  case "foo" -> "Foo";
+                  case "bar" -> "Bar";
+                  case null, default -> "Other";
+              };
+          }
+      }
+    language: java
```
timtebeek:
```suggestion
```

timtebeek:
With this we're now seeing the following on Apache Maven:

image


We can optionally later loosen the restrictions we've placed on the implementation here to for instance allow empty label fall through, or case blocks with more than one statement, but this minimal implementation can be merged already.

PR #810, opened 2025-08-07
Remove limitations of recipe for making safe migration

YarochkinMichael:
## What's changed?
Joda recipe can perform a safe migration. 

## What's your motivation?
To execute a fully automated migration

Remove of check for `unsafeVars`

@amishra-u @timtebeek 

## Any additional context
With this change, tests for Scanner are obsolete.
I am not even sure we need `ScanningRecipe` anymore

### Checklist


src/main/java/org/openrewrite/java/migrate/joda/JodaTimeVisitor.java
```diff
@@ -219,38 +277,50 @@ private J migrateMethodCall(MethodCall original, MethodCall updated) {
         }
         MethodTemplate template = AllTemplates.getTemplate(original);
         if (template == null) {
+            System.out.println("Joda usage is found but mapping is missing: " + original);
             return original; // unhandled case
         }
+        if (template.getTemplate().getCode().equals(JODA_MULTIPLE_MAPPING_POSSIBLE)) {
+            System.out.println(JODA_MULTIPLE_MAPPING_POSSIBLE + ": " + original);
```
greg-at-moderne, line 282: I don't think we want to report anything via stdout. 

If anything, for such a cases we place some Marker and attach to the source LST element. I am not fully if that's necessary in this specific case, maybe just refrain from making any change? I am not sure.

YarochkinMichael: The idea is to report on missing mapping. It was extremely useful for me during development. And can be used for reporting later.

timtebeek:
As discussed, we'll continue development in https://github.com/openrewrite/rewrite-joda

Thanks again!

PR #456, opened 2024-04-16
feat: Missing "Add explicit Common Annotations dependencies" in "Migrate to Java 11"

ghusta:
Refs: #455

PR #816, opened 2025-08-12
Improve "Add explicit Common Annotations dependencies" in "Migrate to Java 11"

jevanlingen:
## What's changed?
The counteract recipe now works as intended. When migrating to Java 17, it will update any added or updated dependency on `jakarta.annotation:jakarta.annotation-api:1.3.5` to version 2.0.

## What's your motivation?
The original PR

- https://github.com/openrewrite/rewrite-migrate-java/pull/456

introduced the _AddCommonAnnotationsDependencies_ recipe. While useful, it made the migration chains harder to reason about. In rewrite-migrate-java, there are two major migration chains:
- Javax → Jakarta EE chain
- Java 7 → 8 → 11 → 17 → 21 → 25 chain

The _AddCommonAnnotationsDependencies_ recipe started to conceptually mingle the two chains. As migrating to Java 11 requires some dependency to support the javax namespace, it needs to be there. However, if you want to migrate to newer versions than Java 11, the Java chain never actually updated the Javax namespace to the Jakarta namespace (the counteract recipe was originally part of the EE chain only). 

In some cases, this could even lead to errors such as:

```
Execution failed for task ':compileJava'.  
> java.lang.NoSuchFieldError: Class com.sun.tools.javac.tree.JCTree$JCImport does not have member field 'com.sun.tools.javac.tree.JCTree qualid'
```
By adding the counteract recipe to the Java chain, we prevent these issues. It’s worth noting that the original counteract recipe wasn’t truly a “counter” recipe. It was more of a best-practice recipe; it's limited to the Jakarta EE chain and only worked for Spring Boot migrations.

### Checklist

### DIFF
--- a/src/main/resources/META-INF/rewrite/jakarta-ee-10.yml
+++ b/src/main/resources/META-INF/rewrite/jakarta-ee-10.yml
@@ -57,11 +57,14 @@ tags:
   - jakarta
 # NOTE: The spec versions in this section comes from https://github.com/jakartaee/jakartaee-api/blob/10.0.0/pom.xml
 recipeList:
-  - org.openrewrite.java.migrate.jakarta.UpdateJakartaAnnotations2
   - org.openrewrite.java.migrate.jakarta.UpdateJakartaXmlWsEE10
   - org.openrewrite.java.dependencies.UpgradeDependencyVersion:
       groupId: jakarta.activation
       artifactId: jakarta.activation-api
+      newVersion: 2.1.x
+  - org.openrewrite.java.dependencies.UpgradeDependencyVersion:
+      groupId: jakarta.annotation
+      artifactId: jakarta.annotation-api
       newVersion: 2.1.x
   - org.openrewrite.java.dependencies.UpgradeDependencyVersion:
       groupId: jakarta.authentication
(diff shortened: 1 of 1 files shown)


src/main/resources/META-INF/rewrite/jakarta-ee-10.yml
```diff
@@ -394,12 +394,6 @@ recipeList:
       groupId: jakarta.annotation
       artifactId: jakarta.annotation-api
       newVersion: 2.1.x
-  - org.openrewrite.java.ChangeType:
```
jevanlingen, line 397: ℹ️ The `UpdateJakartaAnnotations2` recipe should not contain these ChangeType changes, as the `JavaxAnnotationMigrationToJakartaAnnotation` recipe, located in jakarta-ee-9.xml, does already do these migrations:

https://github.com/openrewrite/rewrite-migrate-java/blob/main/src/main/resources/META-INF/rewrite/jakarta-ee-9.yml#L94-L129

Because jakarta-ee-10 calls jakarta-ee-9, the namespace fix will already be applied.

---

The reason these ChangeType recipes were added here, was because there was a bug in the `JavaxAnnotationMigrationToJakartaAnnotation` recipe, what let to an unchanged _javax_ namespace. That bug has been [fixed](https://github.com/openrewrite/rewrite-migrate-java/commit/aae42da80d49cdac948c58505925c0d6987d9cb0), so we don't need there recipe here anymore.

build.gradle.kts
```diff
@@ -77,6 +77,7 @@ dependencies {
     testRuntimeOnly("javax.mail:mail:1.4.7")
     testRuntimeOnly("javax.mail:javax.mail-api:1.6.2")
     testRuntimeOnly("javax.ws.rs:javax.ws.rs-api:2.1.1")
+    testRuntimeOnly("javax.annotation:javax.annotation-api:1.3.2")
```
Laurens-W: Could this go through type table support?

jevanlingen: Sure: https://github.com/openrewrite/rewrite-migrate-java/pull/816/commits/c70fb4cd1dd0509e8828808782dbc617f44594e1.

Laurens-W (approved):

timtebeek (approved):

timtebeek:
Noting it here in case it pops up: there's small chance some of these recipes might have been customized, and those customizations would then refer to a recipe that no longer exists; that's benign when running a recipe, but could fail validation in tests. The solution then is to either update or remove those references in customized recipes.

PR #822, opened 2025-08-13
Improve Java 17 and Java 21 switch expression tests

jevanlingen:
## What's changed?
Nothing. Only the tests are restructured between Java 17 and Java 21.

## What's your motivation?
The recipes target constructs related to the enhanced switch feature introduced in newer Java versions. While the recipes never produce uncompilable code, they can transform code into forms that are only valid starting from Java 21 (for example, a multi-label case that includes null and default).

To ensure tests accurately reflect the language features available in each version, I either split up the test file or used the `version` function to inform the reader this setup is only possible with Java 21.

### Checklist

### DIFF
--- a/src/test/java/org/openrewrite/java/migrate/lang/EnhancedSwitchCaseAssignmentsToSwitchExpressionTest.java
+++ b/src/test/java/org/openrewrite/java/migrate/lang/EnhancedSwitchCaseAssignmentsToSwitchExpressionTest.java
@@ -1,0 +1,330 @@
+/*
+ * Copyright 2025 the original author or authors.
+ * <p>
+ * Licensed under the Moderne Source Available License (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://docs.moderne.io/licensing/moderne-source-available-license
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.java.migrate.lang;
+
+import org.junit.jupiter.api.Test;
+import org.openrewrite.test.RecipeSpec;
+import org.openrewrite.test.RewriteTest;
+
+import static org.openrewrite.java.Assertions.java;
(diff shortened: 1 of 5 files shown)


src/test/java/org/openrewrite/java/migrate/lang/EnhancedSwitchCaseAssignmentsToSwitchExpressionTest.java
```diff
@@ -0,0 +1,330 @@
...
+          //language=java
+          java(
+            """
+              class B {
+                  void doFormat(String str) {
+                      String formatted = "initialValue";
+                      switch (str) {
+                          case "foo" -> formatted = "Foo";
+                          case "bar" -> formatted = "Bar";
+                          case null, default -> formatted = "Other";
```
Jenson3210, line 313: this is not allowed in 17 is it? 

[JDK 17 adds the (preview) ability to include the previously illegal case null clause in a switch.](https://blogs.oracle.com/javamagazine/post/java-switch-expression-arrow-case-null)

src/test/java/org/openrewrite/java/migrate/lang/SwitchExpressionYieldToArrowTest.java
```diff
@@ -197,4 +197,37 @@ void doFormat(TrafficLight light) {
...
+                      return formatted;
+                  }
+              }
+              """,
+              """
+              class Test {
+                  String format(String str) {
+                      String formatted = switch (str) {
+                          case "foo", "bar" -> "FooBar";
+                          case null, default -> "unknown";
```
Jenson3210, line 223: Same question

src/test/java/org/openrewrite/java/migrate/lang/EnhancedSwitchCaseAssignmentsToSwitchExpressionTest.java
jevanlingen: It's allowed in preview mode, which we do not support. But it does not matter really, the recipes works on LST elements. So the migration does "always" work, but the test were just reflecting code which we naturally never would encounter. Thus, it's better to have proper tests, which was why I created this PR in the first place.

src/test/java/org/openrewrite/java/migrate/lang/SwitchExpressionYieldToArrowTest.java
jevanlingen: idem

PR #823, opened 2025-08-13
Prevent switch expression migrations that produce uncompilable code on Java 17

jevanlingen:
## What's changed?
Switches are not changed to switch expressions when it leads to uncompilable code. 

## What's your motivation?
When running the recipes on a Java 17 codebase, certain transformations could result in uncompilable code.
For example, applying the SwitchCaseAssignmentsToSwitchExpression recipe to:

```java
String formatted = "initialValue";
switch (str) {
    case "foo": formatted = "Foo"; break;
    case "ignored", default: formatted = "unknown";
}
```

would produce:

```java
String formatted = switch (str) {
    case "foo" -> "Foo";
    case "ignored", default -> "unknown";
};
```

While the original statement compiles under Java 17, the transformed switch expression triggers the following error:

```
patterns in switch statements are a preview feature and are disabled by default.
```

This happens because combining a constant label (e.g., "ignored") with `default` in a single case label is only permitted with the pattern matching for switch preview feature in Java 17, which was finalized in Java 21.

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/lang/SwitchCaseAssignmentsToSwitchExpression.java
+++ b/src/main/java/org/openrewrite/java/migrate/lang/SwitchCaseAssignmentsToSwitchExpression.java
@@ -63,8 +63,8 @@ public class SwitchCaseAssignmentsToSwitchExpression extends Recipe {
 
             @Override
             public J.CompilationUnit visitCompilationUnit(J.CompilationUnit cu, ExecutionContext ctx) {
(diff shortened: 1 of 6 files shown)


timtebeek:
Thanks for pointing out these issues with the Java 17 migration; from a maintenance point of view I think I'd prefer to just move these recipes over to the Java 21 migration, also to provide incentive for folks to upgrade. Having these conditionals, and spread across files isn't ideal when we need to maintain these longer term, especially as switches will get more options like primitives in the near future.

jevanlingen:
Seems like a valid point! I'll close this PR and move everything to Java 21.

PR #835, opened 2025-08-25
Adding `jakarta.servlet:jakarta.servlet-api` explicitly when the `javax.servlet:javax.servlet-api` it is trying to migrate from was only available transitively.

steve-aom-elliott:
## What's your motivation?
In a similar vein to the changes [here](https://github.com/openrewrite/rewrite-migrate-java/commit/e15dadc0c726287e63e83352b981dbaea0374b82), when `javax.servlet:javax.servlet-api` was only available transitively, and the Jakarta EE 9 migration tried to change the dependency to `jakarta.servlet:jakarta.servlet-api`, it would end up missing it, as `ChangeDependency` only changes direct dependencies.

### Checklist

### DIFF
--- a/src/test/java/org/openrewrite/java/migrate/jakarta/JavaxServletToJakartaServletTest.java
+++ b/src/test/java/org/openrewrite/java/migrate/jakarta/JavaxServletToJakartaServletTest.java
@@ -42,9 +42,7 @@ class JavaxServletToJakartaServletTest implements RewriteTest {
             //language=xml
             pomXml(
               """
-                <?xml version="1.0" encoding="UTF-8"?>
-                <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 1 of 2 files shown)


src/test/java/org/openrewrite/java/migrate/jakarta/JavaxServletToJakartaServletTest.java
```diff
@@ -0,0 +1,212 @@
...
+    @Test
+    void upgradesJakartaServletApiDependencyIfAlreadyExistingAtLowerVersion() {
+        rewriteRun(
+          mavenProject(
+            "Sample",
+            //language=xml
+            pomXml(
+              """
+                <?xml version="1.0" encoding="UTF-8"?>
+                <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
```
timtebeek: Can we strip these dtds and xml entries? They're not needed and hurt readability.

steve-aom-elliott: Will do. One moment.

timtebeek (approved):
Thanks for the fix and quick rework!

PR #838, opened 2025-08-27
AddJaxbRuntime should use `overrideManagedDependency=true`

timtebeek:
- Ahead of https://github.com/openrewrite/rewrite/pull/5880

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/javax/AddJaxbRuntime.java
+++ b/src/main/java/org/openrewrite/java/migrate/javax/AddJaxbRuntime.java
@@ -218,13 +218,9 @@ public class AddJaxbRuntime extends ScanningRecipe<AtomicBoolean> {
         d = (Xml.Document) new org.openrewrite.maven.UpgradeDependencyVersion(sunJaxbRuntimeGroup, sunJaxbRuntimeArtifact, "2.3.x", null, null, null)
                 .getVisitor()
                 .visitNonNull(d, ctx);
-        d = (Xml.Document) new org.openrewrite.maven.ChangeDependencyGroupIdAndArtifactId(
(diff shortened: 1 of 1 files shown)


src/main/java/org/openrewrite/java/migrate/javax/AddJaxbRuntime.java
```diff
@@ -218,13 +218,9 @@ private Xml.Document jaxbDependencySwap(ExecutionContext ctx, Xml.Document d, St
...
-        d = (Xml.Document) new org.openrewrite.maven.ChangeDependencyGroupIdAndArtifactId(
+        return (Xml.Document) new org.openrewrite.maven.ChangeDependencyGroupIdAndArtifactId(
                 glassfishJaxbRuntimeGroup, glassfishJaxbRuntimeArtifact,
-                sunJaxbRuntimeGroup, sunJaxbRuntimeArtifact, "2.3.x", null
-        ).getVisitor().visitNonNull(d, ctx);
-        return (Xml.Document) new org.openrewrite.maven.ChangeManagedDependencyGroupIdAndArtifactId(
-                glassfishJaxbRuntimeGroup, glassfishJaxbRuntimeArtifact,
-                sunJaxbRuntimeGroup, sunJaxbRuntimeArtifact, "2.3.x"
-        ).getVisitor().visitNonNull(d, ctx);
+                sunJaxbRuntimeGroup, sunJaxbRuntimeArtifact, "2.3.x", null, false, true)
```
timtebeek, line 223: Since the version was already set there should be no difference from the change in
- https://github.com/openrewrite/rewrite/pull/5880

These explicit options match the implicit options, which meant we already no longer had to call `ChangeManagedDependencyGroupIdAndArtifactId` explicitly, as that's handled in `ChangeDependencyGroupIdAndArtifactId` for some time now.

Issue #839, opened 2025-09-02
Java 25 migration recipes


timtebeek:
## What problem are you trying to solve?
Upgrade to Java 25, picking up features since Java 21. Each of these can be picked up as a separate sub issue.
https://jdk.java.net/25/release-notes

- [ ] **New Reader.of(CharSequence) Method** ([JDK-8341566](https://bugs.openjdk.org/browse/JDK-8341566))
The static factory method java.io.Reader.of(CharSequence) has been added to get a Reader that reads characters from a CharSequence, String and StringBuilder are examples of a CharSequence. The returned Reader is more efficient than using a java.io.StringReader in some cases, as the latter requires conversion to String and synchronization.

- [ ] **New Method Process.waitFor(Duration)** ([JDK-8336479](https://bugs.openjdk.org/browse/JDK-8336479))
A new, overloaded method, java.lang.Process#waitFor(Duration), has been added. The existing waitFor() method with timeout needs both a primitive time out value and its unit. The new method uses java.time.Duration so that the user will not be confused with the unit.

- [ ] **java.util.zip.ZipError Is Deprecated for Removal** ([JDK-8336843](https://bugs.openjdk.org/browse/JDK-8336843))
java.util.zip.ZipError has been deprecated for removal. This error class has been obsolete for many releases and has not been used by the reference implementation since Java 8.
Any existing code that used to throw `java.util.zip.ZipError` should consider throwing `java.util.zip.ZipException` instead. Code needing to catch this error when running on releases prior to Java 9 may be updated to catch the parent `java.lang.InternalError` class instead.

PR #840, opened 2025-09-08
Java25  Migrate to Process::waitFor(Duration)

JohannisK:
## What's changed?
`Process::waitFor(Long, TimeUnit)` can be migrated to `Process::waitFor(Durarion)` in java 25.

## What's your motivation?
As requested in 
- https://github.com/openrewrite/rewrite-migrate-java/issues/839

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/lang/MigrateProcessWaitForDuration.java
+++ b/src/main/java/org/openrewrite/java/migrate/lang/MigrateProcessWaitForDuration.java
@@ -16,18 +16,13 @@
 package org.openrewrite.java.migrate.lang;
 
 import org.openrewrite.ExecutionContext;
-import org.openrewrite.Preconditions;
 import org.openrewrite.Recipe;
 import org.openrewrite.TreeVisitor;
 import org.openrewrite.java.JavaTemplate;
 import org.openrewrite.java.JavaVisitor;
 import org.openrewrite.java.MethodMatcher;
-import org.openrewrite.java.search.UsesJavaVersion;
 import org.openrewrite.java.tree.Expression;
 import org.openrewrite.java.tree.J;
-
-import java.util.Collections;
-import java.util.List;
 
 public class MigrateProcessWaitForDuration extends Recipe {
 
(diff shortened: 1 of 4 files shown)


src/main/java/org/openrewrite/java/migrate/lang/MigrateProcessWaitForDuration.java
```diff
@@ -0,0 +1,115 @@
...
+
+    @Override
+    public String getDescription() {
+        return "Use `Process#waitFor(Duration)` instead of `Process#waitFor(long, TimeUnit)` in Java 25 or higher.";
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return Preconditions.check(new UsesJavaVersion(25), new JavaVisitor<ExecutionContext>() {
+            private final MethodMatcher waitForMatcher = new MethodMatcher("java.lang.Process waitFor(long, java.util.concurrent.TimeUnit)");
```
Jenson3210: Can we pull this up to static field?

JohannisK: Yes, will do!

Jenson3210: Also, could you verify once if there is a test that uses Long iso long. I think this matcher should still work, but better safe than sorry 😄

src/main/java/org/openrewrite/java/migrate/lang/MigrateProcessWaitForDuration.java
```diff
@@ -0,0 +1,115 @@
...
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.java.migrate.lang;
+
+import org.openrewrite.ExecutionContext;
```
Jenson3210, line 19:
```suggestion
import org.jspecify.annotations.Nullable;
import org.openrewrite.ExecutionContext;
```

src/main/java/org/openrewrite/java/migrate/lang/MigrateProcessWaitForDuration.java
```diff
@@ -0,0 +1,115 @@
...
+
+                    maybeAddImport("java.time.Duration");
+                    maybeAddImport("java.time.temporal.ChronoUnit");
+                    maybeRemoveImport("java.util.concurrent.TimeUnit");
+                    maybeRemoveImport("java.util.concurrent.TimeUnit." + timeUnitName);
+                }
+                return mi;
+            }
+
+            private String getTimeUnitName(Expression timeUnitArg) {
```
Jenson3210:
```suggestion
            private @Nullable String getTimeUnitName(Expression timeUnitArg) {
```

src/main/java/org/openrewrite/java/migrate/lang/MigrateProcessWaitForDuration.java
```diff
@@ -0,0 +1,115 @@
...
+                }
+                if (timeUnitArg instanceof J.Identifier) {
+                    J.Identifier id = (J.Identifier) timeUnitArg;
+                    return id.getSimpleName();
+                }
+                return null;
+            }
+
+            private String getDurationMethod(String timeUnitName) {
+                switch (timeUnitName) {
```
Jenson3210:
```suggestion
            private @Nullable String getDurationMethod(@Nullable String timeUnitName) {
                if (timeUnitName == null) {
                    return null;
                }
                switch (timeUnitName) {
```

src/main/java/org/openrewrite/java/migrate/lang/MigrateProcessWaitForDuration.java
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    @Override
+    public String getDescription() {
+        return "Use `Process#waitFor(Duration)` instead of `Process#waitFor(long, TimeUnit)` in Java 25 or higher.";
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return Preconditions.check(new UsesJavaVersion(25), new JavaVisitor<ExecutionContext>() {
```
Jenson3210:
```suggestion
        return Preconditions.check(new UsesJavaVersion<>(25), new JavaVisitor<ExecutionContext>() {
```

src/main/resources/META-INF/rewrite/java-lang-apis.yml
```diff
@@ -29,6 +29,7 @@ recipeList:
   - org.openrewrite.java.migrate.lang.MigrateSecurityManagerMulticast
   - org.openrewrite.java.migrate.lang.MigrateClassLoaderDefineClass
   - org.openrewrite.java.migrate.lang.MigrateClassNewInstanceToGetDeclaredConstructorNewInstance
+  - org.openrewrite.java.migrate.lang.MigrateProcessWaitForDuration
```
Jenson3210: I think this recipe might have been originally written for the java 11 migration (as that's its only usage). not sure though...

JohannisK: I think that should be reflected in the name then. But I can change the location of this one ofcourse.

Jenson3210: it's also fine for me to keep this recipe in here and keep it as is. Just wanted to add a comment to inform

JohannisK: It's also added to the java 25 migration, so I removed it from the API's.

timtebeek (approved):

PR #844, opened 2025-09-09
Java 25:  Unused variable to underscore

JohannisK:
## What's changed?
Since Java 22 unused variables can be named "_"

## What's your motivation?
Using an "_" makes it explicit a variable is not used.


src/main/java/org/openrewrite/java/migrate/lang/ReplaceUnusedVariablesWithUnderscore.java
```diff
@@ -0,0 +1,124 @@
...
+    public String getDescription() {
+        return "Replace unused variable declarations with underscore (_) for Java 22+. " +
+                "This includes unused variables in enhanced for loops, catch blocks, " +
+                "and lambda parameters where the variable is never referenced.";
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return Preconditions.check(
+                new UsesJavaVersion<>(22),
```
JohannisK: If I saw correctly the feature is available since 22, but the first compatible parser will be for 25 (LST)
Is 22 correct here? Or should we start supporting this from Java25?

I think this change would be fine using the old parser, but I'm not sure what would happen if in a 22 codebase we execute this recipe, and then chain it with others?

Laurens-W: We should use 25, we don't support experimental features :)

JohannisK: As discussed we'll hook this into 25+

timtebeek:
Results look good at scale; turns out there's a ton of ignored exceptions that would be converted.

PR #845, opened 2025-09-10
Java 25: Migrate new Stringreader(..) to Reader.of(CharSequence)

JohannisK:
## What's changed?
Java 25: Migrate new Stringreader(..) to Reader.of(CharSequence)

## What's your motivation?
As requested in 
- https://github.com/openrewrite/rewrite-migrate-java/issues/839

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/util/MigrateStringReaderToReaderOf.java
+++ b/src/main/java/org/openrewrite/java/migrate/util/MigrateStringReaderToReaderOf.java
@@ -1,0 +1,198 @@
+/*
+ * Copyright 2025 the original author or authors.
+ * <p>
+ * Licensed under the Moderne Source Available License (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://docs.moderne.io/licensing/moderne-source-available-license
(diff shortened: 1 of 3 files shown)


src/main/java/org/openrewrite/java/migrate/util/MigrateStringReaderToReaderOf.java
```diff
@@ -0,0 +1,130 @@
...
+    @Override
+    public String getDisplayName() {
+        return "Use `Reader.of(CharSequence)` for non-synchronized readers";
+    }
+
+    @Override
+    public String getDescription() {
+        return "Migrate `new StringReader(String)` to `Reader.of(CharSequence)` in Java 25+. " +
+                "This only applies when assigning to `Reader` variables or returning from methods that return `Reader`. " +
+                "The new method creates non-synchronized readers which are more efficient when thread-safety is not required.";
```
JohannisK, line 50: This might be a reason to not include this recipe by default?

JohannisK: The usecases of StringReader in a mutlithreaded way few if even existing. For almost all cases the lack of synchronization is a performance gain.

PR #848, opened 2025-09-13
Add recipe to migrate `System.out.print/println` to `IO.print/println`

### LINKED ISSUE #846: Replace System.out.print(ln) with IO.print(ln)
[JEP 512](https://openjdk.org/jeps/512), [JDK-8305457](https://bugs.openjdk.org/browse/JDK-8305457) added a new `IO` class to JDK 25 that is automatically imported (as it is part of `java.lang`). It offers convenient methods to read and write from standard input and output. While aimed at beginners, it is a concise way to print to the console, and will feel natural to anyone who started their Java journey at version 25.

## What precondition(s) should be checked before applying this recipe?
JDK version >= 25

```java
class A {
    static void main() {
        System.out.print("Please enter your name: ");
        String name = IO.readln();
        System.out.println("Hello " + name);
    }
}
```

```java
class A {
    static void main() {
        IO.print("Please enter your name: ");
        String name = IO.readln();
        IO.println("Hello " + name);
    }
}
```

## Any additional context
Note that `IO` does not offer a way to print formatted strings, so `System.out.printf` will remain unchanged.

e5LA:
## What's changed?
Added  recipe to migrate `System.out.print()` and `System.out.println()` to `IO.print()` and `IO.println()`

## What's your motivation?
Closes #846.

### Checklist


e5LA:
While preparing the PR, I realized we might also want to handle `import static java.lang.System.out;` and `out.println(...)`. 
I’m not sure how common this pattern is - what do you think? 

I can look into it and we could add it in a follow-up PR

timtebeek:
Great to see thanks! I think we can support that without extra effort if we switch to using MethodMatcher :)

e5LA:
Good point, thanks @timtebeek , I'll update the code.

e5LA:
Thanks for the improvements @timtebeek ! Always great to see yours solutions.
Glad it made it into the Java 25 migration!.

PR #850, opened 2025-09-15
Update type tables via `./gradlew createTypeTable`

Jenson3210:
This PR updates the type tables using the `gradlew createTypeTable` task.
This PR was created using an automated script.


timtebeek:
Great to see, thanks! The `elementAnnotations` column now also shows the inline me annotations we need to generate recipes across different versions of Guava: neat!

PR #855, opened 2025-09-23
Fixed the whiteline removal for the schemaLocations when already properly migrated

Jenson3210:
Added some preconditional approach to the oldValue using regexes. 

I am not sure though if this change is worth the "win" of not seeing the change happening. 

Before this PR, 
```xml
<?xml version="1.0" encoding="UTF-8"?>
<web-app xmlns="https://jakarta.ee/xml/ns/jakartaee"
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation="https://jakarta.ee/xml/ns/jakartaee
          https://jakarta.ee/xml/ns/jakartaee/web-app_6_0.xsd"
         version="6.0">
    <display-name>Unit testing</display-name>
</web-fragment>
```

Would be migrated to 

```xml
<?xml version="1.0" encoding="UTF-8"?>
<web-app xmlns="https://jakarta.ee/xml/ns/jakartaee"
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation="https://jakarta.ee/xml/ns/jakartaee https://jakarta.ee/xml/ns/jakartaee/web-app_5_0.xsd"
         version="5.0">
    <display-name>Unit testing</display-name>
</web-fragment>
```

to than later be migrated back to version 6. (notice the newline diff resulting in the end causing unneeded reviewable change)

### DIFF
--- a/src/main/resources/META-INF/rewrite/jakarta-faces-4.yml
+++ b/src/main/resources/META-INF/rewrite/jakarta-faces-4.yml
@@ -199,7 +199,7 @@ recipeList:
   - org.openrewrite.xml.ChangeTagAttribute:
       attributeName: xsi:schemaLocation
       elementName: web-app
-      oldValue: (?<prefix>[\S\s]*https://jakarta\.ee/xml/ns/jakartaee/web-app_)5_\d+(?<suffix>\.xsd[\S\s]*)
+      oldValue: (?s)(?<prefix>.*https://jakarta\.ee/xml/ns/jakartaee/web-app_)5_\d+(?<suffix>\.xsd.*)
       newValue: ${prefix}6_0${suffix}
       regex: true
 ---


src/main/resources/META-INF/rewrite/jakarta-faces-4.yml
```diff
@@ -193,11 +193,15 @@ recipeList:
...
       attributeName: version
       elementName: web-app
+      oldValue: ^[12345]\.\d+$
       newValue: 6.0
+      regex: true
   - org.openrewrite.xml.ChangeTagAttribute:
       attributeName: xsi:schemaLocation
       elementName: web-app
-      newValue: https://jakarta.ee/xml/ns/jakartaee https://jakarta.ee/xml/ns/jakartaee/web-app_6_0.xsd
+      oldValue: (?<prefix>https://jakarta\.ee/xml/ns/jakartaee\s+https://jakarta\.ee/xml/ns/jakartaee/web-app_)5_\d+(?<extension>\.xsd)
```
timtebeek: This is where I wonder if for clarity (and less chance of mismatches) we can use a simpler pattern of replacements.
```suggestion
      oldValue: (?<prefix>.*https://jakarta\.ee/xml/ns/jakartaee/web-app_)5_\d+(?<extension>\.xsd)
```

src/main/resources/META-INF/rewrite/jakarta-faces-4.yml
```diff
@@ -193,11 +193,15 @@ recipeList:
...
       attributeName: version
       elementName: web-app
+      oldValue: ^[12345]\.\d+$
       newValue: 6.0
+      regex: true
   - org.openrewrite.xml.ChangeTagAttribute:
       attributeName: xsi:schemaLocation
       elementName: web-app
-      newValue: https://jakarta.ee/xml/ns/jakartaee https://jakarta.ee/xml/ns/jakartaee/web-app_6_0.xsd
+      oldValue: (?<prefix>[\S\s]*https://jakarta\.ee/xml/ns/jakartaee/web-app_)5_\d+(?<suffix>\.xsd[\S\s]*)
```
timtebeek: A character group of `[\S\s]` would just mean anything right? Then `.*` feels more intuitive.
```
\s	A whitespace character: [ \t\n\x0B\f\r]
\S	A non-whitespace character: [^\s]
```

Jenson3210: `.*` does not match on `\n` where `\s\S` will match on these newline variations also. I did not get it to work with `.*` and the added unit test/newlines.

timtebeek: Ah guess that would require a multiline flag to be added; in that case we can go with this option, or prefix the regex with `(?m)` to make it explicit that we tolerate multilines as well

Jenson3210: the multiline, I indeed also tried. 

That would work if we do not expect the entire string to match. That matches the line (so as of the newline) that matches the string, not the entire value. 
I did find that the dotall flag does match the entire surrounding values. 
Screenshot 2025-09-25 at 01 32 52
Screenshot 2025-09-25 at 01 32 46

timtebeek:
Any reason you'd held off on a merge here @Jenson3210 ?

PR #856, opened 2025-09-24
Use type tables for tests

knutwannheden:
Now using type tables.


build.gradle.kts
```diff
@@ -15,6 +15,7 @@ recipeDependencies {
     parserClasspath("javax.annotation:javax.annotation-api:1.3.2")
     parserClasspath("com.google.guava:guava:33.5.0-jre")
     parserClasspath("com.google.errorprone:error_prone_core:2.+")
+    testParserClasspath("jakarta.servlet:jakarta.servlet-api:5.0.+")
```
timtebeek: I don't see any jars deleted just yet; should we?
https://github.com/openrewrite/rewrite-migrate-java/tree/main/src/test/resources/META-INF/rewrite/classpath

knutwannheden: Good catch! I will delete that one.

PR #863, opened 2025-10-04
Fix UseVarForGenericMethodInvocations for nested generics with diamond operators

### LINKED ISSUE #862: UseVarForGenericMethodInvocations
I am using

- Maven plugin v6.18.0
- rewrite-migrate-java 3.18.0

I am using the Maven plugin, and my project is a mutli module project.
```xml
<project>
  <build>
    <plugins>
      <plugin>
        <version>6.18.0</version>
        <configuration>
          <exportDatatables>true</exportDatatables>
          <activeRecipes>
            <recipe>org.openrewrite.java.migrate.lang.var.UseVarForGenericMethodInvocations</recipe>
          </activeRecipes>
        </configuration>
        <dependencies>
          <dependency>
            <version>3.18.0</version>
          </dependency>
        </dependencies>
      </plugin>
    </plugins>
  </build>
</project>
```

```java
class A {
    private void dostuff() {
        List<String> strs = Collections.synchronizedList(new ArrayList<>());
        strs.forEach(this::soString);
    }

    public void soString(String s) {
        System.out.println(s);
    }
}
```

```java
class A {
    private void dostuff() {
        var strs = Collections.synchronizedList(new ArrayList<String>());
        strs.forEach(this::soString);
    }

    public void soString(String s) {
        System.out.println(s);
    }
}
```

## What did you see instead?

```java
class A {
    private void dostuff() {
        var strs = Collections.synchronizedList(new ArrayList<>());
        strs.forEach(this::soString);
    }

    public void soString(String s) {
        System.out.println(s);
    }
}
```

Compiler error:  incompatible types: invalid method reference
[ERROR]     incompatible types: java.lang.Object cannot be converted to java.lang.String

Sure, I'll try and create a unit test for this as a starter and see from there on

timtebeek:
## Summary

Fixes #862 by making nested generic type parameters explicit when converting to `var`.

- Adds test cases replicating the issue with `Collections.synchronizedList(new ArrayList<>())`
- Recipe now replaces diamond operators in nested constructor calls with explicit type parameters
- Prevents compilation errors when using method references with nested generics

## Changes

- Modified `UseVarForGenericMethodInvocations` to detect and fix diamond operators in nested `new` expressions
- Added `makeNestedGenericsExplicit()` method to copy type parameters from variable declaration to nested constructors
- Added test cases: `nestedGenericsSimple()` and `nestedGenericsWithMethodReference()`

## Test plan

- [x] Added unit tests that replicate the issue
- [x] All existing tests pass
- [x] New tests verify the fix works correctly

### DIFF
--- a/src/test/java/org/openrewrite/java/migrate/lang/var/UseVarForGenericMethodInvocationsTest.java
+++ b/src/test/java/org/openrewrite/java/migrate/lang/var/UseVarForGenericMethodInvocationsTest.java
@@ -259,5 +259,85 @@ class UseVarForGenericMethodInvocationsTest implements RewriteTest {
               )
             );
         }
+
+        @Test
+        void nestedGenericsSimple() {
+            //language=java
+            rewriteRun(
+              version(
+                java("""
+                  package com.example.app;
+
+                  import java.util.ArrayList;
+                  import java.util.Collections;
+                  import java.util.List;
+
+                  class A {
+                    private void dostuff() {
+                        List<String> strs = Collections.synchronizedList(new ArrayList<>());
+                    }
+                  }
+                  """, """
+                  package com.example.app;
+
+                  import java.util.ArrayList;
+                  import java.util.Collections;
+                  import java.util.List;
+
+                  class A {
+                    private void dostuff() {
+                        var strs = Collections.synchronizedList(new ArrayList<String>());
+                    }
+                  }
+                  """),
+                10
+              )
+            );
+        }
+
+        @Test
+        void nestedGenericsWithMethodReference() {
+            //language=java
+            rewriteRun(
+              version(
+                java("""
+                  package com.example.app;
+
+                  import java.util.ArrayList;
+                  import java.util.Collections;
+                  import java.util.List;
+
+                  class A {
+                    private void dostuff() {
+                        List<String> strs = Collections.synchronizedList(new ArrayList<>());
+                        strs.forEach(this::soString);
+                    }
+
+                    private void soString(String s) {
+                        System.out.println(s);
+                    }
+                  }
+                  """, """
+                  package com.example.app;
+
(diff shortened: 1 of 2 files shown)


src/test/java/org/openrewrite/java/migrate/lang/var/UseVarForGenericMethodInvocationsTest.java
```diff
@@ -259,5 +260,40 @@ void m() {
...
+              version(
+                java(
+                  """
+                    import java.util.ArrayList;
+                    import java.util.Collections;
+                    import java.util.List;
+
+                    class A {
+                      void dostuff() {
+                          List<String> strs = Collections.synchronizedList(new ArrayList<>());
```
timtebeek, line 278: Are there additional cases we might need to cover here you think?

MBoegers: I could think of factory methods that might need special handling, like `List<Inetger> ints = List.of()` --> `var ints = List.<Integer>of()` but if I recall correctly, we decided to not migrate static generic methods at all due to this.
**EDIT:** Recalled correctly; see recipe description.

MBoegers: I wonder why we haven't skipped `Collections.synchronizedList(..)` in the first place; it's a generic static method. Would you mind verifying if you find any loophole there?

MBoegers: all good here, `Collections.synchronizedList(..)` is migrated because it has parameters, `List.of()` is not as it has no parameters.

Issue #897, opened 2025-10-24
Sets.filter is not migrated, which creates a conflict with the PreferJavaUtilPredicate recipe


lredor:
- This issue is similar to the issue #895. The `Predicate` class from Guava is converted in `java.util.Predicate` even when it is used in a `Sets.filter` method that is **not** converted. As a result, the following error appears after migration: `The method filter(Set<E>, Predicate<? super E>) in the type Sets is not applicable for the arguments (Set<Object>, Predicate<Object>)`.

## What version of OpenRewrite are you using?

I am using
- Rewrite Maven Plugin v6.22.1
- Maven v3.9.5
- org.openrewrite.recipe:rewrite-migrate-java v3.20.0

## How are you running OpenRewrite?
I used the Maven command line to launch the recipe on the Sirius Desktop repository : 
```bash
mvn -U org.openrewrite.maven:rewrite-maven-plugin:run -Drewrite.recipeArtifactCoordinates=org.openrewrite.recipe:rewrite-migrate-java:RELEASE -Drewrite.activeRecipes=org.openrewrite.java.migrate.guava.NoGuava -Drewrite.exportDatatables=true -f packaging/org.eclipse.sirius.parent/pom.xml
```

## What is the smallest, simplest way to reproduce the problem?

```java
              import java.util.HashSet;
              import java.util.Objects;
              import java.util.Set;

              import com.google.common.base.Predicate;
              import com.google.common.collect.Sets;

              class Test {
                  public static Set<Object> test() {
                      Set<Object> set = new HashSet<>();
                      Predicate<Object> isNotNull = Objects::nonNull;
                      return Sets.filter(set, isNotNull);
                  }
              }
```

## What did you expect to see?

The method `Sets.filter` can be replaced by `Collection.stream().filter(Predicate<? super T>).collect(Collectors.toSet())`.

```java
              import java.util.HashSet;
              import java.util.Objects;
              import java.util.Set;
              import java.util.function.Predicate;
              import java.util.stream.Collectors;

              class Test {
                  public static Set<Object> test() {
                      Set<Object> set = new HashSet<>();
                      Predicate<Object> isNotNull = Objects::nonNull;
                      return set.stream().filter(isNotNull).collect(Collectors.toSet());
                  }
              }
```

## What did you see instead?
```java
              import java.util.HashSet;
              import java.util.Objects;
              import java.util.Set;
              import java.util.function.Predicate;

              import com.google.common.collect.Sets;

[...]

PR #898, opened 2025-10-24
Convert Guava `Sets.filter`

lredor:
## What's changed?
- Add conversion for `Sets.filter` (as described in #897)

First contributions, so I’m not sure I’ve mastered all aspects of OpenRewrite yet.

@timtebeek 

### Checklist

### DIFF
--- a/src/test/java/org/openrewrite/java/migrate/guava/NoGuavaSetsFilterTest.java
+++ b/src/test/java/org/openrewrite/java/migrate/guava/NoGuavaSetsFilterTest.java
@@ -16,6 +16,8 @@
 package org.openrewrite.java.migrate.guava;
 
 import org.junit.jupiter.api.Test;
+import org.junitpioneer.jupiter.ExpectedToFail;
+import org.junitpioneer.jupiter.Issue;
 import org.openrewrite.DocumentExample;
 import org.openrewrite.InMemoryExecutionContext;
 import org.openrewrite.java.JavaParser;
@@ -103,6 +105,8 @@ class NoGuavaSetsFilterTest implements RewriteTest {
         );
     }
 
+    @ExpectedToFail("Sets.filter has special handling for sorted sets which now gets lost")
+    @Issue("https://github.com/openrewrite/rewrite-migrate-java/pull/898#discussion_r2461706208")
     @Test
     void replaceSetsFilterIndirectlyUsingSortedSet() {
         //language=java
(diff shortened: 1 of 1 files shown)


src/main/java/org/openrewrite/java/migrate/guava/NoGuavaSetsFilter.java
```diff
@@ -0,0 +1,76 @@
...
+                new JavaIsoVisitor<ExecutionContext>() {
+                    @Override
+                    public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, ExecutionContext ctx) {
+                        if (SETS_FILTER.matches(method)) {
+                            maybeRemoveImport("com.google.common.base.Predicate");
+                            maybeRemoveImport("com.google.common.collect.Sets");
+                            maybeAddImport("java.util.function.Predicate");
+                            maybeAddImport("java.util.stream.Collectors");
+
+                            return JavaTemplate.builder("#{any(java.util.Collection)}.stream().filter(#{any(java.util.function.Predicate)}).collect(Collectors.toSet())")
```
timtebeek: There's a slight concern here where `Sets` has special handling for `SortedSet`, whereas the the replacement always uses `toSet`, which uses `HashSet`. Should we conditionally use `.toCollection(TreeSet::new)` ?

lredor: You're right. I added a case for that.

src/test/java/org/openrewrite/java/migrate/guava/NoGuavaSetsFilterTest.java
```diff
@@ -0,0 +1,143 @@
...
+              import java.util.Set;
+              import java.util.SortedSet;
+
+              import com.google.common.base.Predicate;
+              import com.google.common.collect.Sets;
+
+              class Test {
+                  public static Set<Object> test(SortedSet<Object> set, Predicate<Object> isNotNull) {
+                      Set<Object> indirectSet = set;
+                      return Sets.filter(indirectSet, isNotNull);
```
timtebeek, line 125: This case is tricky; we can't really deduce from the type system whether we'd end up in this conditional of special handling:
https://github.com/google/guava/blob/21cdcf6d9a49c1a498acb422471d8d71f9219865/guava/src/com/google/common/collect/Sets.java#L1190-L1194

timtebeek: I'd hope that to be rare enough to perhaps still merge the current work, but there is a risk here which could only be mitigated by data flow analysis to figure out if it might be a SortedSet going into this `filter(Set, Predicate)` method. 🤔

timtebeek: Either way thanks for the quick rework; would love to hear your thoughts on this edge case.

timtebeek (approved):
Thanks again! I think the risk is low here for folks relying on that `instanceof SortedSet` check, so we'll merge this as is with the test documented to fail.

Issue #899, opened 2025-10-24
Collections2.filter is not migrated, which creates a conflict with the PreferJavaUtilPredicate recipe


lredor:
- This issue is similar to the issue #897. The `Predicate` class from Guava is converted in `java.util.Predicate` even when it is used in a `Collections2.filter` method that is **not** converted. As a result, the following error appears after migration: `The method filter(Collection<E>, Predicate<? super E>) in the type Collections2 is not applicable for the arguments (Collection<Object>, Predicate<Object>)`.

## What version of OpenRewrite are you using?

I am using
- Rewrite Maven Plugin v6.22.1
- Maven v3.9.5
- org.openrewrite.recipe:rewrite-migrate-java v3.20.0

## How are you running OpenRewrite?
I used the Maven command line to launch the recipe on the Sirius Desktop repository : 
```bash
mvn -U org.openrewrite.maven:rewrite-maven-plugin:run -Drewrite.recipeArtifactCoordinates=org.openrewrite.recipe:rewrite-migrate-java:RELEASE -Drewrite.activeRecipes=org.openrewrite.java.migrate.guava.NoGuava -Drewrite.exportDatatables=true -f packaging/org.eclipse.sirius.parent/pom.xml
```

## What is the smallest, simplest way to reproduce the problem?

```java
              import java.util.ArrayList;
              import java.util.Collection;
              import java.util.Objects;

              import com.google.common.base.Predicate;
              import com.google.common.collect.Collections2;

              class Test {
                  public static Collection<Object> test() {
                      Collection<Object> collection = new ArrayList<>();
                      Predicate<Object> isNotNull = Objects::nonNull;
                      return Collections2.filter(collection, isNotNull);
                  }
              }
```

## What did you expect to see?

The method `Collections2.filter` can be replaced by `Collection.stream().filter(Predicate<? super T>).toList()`.

```java
              import java.util.ArrayList;
              import java.util.Collection;
              import java.util.Objects;
			  import java.util.function.Predicate;

              class Test {
                  public static Collection<Object> test() {
                      Collection<Object> collection = new ArrayList<>();
                      Predicate<Object> isNotNull = Objects::nonNull;
                      return collection.stream().filter(isNotNull).toList();
                  }
              }
```

## What did you see instead?
```java
              import java.util.ArrayList;
              import java.util.Collection;
              import java.util.Objects;

              import com.google.common.collect.Collections2;

              class Test {
                  public static Collection<Object> test() {
[...]

PR #900, opened 2025-10-24
Convert Guava `Collections2.filter`

lredor:
## What's changed?
- Add conversion for `Collections2.filter` (as described in #899)

First contributions, so I’m not sure I’ve mastered all aspects of OpenRewrite yet.

@timtebeek 

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/guava/NoGuavaCollections2Filter.java
+++ b/src/main/java/org/openrewrite/java/migrate/guava/NoGuavaCollections2Filter.java
@@ -50,7 +50,7 @@ public class NoGuavaCollections2Filter extends Recipe {
     @Override
     public TreeVisitor<?, ExecutionContext> getVisitor() {
         return Preconditions.check(
-                new UsesMethod<>("com.google.common.collect.Collections2 filter(java.util.Collection, com.google.common.base.Predicate)"),
+                new UsesMethod<>(COLLECTIONS2_FILTER),
                 new JavaIsoVisitor<ExecutionContext>() {
                     @Override
                     public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, ExecutionContext ctx) {
--- a/src/test/java/org/openrewrite/java/migrate/guava/NoGuavaCollections2FilterTest.java
+++ b/src/test/java/org/openrewrite/java/migrate/guava/NoGuavaCollections2FilterTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2024 the original author or authors.
+ * Copyright 2025 the original author or authors.
  * <p>
  * Licensed under the Moderne Source Available License (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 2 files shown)


src/main/java/org/openrewrite/java/migrate/guava/NoGuavaCollections2Filter.java
```diff
@@ -0,0 +1,74 @@
...
+                new UsesMethod<>(COLLECTIONS2_FILTER),
+                new JavaIsoVisitor<ExecutionContext>() {
+                    @Override
+                    public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, ExecutionContext ctx) {
+                        if (COLLECTIONS2_FILTER.matches(method)) {
+                            maybeRemoveImport("com.google.common.base.Predicate");
+                            maybeRemoveImport("com.google.common.collect.Collections2");
+                            maybeAddImport("java.util.function.Predicate");
+
+                            return JavaTemplate.builder("#{any(java.util.Collection)}.stream().filter(#{any(java.util.function.Predicate)}).toList()")
```
timtebeek, line 62: I'm not sure these are as interchangeable ; `Collections2.filter` returns a `com.google.common.collect.Collections2.FilteredCollection` which acts as a view on the underlying collection, whereas `.toList()` returns a fully new collection with very different performance and semantics when adding/removing entries. Any thoughts there?

timtebeek (commented):
Great to see! Some concern with the approach taken here, as we strive to [do no harm](https://docs.openrewrite.org/authoring-recipes/recipe-conventions-and-best-practices#do-no-harm). As much as it'll likely mostly be what folks want, we can't be certain they don't rely on the old window-like behavior when applying changes.

src/main/java/org/openrewrite/java/migrate/guava/NoGuavaCollections2Filter.java
lredor: You’re certainly right (about the performance), but the current Guava recipes cause compilation errors, as explained in the description of issue #899.
To respect the “do no harm” principle, one option might be to conditionally replace com.google.common.base.Predicate with java.util.function.Predicate.
However, this seems a bit more complex — probably too complex for me at this stage.

I probably thought a bit too quickly when creating my latest issues and PRs, aiming to eliminate all compilation errors and remove any remaining Guava dependencies.

timtebeek: Yes understandable that you're looking for either a more complete migration off of Guava (with no compilation issues), or a more cautious move off of Guava's Predicate onto Java's Predicate. We have something similar when choosing to move off of more specific Collections methods onto their interfaces, as seen in this recipe:
https://github.com/openrewrite/rewrite-static-analysis/blob/da756464470f9c5dbc056fa2e008123d73a50e05/src/main/java/org/openrewrite/staticanalysis/UseCollectionInterfaces.java#L277-L297

Perhaps good to also/instead add similar handling before we change Predicate: look how it's used, and only make changes when safe to do so.

timtebeek: Either way appreciate the help both in developing recipes, and thinking through the options. 🙏🏻

Issue #905, opened 2025-10-26
Functions.compose is not migrated, which creates a conflict with the PreferJavaUtilFunction recipe


lredor:
Functions.compose is not migrated, which creates a conflict with the PreferJavaUtilFunction recipe

- This issue is similar to the issue #897. The `Function` class from Guava is converted in `java.util.function.Function` even when it is used in a `Functions.compose` method that is **not** converted. As a result, the following error appears after migration: `The method compose(Function<B,C>, Function<A,? extends B>) in the type Functions is not applicable for the arguments (Function<String,Integer>, Function<Object,String>)`.

## What version of OpenRewrite are you using?

I am using
- Rewrite Maven Plugin v6.23.0
- Maven v3.9.5
- org.openrewrite.recipe:rewrite-migrate-java v3.21.0

## How are you running OpenRewrite?
I used the Maven command line to launch the recipe on the Sirius Desktop repository : 
```bash
mvn -U org.openrewrite.maven:rewrite-maven-plugin:6.23.0-SNAPSHOT:run -Drewrite.recipeArtifactCoordinates=org.openrewrite.recipe:rewrite-migrate-java:3.21.0-SNAPSHOT -Drewrite.activeRecipes=org.openrewrite.java.migrate.guava.NoGuava -Drewrite.exportDatatables=true -Dmaven.repo.local=./.repositoryMaven -f packaging/org.eclipse.sirius.parent/pom.xml
```

## What is the smallest, simplest way to reproduce the problem?

```java
import com.google.common.base.Function;
import com.google.common.base.Functions;

class Test {
    public static void test() {
        Function<Object, Integer> composed = Functions.compose(new Function<String, Integer>() {
            @Override
            public Integer apply(String input) {
                return input.length();
            }
        }, new Function<Object, String>() {
            @Override
            public String apply(Object input) {
                return input.toString();
            }
        });
    }
}
```

## What did you expect to see?

```java
import java.util.function.Function;

class Test {
    public static void testResult() {
        Function<Object, Integer> composed = new Function<String, Integer>() {
            @Override
            public Integer apply(String input) {
                return input.length();
            }
        }.compose(new Function<Object, String>() {
            @Override
            public String apply(Object input) {
                return input.toString();
            }
        });
    }
}
```

## What did you see instead?
```java
import com.google.common.base.Functions;

import java.util.function.Function;

class Test {
    public static void test() {
[...]

PR #906, opened 2025-10-26
Convert Guava `Functions.compose`

lredor:
## What's changed?
- Add conversion for `Functions.compose` (as described in #905)

Compared to recent changes, I think that this conversion is safe (respect of "do no harm").

@timtebeek 

### Checklist


timtebeek (approved):
Great to see, thanks!

PR #914, opened 2025-10-28
Change the Jetty dependency coordinates going from EE9 to EE10

ChenyuWang98:
### Checklist

### DIFF
--- a/src/main/resources/META-INF/rewrite/jakarta-ee-10.yml
+++ b/src/main/resources/META-INF/rewrite/jakarta-ee-10.yml
@@ -526,75 +526,60 @@ recipeList:
 ---
 type: specs.openrewrite.org/v1beta/recipe
 name: org.openrewrite.java.migrate.jakarta.JettyUpgradeEE10
-displayName: Update Jetty9 to Jetty12
-description: Update Jetty dependencies from version 9 to version 12.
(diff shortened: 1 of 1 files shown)


src/main/resources/META-INF/rewrite/jakarta-ee-10.yml
```diff
@@ -523,3 +524,81 @@ recipeList:
...
+description: Update Jetty dependencies from version 9 to version 12.
+recipeList:
+  - org.openrewrite.java.dependencies.ChangeDependency:
+      oldGroupId: org.eclipse.jetty.websocket
+      oldArtifactId: websocket-api
+      newArtifactId: jetty-websocket-jetty-api
+      newVersion: 12.0.x
+  - org.openrewrite.java.dependencies.ChangeDependency:
+      oldGroupId: org.eclipse.jetty.websocket
+      oldArtifactId: websocket-server
```
timtebeek: Here and below I think we should expect the `ee9` specific dependencies and change those to `ee10`.
```suggestion
      oldGroupId: org.eclipse.jetty.ee9.websocket
      oldArtifactId: jetty-ee9-websocket-jetty-server
```

ChenyuWang98:
I think there's no problem. My only wish is that Jetty be upgraded to EE10 concurrently with the Jakarta EE10 upgrade.

PR #921, opened 2025-10-28
Ignore `java.util.Iterator` in NoGuavaSetsNewHashSet recipe

lredor:
## What's changed?

Before this commit, `java.lang.Iterable` was ignored to avoid generating broken code.
This commit adds `java.util.Iterator` to the ignore list for the same reason.
A corresponding test has also been added.

### Checklist


lredor:
Thanks for the clarification 👍

PR #923, opened 2025-10-29
Do not migrate to Jakarta annotations as part of Java 17 upgrade

Laurens-W:
## What's changed?

Removed java-ee migration steps from java migration

## What's your motivation?

The removed step caused the jakarta namespace to be adopted when perhaps a framework used hadn't adopted that yet. These steps should be separate and composed together when appropriate

## Any additional context

The only time when java-ee and java migrations go hand in hand is when some modules were removed from the jdk with java 11

### Checklist

### DIFF
--- a/.sdkmanrc
+++ b/.sdkmanrc
@@ -1,3 +1,3 @@
 # Enable auto-env through the sdkman_auto_env config
 # Add key=value pairs of SDKs to use below
-java=21.0.7-tem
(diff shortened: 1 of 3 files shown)


Laurens-W:
I think we should also remove the changing of the scope to `provided` during the Java 11 migration, or at least see if we can guard that against causing a non compiling state

PR #924, opened 2025-10-29
Set build tools to use Java 8 when running Java 8 upgrade

MBoegers:
## What's changed?

added Java `org.openrewrite.java.migrate.UpgradeJavaVersion` to java updates predating java 17. 

## What's your motivation?

Enable iterative updates 7 -> 8 -> 11 -> 17

without you need to manual fix.

### Checklist


timtebeek (approved):
Minimized the changes with the following steps
- Removed a duplicate for Java 11
- Inlined the recipe for Java 8, to avoid confusion in search & docs
- Removed the recipes for 6 & 7, as they delegate to recipes that expect "1.6" and "1.7", which we need not support

PR #927, opened 2025-10-30
Always add explicit type to the `Predicate.isEqual` method

lredor:
## What's changed?
After analysis, I discovered 2 new cases detailed in the commit message.

> In some cases, the type can be inferred, but not always.
To follow the “Do no harm” principle, it’s safer to always "cast" the
`Predicate.isEqual` method explicitly.
Before this commit, certain cases resulted in compilation errors (see
the new test inlinedPredicatesEqualToWithStandardType, which reveals
this problem).

### Checklist


timtebeek:
I've briefly tried adding https://github.com/openrewrite/rewrite-static-analysis/blob/main/src/main/java/org/openrewrite/staticanalysis/UnnecessaryExplicitTypeArguments.java here, but that didn't help remove the types where they are not required. Still better to compile with excess types than to not compile at all, so let's merge this. Thanks again!

PR #929, opened 2025-10-31
`jaxws-maven-plugin` to 2.3.X

MBoegers:
## What's changed?
The WSDL generation maven plugin predating 2.3.0 is not Java 11 compatible and has breaking configuration changes.

## What's your motivation?
Smoth updates of SOAP Services

### Checklist

### DIFF
--- a/src/test/java/org/openrewrite/java/migrate/maven/MigrateJaxwsMavenPluginTest.java
+++ b/src/test/java/org/openrewrite/java/migrate/maven/MigrateJaxwsMavenPluginTest.java
@@ -20,7 +20,6 @@ import org.openrewrite.DocumentExample;
 import org.openrewrite.test.RecipeSpec;
 import org.openrewrite.test.RewriteTest;
 
-import static org.assertj.core.api.Assertions.assertThat;
 import static org.openrewrite.maven.Assertions.pomXml;
 
 class MigrateJaxwsMavenPluginTest implements RewriteTest {


src/main/resources/META-INF/rewrite/jaxb-plugins.yml
```diff
@@ -0,0 +1,36 @@
+#
+# Copyright 2024 the original author or authors.
```
timtebeek:
```suggestion
# Copyright 2025 the original author or authors.
```

src/main/resources/META-INF/rewrite/jaxb-plugins.yml
```diff
@@ -0,0 +1,36 @@
...
+displayName: Migrate JAXB-WS Plugin
+description: Upgrade the JAXB-WS Maven plugin to be compatible with Java 11.
+tags:
+  - java11
+recipeList:
+  - org.openrewrite.maven.ChangePluginGroupIdAndArtifactId:
+      oldGroupId: org.jvnet.jax-ws-commons
+      oldArtifactId: jaxws-maven-plugin
+      newGroupId: com.sun.xml.ws
+      newVersion: 2.3.5
```
timtebeek: Why 2.3.5 specifically and not say 2.3.7, or 2.3.+?
https://mvnrepository.com/artifact/com.sun.xml.ws/jaxws-maven-plugin

src/test/java/org/openrewrite/java/migrate/maven/MigrateJaxwsMavenPluginTest.java
```diff
@@ -0,0 +1,113 @@
...
+                          <plugin>
+                              <groupId>org.jvnet.jax-ws-commons</groupId>
+                              <artifactId>jaxws-maven-plugin</artifactId>
+                              <version>2.1</version>
+                              <executions>
+                                  <execution>
+                                      <id>wsimport-from-jdk</id>
+                                      <goals>
+                                          <goal>wsimport</goal>
+                                      </goals>
```
timtebeek, line 55: This seems to be the target desired state already; should we not test for whatever the old value was?

src/main/resources/META-INF/rewrite/jaxb-plugins.yml
```diff
@@ -0,0 +1,36 @@
...
+  - org.openrewrite.maven.ChangePluginExecutions:
+      groupId: org.jvnet.jax-ws-commons
+      artifactId: jaxws-maven-plugin
+      executions: "<id>wsimport-from-jdk</id><phase>generate-sources</phase><goals><goal>wsimport</goal></goals>"
+  - org.openrewrite.xml.ChangeTagName:
+      elementName: '/project/build/plugins/plugin/executions/execution/configuration/nocompile'
```
timtebeek: This will change `nocompile` in any plugin, but not when used in a profile for instance. Can we specifically match the `jaxws-maven-plugin` plugin by idea, but anywhere where it might be found (start with `//plugin[...]`) ?

src/main/resources/META-INF/rewrite/jaxb-plugins.yml
MBoegers: because 2.3.X makes it into the pom :D resuling in `<version>2.3.X</version>`

src/test/java/org/openrewrite/java/migrate/maven/MigrateJaxwsMavenPluginTest.java
MBoegers: Good catch, this is missing the explicit tying to `<phase>generate-sources</phase>`

src/main/resources/META-INF/rewrite/jaxb-plugins.yml
timtebeek: So then use the following? Or we can look to fix the upstream issue to resolve the versions, but I'd suggest for now:
```suggestion
      newVersion: 2.3.7
```

MBoegers: [ChangePluginGroupIdAndArtifactId](https://docs.openrewrite.org/recipes/maven/changeplugingroupidandartifactid#options) has a pattern as an example but uses it unresolved. so `2.3.+` make it into the pom.xml

timtebeek: Fixed in https://github.com/openrewrite/rewrite/commit/d330b7eac63d206d1ca2b5e0d643ad42e63df83a ; can we adopt `2.3.7` here as suggested above?

src/main/resources/META-INF/rewrite/jaxb-plugins.yml
MBoegers: I tried the wrong XPath syntax, used `//plugin[groupId/text() = '..']` instead, which didn't work as we don't implement the functions.

PR #936, opened 2025-11-12
Handle `JaxbAnnotationIntrospector` and `JacksonJsonProvider` in migration

CalvinZengyi:
## What's changed?

Added two more type changes dealing with JakartaEE 9 migration:
- `com.fasterxml.jackson.module.jaxb.JaxbAnnotationIntrospector` -> `com.fasterxml.jackson.module.jakarta.xmlbind.JakartaXmlBindAnnotationIntrospector`
- `com.fasterxml.jackson.jaxrs.json.JacksonJsonProvider` -> `com.fasterxml.jackson.jakarta.rs.json.JacksonJsonProvider`
## What's your motivation?

These two classes are impacted by Jakarta namespace migration ([reference from stackoverflow](https://stackoverflow.com/questions/77639313/cannot-use-jackson-jaxb-after-spring-upgrade))
I have been doing so in my local recipe, and I thought it would be helpful to upstream them to OpenRewrite.

Yes. I have had them in my local recipe as an enhancement on top of the JakartaEE migration recipe
### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/lang/ReplaceUnusedVariablesWithUnderscore.java
+++ b/src/main/java/org/openrewrite/java/migrate/lang/ReplaceUnusedVariablesWithUnderscore.java
@@ -51,7 +51,7 @@ public class ReplaceUnusedVariablesWithUnderscore extends Recipe {
 
     @Override
     public TreeVisitor<?, ExecutionContext> getVisitor() {
-        return Preconditions.check(new UsesJavaVersion<>(25), new JavaIsoVisitor<ExecutionContext>() {
+        return Preconditions.check(new UsesJavaVersion<>(22), new JavaIsoVisitor<ExecutionContext>() {
             @Override
             public J.ForEachLoop visitForEachLoop(J.ForEachLoop forLoop, ExecutionContext ctx) {
                 J.ForEachLoop l = super.visitForEachLoop(forLoop, ctx);
--- a/src/main/resources/META-INF/rewrite/examples.yml
+++ b/src/main/resources/META-INF/rewrite/examples.yml
@@ -3458,7 +3458,7 @@ examples:
 
       class A {
           public static Predicate<String> isHelloPredicate() {
-              return Predicate.isEqual("hello");
+              return Predicate.<String>isEqual("hello");
           }
       }
     language: java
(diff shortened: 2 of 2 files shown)


src/main/resources/META-INF/rewrite/jakarta-ee-9.yml
```diff
@@ -1055,6 +1055,12 @@ recipeList:
   - org.openrewrite.java.ChangeType:
       oldFullyQualifiedTypeName: com.fasterxml.jackson.module.jaxb.JaxbAnnotationModule
       newFullyQualifiedTypeName: com.fasterxml.jackson.module.jakarta.xmlbind.JakartaXmlBindAnnotationModule
+  - org.openrewrite.java.ChangeType:
+      oldFullyQualifiedTypeName: com.fasterxml.jackson.module.jaxb.JaxbAnnotationIntrospector
+      newFullyQualifiedTypeName: com.fasterxml.jackson.module.jakarta.xmlbind.JakartaXmlBindAnnotationIntrospector
+  - org.openrewrite.java.ChangeType:
+      oldFullyQualifiedTypeName: com.fasterxml.jackson.jaxrs.json.JacksonJsonProvider
+      newFullyQualifiedTypeName: com.fasterxml.jackson.jakarta.rs.json.JacksonJsonProvider
```
timtebeek: Thanks a lot for this proposed fix; Seeing the lines just above also changing classes from those same packages I wonder if it would make sense to swap out these `ChangeType` instances for `ChangePackage`. Any thoughts there? Are there classes in the old packages that we should not convert similarly?

CalvinZengyi: Thanks for the rapid response!

Good question. I did a little bit more digging, and it appears that it may not be feasible to generalize these into `ChangePackage`. Existing line 1052-1054 is an example (see screenshot as well): the class name has also changed.

btw, let me know if you think it would be helpful to add a few more lines to cover all the classes in the screenshot  .

image

timtebeek: Thanks for the helpful screenshot! We could then still use ChangePackage to handle the cases where the class name hasn't changed, which I've applied just now. With that we're good to merge, thanks!

CalvinZengyi: Thank you!
(and wow, I learned a new trick on writing recipes from you today!)

PR #946, opened 2025-11-19
Update jakarta-ee-10.yml

ChenyuWang98:
### Checklist


timtebeek (approved):
Thanks for expanding on the recipes here! I've applied a few small polishing changes, to for instance expand the result of previous recipes executed in sequence.

PR #952, opened 2025-11-28
update geotools 20 to 34

ChenyuWang98:
### Checklist

### DIFF
--- a/src/main/resources/META-INF/rewrite/geotools-34.yml
+++ b/src/main/resources/META-INF/rewrite/geotools-34.yml
@@ -1,3 +1,19 @@
+#
+# Copyright 2025 the original author or authors.
+# <p>
+# Licensed under the Moderne Source Available License (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+# <p>
+# https://docs.moderne.io/licensing/moderne-source-available-license
+# <p>
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
 ---
 type: specs.openrewrite.org/v1beta/recipe
 name: org.geotools.20migrateTo34


src/main/resources/META-INF/rewrite/geotools-34.yml
```diff
@@ -0,0 +1,192 @@
...
+      newFullyQualifiedTypeName: org.geotools.api.data.DataStoreFinder
+  - org.openrewrite.java.ChangeType:
+      oldFullyQualifiedTypeName: org.geotools.data.DataAccess
+      newFullyQualifiedTypeName: org.geotools.api.data.DataAccess
+  - org.openrewrite.java.ChangeType:
+      oldFullyQualifiedTypeName: org.geotools.data.DataStore
+      newFullyQualifiedTypeName: org.geotools.api.data.DataStore
+  - org.openrewrite.java.ChangeType:
+      oldFullyQualifiedTypeName: org.geotools.data.FileDataStoreFinder
+      newFullyQualifiedTypeName: org.geotools.api.data.FileDataStoreFinder
```
timtebeek, line 148: I wonder if all of these can be replaced with 
```
  - org.openrewrite.java.ChangePackage:
      oldPackageName: org.geotools.data
      newPackageName: org.geotools.data.api
      recursive: false
```

src/main/resources/META-INF/rewrite/geotools-34.yml
```diff
@@ -0,0 +1,192 @@
...
+  - org.openrewrite.java.ChangeType:
+      oldFullyQualifiedTypeName: org.geotools.api.data.Join
+      newFullyQualifiedTypeName: org.geotools.data.Join
```
timtebeek, line 151: Judging by the others I'm seeing, isn't this the wrong way round?
```suggestion
  - org.openrewrite.java.ChangeType:
      oldFullyQualifiedTypeName: org.geotools.data.Join
      newFullyQualifiedTypeName: org.geotools.api.data.Join
```

src/main/resources/META-INF/rewrite/geotools-34.yml
ChenyuWang98: The reason why I list it this way is that what is not listed does not need to be modified.

src/main/resources/META-INF/rewrite/geotools-34.yml
ChenyuWang98: This might be the class I previously identified that didn't need modification; we can remove it for now.

ChenyuWang98:
Okay, I'll ask the geotools project.

PR #955, opened 2025-12-10
Lombok: add `onMethod_` argument If getter/setter method is annotated

zakaria-shahen:
resolve openrewrite/rewrite/issues/5015 

I think supporting `onMethod_` for getter/setter recipe would be helpful.


sambsnyd (approved):
Nice work, very useful improvement

PR #961, opened 2026-01-05
Add bulk read method to InputStream implementations

pstreef:
## Problem

Java's default `InputStream.read(byte[], int, int)` implementation calls the single-byte `read()` in a loop, causing up to 350x slower performance for bulk reads. Many InputStream subclasses only override the single-byte `read()` method but not the bulk read method.

- See openrewrite/rewrite#6383 for more context and original discussion.

## Solution

This recipe detects InputStream subclasses that:
- Override `read()` but not `read(byte[], int, int)`  
- Delegate to another `InputStream`

For simple delegation patterns, it adds the missing bulk read method. For complex bodies (side effects, transformations), it adds a search marker for manual review.

Handles both anonymous and named classes, ternary and if-statement null check styles, and skips FilterInputStream subclasses.

- Moved from openrewrite/rewrite#6383 per review feedback from @timtebeek suggesting this belongs in rewrite-migrate-java's `org.openrewrite.java.migrate.io` package.

### DIFF
--- a/src/main/resources/META-INF/rewrite/examples.yml
+++ b/src/main/resources/META-INF/rewrite/examples.yml
@@ -1144,6 +1144,27 @@ examples:
 type: specs.openrewrite.org/v1beta/example
 recipeName: org.openrewrite.java.migrate.UpgradePluginsForJava11
 examples:
+- description: '`Java8toJava11Test#needToJaxb2MavenPlugin`'
+  sources:
+  - before: project
+    language: mavenProject
+  - before: |
+      <project>
+        <groupId>com.mycompany.app</groupId>
+        <artifactId>my-app</artifactId>
+        <version>1</version>
+        <build>
+          <plugins>
+            <plugin>
+              <groupId>org.codehaus.mojo</groupId>
+              <artifactId>jaxb2-maven-plugin</artifactId>
+              <version>2.3.1</version>
+            </plugin>
+          </plugins>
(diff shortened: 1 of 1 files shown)


src/main/java/org/openrewrite/java/migrate/io/AddInputStreamBulkReadMethod.java
```diff
@@ -0,0 +1,483 @@
...
+        return "Adds a `read(byte[], int, int)` method to `InputStream` subclasses that only override the single-byte " +
+                "`read()` method. Java's default `InputStream.read(byte[], int, int)` implementation calls the " +
+                "single-byte `read()` method in a loop, which can cause severe performance degradation (up to 350x " +
+                "slower) for bulk reads. This recipe detects `InputStream` implementations that delegate to another " +
+                "stream and adds the missing bulk read method to delegate bulk reads as well.";
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return Preconditions.check(new DeclaresType<>(JAVA_IO_INPUT_STREAM, true), new JavaIsoVisitor<ExecutionContext>() {
```
timtebeek, line 56: This required a small change upstream to allow the precondition to be used: https://github.com/openrewrite/rewrite/commit/e91804ae8e2acb2d908a301317963652ff3b0859

PR #965, opened 2026-01-11
Use `var` for variables initialized with type casts

timtebeek:
## What's changed?
Added a new recipe with very limited scope in applying `var` only where initialized by a type cast within methods.

## What's your motivation?
This seemed like a simple and safe, non-controversial usage of `var` that we can then adopt in our tests, which frequently unpack and cast.

Any edge cases where this might be unsafe? 

Could have also tackled static initializer blocks, but that seemed rare to me to have casts there.


MBoegers (approved):
LGTM, great to see another edge case closed.

PR #975, opened 2026-01-27
Avoid line continuation escapes in text blocks with mixed newline content

Jenson3210:
## Summary

- When `UseTextBlocks` converts a string concatenation that mixes newline-terminated strings with non-newline strings, it no longer inserts `\` line continuations for the non-newline joins
- Instead, those parts are placed on the same text block line
- Line continuations are still used when the entire content has no newlines, where they provide useful visual structure

## Problem

When converting SQL query string concatenations to text blocks, the recipe inserts `\` (line continuation) escapes between adjacent strings that aren't separated by `\n`. In a mixed concatenation like:

```java
String query = "select count(1)\n" +
               "from my_table\n" +
               "where a.id = b.id\n" +
               "and a.stat_cd = 'ACTV'" +
               " and a.del_fl = 'N'" +
               " and a.rgn = :region";
```

The recipe would produce unexpected `\` characters:

```java
String query = """
               ...
               and a.stat_cd = 'ACTV'\
                and a.del_fl = 'N'\
                and a.rgn = :region""";
```

While semantically correct, the `\` line continuation is unfamiliar to many developers and perceived as a bug.

## Solution

Added a `contentHasNewlines` guard to the passPhrase insertion logic. When the concatenated content already contains interior newlines (mixed case), non-newline-joined strings are simply concatenated on the same text block line instead of being split with `\` continuations:

```java
String query = """
               ...
               and a.stat_cd = 'ACTV' and a.del_fl = 'N' and a.rgn = :region""";
```

## Test plan

- [x] Existing tests pass (all 27 UseTextBlocksTest cases)
- [x] New tests: `noLineContinuationWhenContentHasNewlines` — trailing non-newline strings
- [x] New tests: `noLineContinuationInMixedConcatenation` — end-of-query non-newline strings
- [x] New tests: `noLineContinuationInMiddleOfMixedConcatenation` — mid-query non-newline join

- Fixes moderneinc/customer-requests#1763

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/lang/UseTextBlocks.java
+++ b/src/main/java/org/openrewrite/java/migrate/lang/UseTextBlocks.java
@@ -52,12 +52,27 @@ public class UseTextBlocks extends Recipe {
             required = false)
     boolean convertStringsWithoutNewlines;
 
+    @Option(displayName = "Whether to avoid line continuation escape sequences.",
+            description = "When enabled, the recipe avoids using `\\` line continuation escapes in text blocks " +
+                          "where the content contains newlines. Non-newline-joined strings are placed on the same " +
+                          "text block line instead. The default value is false.",
+            example = "true",
+            required = false)
+    boolean avoidLineContinuations;
+
     public UseTextBlocks() {
         convertStringsWithoutNewlines = true;
+        avoidLineContinuations = false;
     }
 
     public UseTextBlocks(boolean convertStringsWithoutNewlines) {
         this.convertStringsWithoutNewlines = convertStringsWithoutNewlines;
+        this.avoidLineContinuations = false;
+    }
+
+    public UseTextBlocks(boolean convertStringsWithoutNewlines, boolean avoidLineContinuations) {
+        this.convertStringsWithoutNewlines = convertStringsWithoutNewlines;
+        this.avoidLineContinuations = avoidLineContinuations;
     }
 
     String displayName = "Use text blocks";
@@ -122,6 +137,7 @@ public class UseTextBlocks extends Recipe {
                 stringLiterals = stringLiterals.stream()
                         .filter(s -> s.getValue() != null && !s.getValue().toString().isEmpty())
                         .collect(toList());
+                boolean skipLineContinuations = avoidLineContinuations && containsNewLineInContent(content);
                 for (int i = 0; i < stringLiterals.size(); i++) {
                     String s = requireNonNull(stringLiterals.get(i).getValue()).toString();
                     sb.append(s);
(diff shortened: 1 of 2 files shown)


src/test/java/org/openrewrite/java/migrate/lang/UseTextBlocksTest.java
```diff
@@ -881,6 +881,96 @@ void shouldNotUpdateKotlinCode() {
...
+              """,
+            """
+              class Test {
+                  String query = \"""
+                                 select count(1) cls_count
+                                 from my_table
+                                 where a.id = b.id
+                                 and a.ts > b.ts
+                                 and a.id = :tag_id\\s
+                                 and a.stat_cd = 'ACTV' and a.del_fl = 'N' and a.rgn = :region\""";
```
timtebeek, line 909: I worry a bit about changing a default here and stripping out intentional newlines the developer had added. Can we make it an option they have to explicitly opt in to?

In the examples here with short lines concatenated looks _ok_, but if these lines were larger I'd prefer the smaller diff on continuation chars by default, and see it as an opportunity to educate developers on how to read & write text blocks.

Jenson3210: Have to say I agree here. 
That's why I made the comment also myself. 

We could add a flag to enable this behavior indeed.

Jenson3210: @timtebeek , added this as a flag (default behavior = previous behavior)

src/main/java/org/openrewrite/java/migrate/lang/UseTextBlocks.java
```diff
@@ -52,12 +52,27 @@ public class UseTextBlocks extends Recipe {
...
+
     public UseTextBlocks() {
         convertStringsWithoutNewlines = true;
+        avoidLineContinuations = false;
     }
 
     public UseTextBlocks(boolean convertStringsWithoutNewlines) {
         this.convertStringsWithoutNewlines = convertStringsWithoutNewlines;
+        this.avoidLineContinuations = false;
+    }
```
timtebeek:
```suggestion
```

src/test/java/org/openrewrite/java/migrate/lang/UseTextBlocksTest.java
timtebeek: Thanks! Did we also already have tests that verify adding line continuations by default under normal circumstances? Might be good to make sure we don't lose that ability if we didn't already have such tests.

PR #987, opened 2026-02-17
Retain javax.xml.bind:jaxb-api when jackson-module-jaxb-annotations is present

### LINKED ISSUE #504: AddJaxbDependencies removes `javax.xml.bind:jaxb-api` when it is required by another dependency
Moderne SaaS

Moderne SaaS

```
api("org.openrewrite:rewrite-maven:latest.integration")
implementation('javax.xml.bind:jaxb-api:latest.release')
implementation("net.logstash.logback:logstash-logback-encoder:latest.release")
```

```
api("org.openrewrite:rewrite-maven:latest.integration")
implementation('javax.xml.bind:jaxb-api:latest.release')
implementation('jakarta.xml.bind:jakarta.xml.bind-api:4.0.2') {
    because("required by: rewrite-maven --> jackson-module-jaxb-annotations")
}
implementation("net.logstash.logback:logstash-logback-encoder:latest.release")
```

## What did you see instead?

```
api("org.openrewrite:rewrite-maven:latest.integration")
implementation('jakarta.xml.bind:jakarta.xml.bind-api:4.0.2') {
    because("required by: rewrite-maven --> jackson-module-jaxb-annotations")
}
implementation("net.logstash.logback:logstash-logback-encoder:latest.release")
```

```
java.lang.NoClassDefFoundError: javax/xml/bind/annotation/XmlElement
```

Jenson3210:
## Summary

- Adds `RetainJaxbApiForJackson` recipe that re-adds `javax.xml.bind:jaxb-api` as a runtime dependency when `jackson-module-jaxb-annotations` is on the classpath
- Wires the recipe into `JavaxXmlBindMigrationToJakartaXmlBind` in `jakarta-ee-9.yml`, after the `ChangeDependency` step
- Adds tests for both Gradle and Maven, plus a negative test when Jackson is not present

## Problem

When running the Jakarta EE 9 migration (`JavaxXmlBindMigrationToJakartaXmlBind`), `ChangeDependency` replaces `javax.xml.bind:jaxb-api` with `jakarta.xml.bind:jakarta.xml.bind-api:3.0.x`. However, `jackson-module-jaxb-annotations` transitively depends on `jakarta.xml.bind:jakarta.xml.bind-api:2.3.x` (the bridge version that still provides the `javax.xml.bind` namespace). Version resolution picks the highest version (3.0.x), which no longer provides `javax.xml.bind` classes, causing `NoClassDefFoundError` at runtime.

## Solution

A new imperative recipe `RetainJaxbApiForJackson` checks if `jackson-module-jaxb-annotations` is on the resolved classpath (for both Gradle and Maven). If found, it adds `javax.xml.bind:jaxb-api:2.3.x` as a `runtimeOnly`/`runtime` dependency so the javax namespace classes remain available.

## Test plan

- [x] New tests reproduce the bug on unmodified `main` (both Gradle and Maven assertions fail without the fix)
- [x] New tests pass with the fix applied
- [x] Negative test confirms `jaxb-api` is NOT retained when Jackson is absent
- [x] Existing `AddJaxbDependenciesWithRuntimeTest`, `AddJaxbDependenciesWithoutRuntimeTest`, and `JavaxToJakartaTest` all pass (no regressions)

- Fixes #504


src/test/java/org/openrewrite/java/migrate/jakarta/JavaxXmlBindMigrationToJakartaXmlBindTest.java
```diff
@@ -0,0 +1,216 @@
...
+              """,
+            spec -> spec.after(pom -> {
+                assertThat(pom)
+                  .contains("jakarta.xml.bind")
+                  .contains("jakarta.xml.bind-api")
+                  .contains("javax.xml.bind")
+                  .contains("jaxb-api")
+                  .contains("jackson-module-jaxb-annotations");
+                return pom;
+            })
```
timtebeek: Nit: try to use `actual()`; optional though, as we also have a recipe to fix it, but figured good to point out in case you had hand written this test. Otherwise we can pick it up with the automatically applied recipes.
```suggestion
            spec -> spec.after(pom ->
                assertThat(pom)
                  .contains("jakarta.xml.bind")
                  .contains("jakarta.xml.bind-api")
                  .contains("javax.xml.bind")
                  .contains("jaxb-api")
                  .contains("jackson-module-jaxb-annotations")
                  .actual())
```

src/main/resources/META-INF/rewrite/jakarta-ee-9.yml
```diff
@@ -829,6 +829,7 @@ recipeList:
       newGroupId: jakarta.xml.bind
       newArtifactId: jakarta.xml.bind-api
       newVersion: 3.0.x
+  - org.openrewrite.java.migrate.jakarta.RetainJaxbApiForJackson
```
timtebeek, line 832: I would have expected a declarative recipe here with a precondition, which would then be able to use the Maven/Gradle agnostic recipes. Any reason that wasn't an option?

Jenson3210: I considered a recipe with only if using, but not one with ModuleHasDependency precondition. That might work also. Let me verify

src/test/java/org/openrewrite/java/migrate/jakarta/JavaxXmlBindMigrationToJakartaXmlBindTest.java
Jenson3210: I did not recall this existed (but in this case I did not write the test)

src/main/resources/META-INF/rewrite/jakarta-ee-9.yml
Jenson3210: had it reworked. It's all declarative now (kept the unit test)

PR #990, opened 2026-02-19
`UpdateSdkMan` recipe to support `newJavaVersion=latest.patch`

greg-at-moderne:
Please review the actual proposed interface to the recipe. Is specifying `sameMajor` String to `newJavaVersion` option a decent way of handling this? My hunch is it's not, but not sure what's the better way.

## What's changed?

Make `UpdateSdkMan` recipe support `sameMajor` as a possible value to `newJavaVersion` option.
If specified, it will update Java within the same major version, using the same distribution (unless specified otherwise).

## What's your motivation?

Have a way to quickly update all the sdkmanrc files to use unified versions. And using latest version within the same major seems to make sense.

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/UpdateSdkMan.java
+++ b/src/main/java/org/openrewrite/java/migrate/UpdateSdkMan.java
@@ -43,7 +43,7 @@ import static java.util.stream.Collectors.toList;
 public class UpdateSdkMan extends Recipe {
 
     @Option(displayName = "Java version",
-            description = "The Java version to update to.",
+            description = "The Java version to update to. Use `latest.patch` to upgrade to the latest version within the current major version.",
             required = false,
             example = "17")
     @Nullable


src/main/java/org/openrewrite/java/migrate/UpdateSdkMan.java
```diff
@@ -85,7 +85,14 @@ public Tree visit(@Nullable Tree tree, ExecutionContext ctx) {
                 Pattern pattern = Pattern.compile("java=(.*?)([.a-z]*-.*)");
                 Matcher matcher = pattern.matcher(plainText.getText());
                 if (matcher.find()) {
-                    String ver = newVersion == null ? matcher.group(1) : newVersion;
+                    String ver;
+                    if ("sameMajor".equalsIgnoreCase(newVersion)) {
```
timtebeek: https://docs.openrewrite.org/reference/dependency-version-selectors#latest-release-version
```suggestion
                    if ("latest.patch".equalsIgnoreCase(newVersion)) {
```

src/test/java/org/openrewrite/java/migrate/UpdateSdkManTest.java
```diff
@@ -199,6 +199,58 @@ void upgradeIfNewVersionIsStillSameVersionBasisAndDifferentDistribution() {
...
+                .endsWith("-amzn")
+                .actual())
+          )
+        );
+    }
+
+    @Test
+    void minorDoesNotCrossMajorVersion() {
+        rewriteRun(
+          spec -> spec.recipe(new UpdateSdkMan("sameMajor", null)),
```
timtebeek:
```suggestion
          spec -> spec.recipe(new UpdateSdkMan("latest.patch", null)),
          text(
            """
              java=21.0.6-zulu
              """,
            spec -> spec.path(".sdkmanrc")
              .after(str -> assertThat(str)
                .startsWith("java=21.0.")
                .doesNotContain("21.0.6")
                .endsWith("-zulu")
                .actual())
          )
        );
    }

    @Test
    void minorUpgradeWithNewDistribution() {
        rewriteRun(
          spec -> spec.recipe(new UpdateSdkMan("latest.patch", "amzn")),
          text(
            """
              java=21.0.6-zulu
              """,
            spec -> spec.path(".sdkmanrc")
              .after(str -> assertThat(str)
                .startsWith("java=21.0.")
                .endsWith("-amzn")
                .actual())
          )
        );
    }

    @Test
    void minorDoesNotCrossMajorVersion() {
        rewriteRun(
          spec -> spec.recipe(new UpdateSdkMan("latest.patch", null)),
```

PR #1005, opened 2026-03-09
Fix `IfElseIfConstructToSwitch` dropping pattern variables for unresolvable types

### LINKED ISSUE #1004: `IfElseIfConstructToSwitch` fails to add variable declarations for custom types used in `instanceof`
- Following https://github.com/openrewrite/rewrite-migrate-java/pull/994
- We saw https://github.com/moderneinc/rewrite-migrate-kotlin/commit/4a05bb4d6959fa7196176f4597536d2030aba36f#diff-d0136ee922ed73e329538416ecf67d2e45a4b3d8a63287b96e6fb8b0975c2e30 introduce missing variables in cases
  Image
- Fixed manually in https://github.com/moderneinc/rewrite-migrate-kotlin/commit/e73206c3f15af367dd640c07abc12000ce76baf0
  Image

timtebeek:
## Summary
- When `IfElseIfConstructToSwitch` converts if-else-if chains with `instanceof` checks on non-JDK types, JavaTemplate's raw string substitution (`#{}`) can fail to produce `J.VariableDeclarations` AST nodes, causing pattern variable declarations to be dropped (e.g., `case Dog d ->` becomes `case Dog ->`)
- Uses positional case identification (null case offset + pattern case count) instead of type-checking labels, ensuring the `instanceOfs` iterator always advances correctly for pattern match cases
- Adds a `buildVariableDeclarations()` fallback that reconstructs a proper `J.VariableDeclarations` from the original `J.InstanceOf` when JavaTemplate fails to produce one

## Test plan
- [x] All 14 existing `IfElseIfConstructToSwitchTest` tests pass (including `nullCheckWithNonJdkTypes` and `threeNonJdkTypes`)
- [x] No regressions in other tests

- Fixes #1004

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/lang/IfElseIfConstructToSwitch.java
+++ b/src/main/java/org/openrewrite/java/migrate/lang/IfElseIfConstructToSwitch.java
@@ -235,13 +235,11 @@ public class IfElseIfConstructToSwitch extends Recipe {
             // Cases are ordered: [null case (optional)], [pattern cases...], [default case]
             int nullCaseOffset = nullCheckedParameter != null ? 1 : 0;
             int patternCaseCount = patternMatchers.size();
-            int[] index = {0};
             return switch_.withCases(switch_.getCases().withStatements(
-                    ListUtils.map(switch_.getCases().getStatements(), stmt -> {
+                    ListUtils.map(switch_.getCases().getStatements(), (currentIndex, stmt) -> {
                         if (!(stmt instanceof J.Case)) {
                             return stmt;
                         }
-                        int currentIndex = index[0]++;
                         int patternIndex = currentIndex - nullCaseOffset;
                         if (patternIndex < 0 || patternIndex >= patternCaseCount || !instanceOfs.hasNext()) {
                             return stmt; // null case or default case


src/main/java/org/openrewrite/java/migrate/lang/IfElseIfConstructToSwitch.java
```diff
@@ -225,36 +227,79 @@ public J.Return visitReturn(J.Return return_, AtomicBoolean atomicBoolean) {
...
-                                // Fix variable type from original pattern
-                                if (instanceOf.getPattern() instanceof J.Identifier && !varDecl.getVariables().isEmpty()) {
-                                    J.Identifier originalPattern = (J.Identifier) instanceOf.getPattern();
-                                    J.VariableDeclarations.NamedVariable var0 = varDecl.getVariables().get(0);
-                                    varDecl = varDecl.withVariables(singletonList(
-                                            var0.withType(originalPattern.getType())
-                                                    .withName(var0.getName().withType(originalPattern.getType()))));
-                                }
-                                return case_.withCaseLabels(singletonList(varDecl.withPrefix(case_.getCaseLabels().get(0).getPrefix())));
+                        if (!(stmt instanceof J.Case)) {
```
Jenson3210, line 240: instead of this `int[]` approach, would we not want to use `ListUtils.map(list, (stmt, index) -> {...})`?

Jenson3210 (approved):
Left a question/comment. Other that I think it looks okay. Too bad we do not have the original types of the code at hand during JavaTemplate generation as that would make this not necessary.

Issue #1023, opened 2026-03-23
FacesManagedBeansRemoved for Faces 4.0 recipe updates


evie-lau:
# Faces 4.0 recipe - `FacesManagedBeansRemoved` - needs additions 

I found a few things the [org.openrewrite.java.migrate.jakarta.FacesManagedBeansRemoved](https://github.com/openrewrite/rewrite-migrate-java/blob/e663e7be87fbc47ff04574746d75e3501125bed5/src/main/resources/META-INF/rewrite/jakarta-faces-4.yml#L221) recipe is missing.

## Reference documents
- [Previous Jakarta Faces 3.0 docs](https://jakarta.ee/specifications/faces/3.0/apidocs/jakarta/faces/bean/package-summary) shows replacements needed. Most are currently covered
- Additional docs on migration to Faces 4.0 for EE10: https://docs.redhat.com/en/documentation/red_hat_jboss_enterprise_application_platform/8.0/html/migration_guide/application-migration-from-jakarta-ee-8-to-ee-10_default#faces-managed-beans_default

### 1. CustomScoped

<img width="814" height="57" alt="Image" src="https://github.com/user-attachments/assets/c978730e-0234-4894-b546-6af912bfbc30" />

- `java.faces.bean.CustomScoped` and `jakarta.faces.bean.CustomScoped` seem to need a custom recipe to use CDI custom scopes and `jakarta.enterprise.context.spi.Context`. Unsure of recipe scope, and if feasible.

### 2. ManagedProperty name
[Faces 3.0 ManagedProperty doc](https://jakarta.ee/specifications/faces/3.0/apidocs/jakarta/faces/bean/managedproperty)
[Faces 4.0 ManagedProperty doc](https://jakarta.ee/specifications/faces/4.0/apidocs/jakarta/faces/annotation/managedproperty)
- `name` was an optional element in Faces 3.0, and removed in Faces 4.0. Possibly remove entirely unless there is another migration path to maintain functionality.

I've seen one migration case with this build error after running the recipes:
```
cannot find symbol
[ERROR]   symbol:   method name()
[ERROR]   location: @interface jakarta.faces.annotation.ManagedProperty
```

salaheddinemchich:
Hi, I'd like to work on this issue. Is it still available and are there any specific guidelines I should follow?

timtebeek:
[...]

PR #1025, opened 2026-03-23
Add notes for Faces 4.0 FacesManagedBeansRemoved

evie-lau:
## What's changed?
Add notes in recipe for future enhancements

## What's your motivation?
- Inline comments for #1023


evie-lau:
> Does that removal through `RemoveAnnotationAttribute` make sense to you? Or would that lose something we still need to migrate?

I have no personal experience with usage of these annotations, but it seems to make sense to me.
The current case is compilation error. And with removing the attribute, it seems the name is automatically based on the annotated field's name.

PR #1027, opened 2026-03-23
Fix order for ManagedProperty migration in Faces 4.0

evie-lau:
- Followup to #1025

Fix order of recipes to ensure `name` attribute removal for migrated `@ManagedProperty`


timtebeek:
Doh! Thanks for the quick fix.

PR #1030, opened 2026-03-27
Add sealed class modifier recipe for Java 17+

timtebeek:
## Summary
- Adds `AddSealedClassModifier` recipe that applies the `sealed` keyword to classes and interfaces whose only subclasses/implementations are nested within the same class declaration and whose constructors are all private.
- Uses a two-pass `ScanningRecipe`: first pass builds a complete superclass→subclass map across the entire codebase, second pass seals only classes where the nested subclass set exactly matches all known subclasses (zero false-positive risk).
- Also handles sealed interfaces with nested record implementors.
- Included in the `UpgradeToJava17` recipe chain.

## Test plan
- 12 unit tests covering positive cases (classes, abstract classes, interfaces with records) and no-change cases (public/protected/default constructors, no subclasses, already sealed, external subclasses, non-final subclasses, Java < 17).


.gitignore
```diff
@@ -10,3 +10,5 @@ out/
 bin/
 .claude/settings.local.json
 working-set*/
+working-set-square*/
+working-set-spring*/
```
timtebeek:
```suggestion
```

PR #1031, opened 2026-03-30
Only add lombok annotation processor to modules that use lombok

Jenson3210:
## Summary

- Changed `EnableLombokAnnotationProcessor` precondition from `RepositoryHasDependency` to `DependencyInsight`
- Added tests covering the multi-module scenario where only some modules use lombok

## Problem

`EnableLombokAnnotationProcessor` used `RepositoryHasDependency` as its precondition, which checks the **entire repository**. If any module has lombok as a dependency, the annotation processor gets added to ALL modules — including those that don't use lombok. This causes unexpected `maven-compiler-plugin` configuration changes in projects without lombok.

## Solution

Switch the precondition from `RepositoryHasDependency` (repo-wide) to `DependencyInsight` (per-module). This matches the pattern already used by `AddLombokMapstructBinding` and ensures the annotation processor is only added to modules that actually declare lombok as a dependency.

## Test plan

- [x] Existing `UpgradeToJava25Test.addsLombokAnnotationProcessor` passes (positive case)
- [x] Existing `UpgradeToJava17Test` passes
- [x] New test: `doesNotAddLombokToModuleWithoutLombok` — multi-module project where one module has lombok and another doesn't; only the lombok module gets the annotation processor
- [x] New test: `doesNotAddLombokWhenNotPresent` — single module without lombok gets no changes
- [x] New test: `doesNotAddLombokWhenNotPresentViaUpgradeToJava25` — same via composite recipe

- Fixes moderneinc/customer-requests#2110

### DIFF
--- a/src/test/java/org/openrewrite/java/migrate/EnableLombokAnnotationProcessorTest.java
+++ b/src/test/java/org/openrewrite/java/migrate/EnableLombokAnnotationProcessorTest.java
@@ -19,6 +19,7 @@ import org.junit.jupiter.api.Test;
 import org.openrewrite.test.RecipeSpec;
 import org.openrewrite.test.RewriteTest;
 
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.openrewrite.java.Assertions.mavenProject;
 import static org.openrewrite.maven.Assertions.pomXml;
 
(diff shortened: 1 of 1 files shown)


src/test/java/org/openrewrite/java/migrate/EnableLombokAnnotationProcessorTest.java
```diff
@@ -0,0 +1,148 @@
...
+                    <dependencies>
+                        <dependency>
+                            <groupId>org.projectlombok</groupId>
+                            <artifactId>lombok</artifactId>
+                            <version>1.18.40</version>
+                        </dependency>
+                    </dependencies>
+                </project>
+                """,
+              spec -> spec.after(actual -> actual)
```
timtebeek: Should we validate the annotation processor was added here?

PR #1035, opened 2026-03-31
Cap Java version at 24 for Kotlin <2.3 in UpgradeToJava25

timtebeek:
## Summary
- Cap Java version at 24 when a project uses Kotlin <2.3 (which doesn't support Java 25)
- Replace the single `UpgradeJavaVersion(25)` call with two preconditioned recipes:
  - `UpgradeBuildToJava24` — applies when `kotlin-stdlib <2.3` is present
  - `UpgradeBuildToJava25` — applies otherwise (no Kotlin, or Kotlin >=2.3) via `ModuleHasDependency` with `invertMarking: true`
- Pure YAML solution, no custom Java code needed

- Fixes https://github.com/moderneinc/customer-requests/issues/2126

## Test plan
- [x] Kotlin <2.3 project gets Java 24
- [x] Kotlin >=2.3 project gets Java 25
- [x] Project without Kotlin gets Java 25
- [x] Existing tests (plugin upgrades, Gradle wrapper, Lombok annotation processor) still pass


src/main/java/org/openrewrite/java/migrate/HasNoOldKotlinDependency.java
```diff
@@ -0,0 +1,109 @@
...
+import org.openrewrite.maven.tree.Scope;
+
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+@Value
+@EqualsAndHashCode(callSuper = false)
+public class HasNoOldKotlinDependency extends ScanningRecipe<HasNoOldKotlinDependency.Accumulator> {
```
timtebeek: Not happy with this; exploring `invertMarkings` on ModuleHasDependency` as an alternative.

Issue #934, opened 2025-11-09
MigrateFromJavaxAnnotationApi does not modify annotations to the type level


protocol7:
## What version of OpenRewrite are you using?

- org.openrewrite:rewrite-core:8.65.0
- org.openrewrite.recipe:rewrite-migrate-java:3.20.0

## What is the smallest, simplest way to reproduce the problem?

When arrays are annotated, the annotation is not adjusted to be at the type level when migrating to JSpecify. This causes the nullability analysis to fail with tools like NullAway. 

See https://jspecify.dev/docs/user-guide/#type-use-annotation-syntax for a reference on these annotations.

```java
  @Test
  void jspecifyArrayNullable() {
    final Environment env = Environment.builder().scanRuntimeClasspath().build();
    final Recipe recipe =
        env.activateRecipes("org.openrewrite.java.jspecify.MigrateFromJavaxAnnotationApi");
    rewriteRun(
        spec -> spec.recipe(recipe),
        // language=java
        java(
            """
            import javax.annotation.Nullable;

            class Foo {
                    @Nullable
                    byte[] bar() {
                        return null;
                    }

                    void baz(@Nullable byte[] a) {
                        return null;
                    }
            }
            """,
            """
            import org.jspecify.annotations.Nullable;

            class Foo {
                    byte @Nullable [] bar() {
                        return null;
                    }

                    void baz(byte @Nullable [] a) {
                        return null;
                    }
            }
            """));
  }
```

```
org.opentest4j.AssertionFailedError: [Unexpected result in "Foo.java":
diff --git a/Foo.java b/Foo.java
index 891ac0b..a4c2b3c 100644
--- a/Foo.java
+++ b/Foo.java
@@ -1,11 +1,12 @@
 import org.jspecify.annotations.Nullable;

 class Foo {
-        byte @Nullable [] bar() {
+        @Nullable
+        byte[] bar() {
             return null;
         }

-        void baz(byte @Nullable [] a) {
+        void baz(@Nullable byte[] a) {
             return null;
         }
 }
\ No newline at end of file
]
expected:
  "import org.jspecify.annotations.Nullable;

  class Foo {
          byte @Nullable [] bar() {
              return null;
          }

          void baz(byte @Nullable [] a) {
              return null;
          }
  }"
 but was:
  "import org.jspecify.annotations.Nullable;

  class Foo {
          @Nullable
          byte[] bar() {
              return null;
          }

          void baz(@Nullable byte[] a) {
[...]

PR #1038, opened 2026-03-31
Move type-use annotations to array brackets during JSpecify migration

timtebeek:
## Summary
- Fixes https://github.com/openrewrite/rewrite-migrate-java/issues/934

- Adds a new `MoveAnnotationToArrayType` recipe that moves type-use annotations (e.g. `@Nullable`) from declaration position to the array brackets (e.g. `@Nullable byte[]` → `byte @Nullable[] `), as required by the JSpecify spec.
- Wires the recipe into all 6 JSpecify migration recipes in `jspecify.yml`, running after `MoveFieldAnnotationToType`.
- Handles method return types, method parameters, and field declarations.

## Test plan
- [x] Unit tests for `MoveAnnotationToArrayType`: return types, parameters, fields, non-array no-ops, multi-dimensional arrays
- [x] Integration test in `JSpecifyBestPracticesTest` for end-to-end javax → JSpecify migration with array types
- [x] All existing jspecify tests continue to pass

### DIFF
--- a/src/main/resources/META-INF/rewrite/jspecify.yml
+++ b/src/main/resources/META-INF/rewrite/jspecify.yml
@@ -95,8 +95,8 @@ recipeList:
       version: 1.0.0
       onlyIfUsing: jakarta.annotation.*ull*
       acceptTransitive: true
-  - org.openrewrite.java.migrate.jspecify.MoveAnnotationToArrayType:
-      annotationType: jakarta.annotation.*ull*
+  # Not moving jakarta annotations to array brackets; they target TYPE_USE,
+  # so `@Nullable X[]` (element nullable) already differs from `X @Nullable[]` (array nullable).
   - org.openrewrite.java.ChangeType:
       oldFullyQualifiedTypeName: jakarta.annotation.Nullable
       newFullyQualifiedTypeName: org.jspecify.annotations.Nullable
@@ -121,8 +121,8 @@ recipeList:
       version: 1.0.0
       onlyIfUsing: org.jetbrains.annotations.*ull*
       acceptTransitive: true
-  - org.openrewrite.java.migrate.jspecify.MoveAnnotationToArrayType:
-      annotationType: org.jetbrains.annotations.*ull*
+  # Not moving JetBrains annotations to array brackets; they target TYPE_USE,
+  # so `@Nullable X[]` (element nullable) already differs from `X @Nullable[]` (array nullable).
   - org.openrewrite.java.ChangeType:
       oldFullyQualifiedTypeName: org.jetbrains.annotations.Nullable
       newFullyQualifiedTypeName: org.jspecify.annotations.Nullable
@@ -199,8 +199,8 @@ recipeList:
       version: 1.0.0
       onlyIfUsing: io.micronaut.core.annotation.*ull*
       acceptTransitive: true
-  - org.openrewrite.java.migrate.jspecify.MoveAnnotationToArrayType:
-      annotationType: io.micronaut.core.annotation.*ull*
+  # Not moving Micronaut annotations to array brackets; they target TYPE_USE,
+  # so `@Nullable X[]` (element nullable) already differs from `X @Nullable[]` (array nullable).
   - org.openrewrite.java.ChangeType:
       oldFullyQualifiedTypeName: io.micronaut.core.annotation.Nullable
       newFullyQualifiedTypeName: org.jspecify.annotations.Nullable


src/main/java/org/openrewrite/java/migrate/jspecify/MoveAnnotationToArrayType.java
```diff
@@ -0,0 +1,127 @@
...
+                          "Should target the pre-migration annotation type to avoid changing the semantics " +
+                          "of pre-existing type-use annotations on object arrays.",
+            example = "javax.annotation.*ull*")
+    String annotationType;
+
+    String displayName = "Move annotation to array type";
+
+    String description = "When an annotation like `@Nullable` is applied to an array type in declaration position, " +
+                         "this recipe moves it to the array brackets. " +
+                         "For example, `@Nullable byte[]` becomes `byte @Nullable[]`. " +
```
steve-aom-elliott, line 48: Is there the scenario to handle when the type of the elements in the array is a nested class? Pulled from the implementation below `J.@Nullable Annotation[]` is the ideal form, but do we need to consider if someone wrote `@Nullable J.Annotation[]`, or would that just not compile to begin with?

src/main/java/org/openrewrite/java/migrate/jspecify/MoveAnnotationToArrayType.java
```diff
@@ -0,0 +1,127 @@
...
+                    if (match[0] == null && matchesType(a)) {
+                        match[0] = a;
+                        return null;
+                    }
+                    return a;
+                });
+                if (leading == mv.getLeadingAnnotations()) {
+                    return mv;
+                }
+                mv = mv.withLeadingAnnotations(leading);
```
steve-aom-elliott, line 108: I suppose I'm trying to figure out if we have recipes that are going to be fragile in light of leading annotations being dropped and whether any of our recipes need to be adapted to handle the change in position for the annotations (have not had a chance to dig into this further yet)

src/main/java/org/openrewrite/java/migrate/jspecify/MoveAnnotationToArrayType.java
timtebeek: We were working through parallel PRs also related to type use annotations
- https://github.com/openrewrite/rewrite-static-analysis/pull/843
- https://github.com/openrewrite/rewrite-static-analysis/pull/845 

There we settled on recognizing a select few, which if folks were adding these to nested types already, I imagine they knew what they were doing and wouldn't want those moved to the array itself.

src/main/java/org/openrewrite/java/migrate/jspecify/MoveAnnotationToArrayType.java
timtebeek: Let me know what you'd need to be comfortable seeing this merged!

src/test/java/org/openrewrite/java/migrate/jspecify/MoveAnnotationToArrayTypeTest.java
```diff
@@ -204,6 +204,32 @@ class Foo {
...
+              import javax.annotation.Nullable;
+              import java.util.Map;
+
+              class Foo {
+                  public Map.Entry<String, String> @Nullable[] entries;
+              }
+              """
+          )
+        );
+    }
```
steve-aom-elliott, line 231: I think this test may actually be incorrect. Based on [this](https://docs.spring.io/spring-framework/reference/core/null-safety.html#:~:text=Arrays%20and%20varargs%20@Nullable%20Object%5B%5D%20array%20means,and%20the%20array%20can%20be%20null%20), these could be different meanings:

```java
@Nullable X[] x

// I believe the nested equivalent would be
y.@Nullable X[] x
```
means individual elements can be `null`, but not the array itself

```java
X @Nullable [] x

// I believe the nested equivalent would be
y.X @Nullable [] x
```
means individual elements cannot be `null`, but the array can

```java
@Nullable X @Nullable [] x

// I believe the nested equivalent would be
y.@Nullable X @Nullable [] x
```
means both individual elements and the array itself can be `null`

timtebeek: hmm; I thought it was different for the javax nullable annotation:
> javax.annotation.Nullable is NOT a TYPE_USE annotation — it only targets METHOD, FIELD, PARAMETER, etc. When someone writes @javax.annotation.Nullable Map.Entry[], the annotation is in declaration position — it doesn’t have the element-vs-array distinction. It’s syntactically equivalent to “this whole thing might be null.”

> The recipe deliberately targets pre-migration annotations (before ChangeType renames them to JSpecify). By the time the annotation becomes org.jspecify.annotations.Nullable (which IS TYPE_USE), it’s already in the correct array-bracket position with the intended meaning: “the array reference can be null.”

timtebeek: > The annotations targeted by `MoveAnnotationToArrayType` are:
> 
> 1. **`javax.annotation.*ull*`** — No `@Target` at all (JSR-305). Implicitly applies everywhere including TYPE_USE.
> 2. **`jakarta.annotation.*ull*`** — `@Target({METHOD, PARAMETER, FIELD, TYPE_USE})` — **yes, TYPE_USE**
> 3. **`org.jetbrains.annotations.*ull*`** — `@Target({METHOD, FIELD, PARAMETER, LOCAL_VARIABLE, TYPE_USE})` — **yes, TYPE_USE**
> 4. **`io.micrometer.core.lang.*ull*`** — `@Target({METHOD, FIELD, PARAMETER, LOCAL_VARIABLE})` — **no TYPE_USE**
> 5. **`org.springframework.lang.*ull*`** — `@Target({METHOD, PARAMETER, FIELD})` — **no TYPE_USE**
> 6. **`io.micronaut.core.annotation.*ull*`** — `@Target({METHOD, PARAMETER, FIELD, TYPE_USE, ANNOTATION_TYPE})` — **yes, TYPE_USE**
> 
> So **three of the six** (jakarta, JetBrains, Micronaut) are explicitly TYPE_USE annotations, and javax is implicitly usable in TYPE_USE positions. Only Micrometer and Spring are purely declaration-position.
> 
> Your concern is valid for at least half the migration paths — for those annotations, `@Nullable X[]` on the element type genuinely has different semantics than `X @Nullable[]` on the array. The recipe is making a deliberate semantic choice to reposition the annotation to mean "array can be null" rather than "elements can be null."

steve-aom-elliott: You could be correct on that. I guess I'm just not as familiar with the intricacies of different targets for annotations and how they affect the interpretation of the statement. Given we're targetting specific sets of annotations, it's probably fine, if just possible to misuse given there aren't target guards in the recipe if you were for example to give it JSpecify annotation FQNs to start with. Would it make sense to just guard against asking to move JSpecify annotations to JSpecify annotations?

PR #1045, opened 2026-04-05
Replace Collections.emptyXXX with Immutable Static Factory Methods

### LINKED ISSUE #1044: Replace Collections.emptyXXX with Immutable Static Factory Methods
- This is similar to https://github.com/openrewrite/rewrite-migrate-java/issues/72

Replace
```java
List<String> list = Collections.emptyList()
Map<String,String> map = Collections.emptyMap();
Set<String> set = Collections.emptySet();
```
With
```java
List<String> list =  List.of()
Map<String,String> map = Map.of();
Set<String> set  = Set.of();
```

mkarg:
- Closes #1044


src/main/java/org/openrewrite/java/migrate/util/MigrateCollectionsEmptyList.java
```diff
@@ -0,0 +1,60 @@
...
+                new UsesMethod<>(EMPTY_LIST));
+        return Preconditions.check(check, new JavaIsoVisitor<ExecutionContext>() {
+
+            @Override
+            public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, ExecutionContext ctx) {
+                J.MethodInvocation m = super.visitMethodInvocation(method, ctx);
+
+                if (EMPTY_LIST.matches(m)) {
+                    maybeRemoveImport("java.util.Collections");
+                    maybeAddImport("java.util.List");
```
mkarg, line 48: @timtebeek As you added `.imports("java.util.List")` I wonder if `maybeAddImport` is still needed?

timtebeek: Ah yes, those are regrettably separate:
- `.imports` ensures the import is present when we compile the JavaTemplate snippet in isolation
- `maybeAddImport` schedules a visitor to add the import the the LST we're visiting when we're done

PR #1062, opened 2026-04-15
Support `copyOf` migration for Guava

### LINKED ISSUE #584: Recipes to replace Guava `Immutable{Set|List|Map}.copyOf()` calls with Java `{Set|List|Map}.copyOf()` calls

greg-at-moderne:
## What's changed?

Adding three new building recipes for No Guava recipe:
- NoGuavaImmutableListCopyOf
- NoGuavaImmutableSetCopyOf
- NoGuavaImmutableMapCopyOf

modelled after (or actually now sharing code with) the existing:
- NoGuavaImmutableListOf
- NoGuavaImmutableMapOf
- NoGuavaImmutableSetOf

## What's your motivation?

- fixes #584


steve-aom-elliott:
Looking at the comments on
- https://github.com/openrewrite/rewrite-migrate-java/pull/585

(a previous attempt), would it have made more sense to have an `AbstractNoGuavaImmutable` that has the generic logic and intermediate layers being `AbstractNoGuavaImmutableOf` and `AbstractNoGuavaImmutableCopyOf` that would provide the `copyOf` or `of` method name rather than bringing that down to the individual recipes and changing the signature of the `AbstractNoGuavaImmutableOf`?

timtebeek (approved):

greg-at-moderne:
> intermediate layers

To be honest, that was one of the options I have considered. But this looked like somewhat artificial exercise in OOP.
Happy to adapt to it, if we agree on this.

greg-at-moderne:
Merging, we can always consider refaxtoring the code later on.

PR #1063, opened 2026-04-16
Rename `_` identifier to `__` for Java ≤ 8 source files

Jenson3210:
## Summary

- Add `RenameUnderscoreIdentifier` recipe that renames single-underscore identifiers (`_`) to double-underscore (`__`) in Java source files with source compatibility ≤ Java 8
- Handles all identifier positions: local variables, method parameters, fields, method names (declaration + call sites), member references, and class names
- Gated by `UsesJavaVersion(1, 8)` precondition — no changes on Java 9+
- Registered in `Java8toJava11` composite recipe

- Closes https://github.com/moderneinc/customer-requests/issues/2221

## Test plan
- [x] Local variable rename with reference sites
- [x] Method parameter rename with usage in body
- [x] Instance field rename with `this.` access
- [x] Method name rename (declaration + invocation)
- [x] Catch parameter, lambda parameter, for-each loop variable
- [x] No change on Java 9+ source files
- [x] No change for `_foo`, `foo_`, `my_value` identifiers
- [x] No change for `__` (idempotent)


src/test/java/org/openrewrite/java/migrate/lang/RenameUnderscoreIdentifierTest.java
```diff
@@ -0,0 +1,359 @@
...
+            }
+
+            @Override
+            public String getDescription() {
+                return "Test setup recipe.";
+            }
+
+            @Override
+            public TreeVisitor<?, ExecutionContext> getVisitor() {
+                return new JavaIsoVisitor<ExecutionContext>() {
```
timtebeek: Surprised were copying the visitor here. Can we reuse that  (in part) at least from the actual implementation?

timtebeek (approved):

PR #1117, opened 2026-05-28
fix: do not replace lambda params with underscore when used as assignment target

### LINKED ISSUE #1116: ReplaceUnusedVariablesWithUnderscore incorrectly renames lambda params used as assignment targets producing uncompilable code
What version of OpenRewrite are you using?
- rewrite-migrate-java 3.35.0
- rewrite-maven-plugin 6.40.0

How are you running OpenRewrite?

Maven plugin, single module project.

```
<plugin>
  <version>6.40.0</version>
  <configuration>
    <activeRecipes>
      <recipe>org.openrewrite.java.migrate.UpgradeToJava25</recipe>
    </activeRecipes>
  </configuration>
</plugin>
```

What is the smallest, simplest way to reproduce the problem?
```
import java.util.function.BinaryOperator;

class Test {
    void test() {
        BinaryOperator<String> op = (a, b) -> a = "value";
    }
}
```

What did you expect to see?
```
import java.util.function.BinaryOperator;

class Test {
    void test() {
        BinaryOperator<String> op = (a, _) -> a = "value";
    }
}
```

What did you see instead?
```
import java.util.function.BinaryOperator;

class Test {
    void test() {
        BinaryOperator<String> op = (_, _) -> _ = "value";
    }
}
```

This does not compile because _ cannot be used as an assignment target.
Full stack trace: N/A because no error thrown, recipe silently produces uncompilable code.

Are you interested in contributing a fix? Yes, a fix and tests are ready as a PR.

darrencilia:
## Problem
`ReplaceUnusedVariablesWithUnderscore` incorrectly renames lambda parameters to `_` when they appear as LHS assignment targets, producing uncompilable code.

## Fix
Added `usedAsAssignmentTarget()` guard following the same pattern as the existing `usedInModifyingUnary()` method, and wired it into `renameVariableIfUnusedInContext`.

## Tests
Added three test cases covering:
- First param as assignment target, second correctly renamed
- Second param as assignment target, first correctly renamed
- Single param as assignment target

- Fixes #1116

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/lang/ReplaceUnusedVariablesWithUnderscore.java
+++ b/src/main/java/org/openrewrite/java/migrate/lang/ReplaceUnusedVariablesWithUnderscore.java
@@ -102,7 +102,8 @@ public class ReplaceUnusedVariablesWithUnderscore extends Recipe {
             private boolean renameVariableIfUnusedInContext(J.VariableDeclarations.NamedVariable variable, J context) {
                 if (!UNDERSCORE.equals(variable.getName().getSimpleName()) &&
                         VariableReferences.findRhsReferences(context, variable.getName()).isEmpty() &&
-                        !usedInModifyingUnary(variable.getName(), context)) {
+                        !usedInModifyingUnary(variable.getName(), context) &&
+                        !usedAsAssignmentTarget(variable.getName(), context)) {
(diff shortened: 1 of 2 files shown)


timtebeek (approved):
Thanks for the immediate fix! Looks perfect.

PR #1122, opened 2026-06-01
Route WSS4J message.token classes to their real 2.x packages

Jenson3210:
## Summary

- `UpdateApacheWSSecurityPackages`'s recursive `ChangePackage org.apache.ws.security → org.apache.wss4j.common.ext` rewrites `org.apache.ws.security.message.token.*` to the non-existent `org.apache.wss4j.common.ext.message.token.*`. The real WSS4J 2.x homes for those classes live in three different packages.
- Append `ChangeType` entries that route each of the 18 affected classes to its real destination — split across `org.apache.wss4j.common.token` (8 classes), `org.apache.wss4j.dom.message.token` (6 classes), and `org.apache.wss4j.common.kerberos` (4 classes).
- Add `wss4j-ws-security-dom` to the test parser classpath so the new test can resolve the `dom.message.token.*` destinations.

`KerberosClientAction` and `KerberosServiceAction` were removed entirely in WSS4J 2.1 with no same-API successor; intentionally not mapped.

- Tracks moderneinc/customer-requests#2462 (item 9).

## Test plan
- [x] `./gradlew build` — green
- [x] New `UpdateApacheWSSecurityPackagesTest#messageTokenTypesRouteToTheirRealWss4j2xPackages` covers all three destination packages
- [x] Existing `updateApacheWSSecurityPackages` test still passes


src/main/resources/META-INF/rewrite/jakarta-ee-9.yml
```diff
@@ -1391,6 +1391,63 @@ recipeList:
       oldPackageName: org.apache.ws.security.components.crypto
       newPackageName: org.apache.wss4j.common.crypto
       recursive: true
+  # The recursive ChangePackage above misroutes message.token classes to
+  # `org.apache.wss4j.common.ext.message.token.*`, a namespace that does not exist in any WSS4J
```
Jenson3210: If we move these above the misrouting, then we only need 1 correct routing, no?

src/main/resources/META-INF/rewrite/jakarta-ee-9.yml
```diff
@@ -1383,6 +1383,62 @@ description: >-
 preconditions:
   - org.openrewrite.Singleton
 recipeList:
+  # Route message.token classes to their real WSS4J 2.x locations before the recursive
+  # ChangePackage below, which would otherwise misroute them to a namespace that does not exist.
```
Jenson3210:
```suggestion
```

timtebeek (approved):

PR #1134, opened 2026-06-20
use . instead # for FQN link in Markdown javadoc

barbulescu:
## What's changed?
JavadocToMarkdownDocComment now correctly uses . as the package separator in fully-qualified Javadoc links. Previously, J.FieldAccess nodes in printJRef used # for every level of the reference chain, so {@link java.util.List} became [java#util#List] and {@link java.util.List#get(int)} became [java#util#List#get(int)].

The fix is a one-character change in printJRef: when a J.FieldAccess has a non-empty target, join with . rather than #. The # separator between a class and its member is already emitted by the J.MethodInvocation and J.MemberReference branches, so no other change is needed.

## What's your motivation?
The recipe produced broken Markdown links for any {@link} or @see tag that used a fully-qualified class name. The pattern [org#assertj#core#...#ClassName#method()] is invalid Markdown Javadoc syntax; the correct form is [org.assertj.core...ClassName#method()]. This affected at least 30 files in practice.

@timtebeek 

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/lang/JavadocToMarkdownDocComment.java
+++ b/src/main/java/org/openrewrite/java/migrate/lang/JavadocToMarkdownDocComment.java
@@ -548,7 +548,7 @@ public class JavadocToMarkdownDocComment extends Recipe {
                 if (target.isEmpty()) {
                     return "#" + name;
                 }
-                return target + "#" + name;
+                return target + "." + name;
             }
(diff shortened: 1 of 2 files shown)


timtebeek (approved):
Oh wow, thanks so much for flagging and fixing; I need to step up my OSS contributions as well again to discover these kind of issues. 🙏🏻

PR #1140, opened 2026-06-23
Make ExtractExplicitConstructorInvocationArguments templates context-free

knutwannheden:
## Motivation

`ExtractExplicitConstructorInvocationArguments` (JEP 513) applied both of its `JavaTemplate`s with `contextSensitive()`. Context-sensitive templating reconstructs and reparses the enclosing class and method on every application, which is the dominant cost when the recipe actually fires. Neither template needs it: `contextSensitive()` only reconstructs the surrounding *local* scope (leading variable declarations, method parameters) into the parse stub — it does not govern type resolution. Every type these templates declare is the type of an existing `super(..)`/`this(..)` argument, so it is already present in the unit, resolves via the shared `JavaTypeCache`, and the spliced result is re-attributed against the real cursor afterward. Dropping `contextSensitive()` therefore produces byte-identical, fully type-attributed output while skipping the enclosing-context reparse.

## Summary

- Drop `contextSensitive()` from both templates (the local-variable declarations and the argument-list rewrite).
- Bail on the cheapest decisive check — no method-invocation/object-creation argument — *before* any variable-name generation, cursor-scope scanning, or templating, so a constructor the recipe will not touch pays nothing.
- Add two regression tests that lock in the type-attribution behavior the context-free path must preserve: one where the extracted variable's type is a source-path-only type defined in a sibling source, and one where the extracted variable is declared with a *widening* source-path supertype that is referenced nowhere else in the unit. They assert the rewritten reference keeps its variable binding (`fieldType`) and the extracted declaration keeps a resolved fully-qualified type (not `Unknown`).

## Test plan

- [x] `ExtractExplicitConstructorInvocationArgumentsTest` passes (all existing cases plus the two new source-path regression tests).
- [x] New tests assert post-recipe type attribution via `afterRecipe`: the `super(..)` reference has a non-null type and `fieldType`, and the extracted declaration's type resolves to a real FQ type.

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/lang/ExtractExplicitConstructorInvocationArguments.java
+++ b/src/main/java/org/openrewrite/java/migrate/lang/ExtractExplicitConstructorInvocationArguments.java
@@ -79,14 +79,26 @@ public class ExtractExplicitConstructorInvocationArguments extends Recipe {
                     return md;
                 }
 
+                // Only act when at least one argument actually does work worth surfacing. This is the
+                // cheapest decisive check, so make it first and bail before any name generation, scope
+                // scanning, or templating that an unaffected constructor would not need.
+                boolean anyComplex = false;
+                for (Expression arg : args) {
+                    if (arg instanceof J.MethodInvocation || arg instanceof J.NewClass) {
+                        anyComplex = true;
+                        break;
+                    }
+                }
+                if (!anyComplex) {
+                    return md;
+                }
+
                 // Plan the extraction: hoist everything but trivial, side-effect-free arguments, in order.
-                // Only act when at least one extracted argument actually does work worth surfacing.
                 Set<String> usedNames = new LinkedHashSet<>();
                 String[] names = new String[args.size()];
                 Set<String> imports = new LinkedHashSet<>();
                 StringBuilder declarations = new StringBuilder();
                 List<Expression> declarationArgs = new ArrayList<>();
-                boolean anyComplex = false;
                 for (int i = 0; i < args.size(); i++) {
                     Expression arg = args.get(i);
                     if (isInlineSafe(arg)) {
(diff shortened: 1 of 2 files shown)


timtebeek (approved):

src/main/java/org/openrewrite/java/migrate/lang/ExtractExplicitConstructorInvocationArguments.java
```diff
@@ -79,14 +79,26 @@ public J.MethodDeclaration visitMethodDeclaration(J.MethodDeclaration method, Ex
...
                 }
 
+                // Only act when at least one argument actually does work worth surfacing. This is the
+                // cheapest decisive check, so make it first and bail before any name generation, scope
+                // scanning, or templating that an unaffected constructor would not need.
+                boolean anyComplex = false;
+                for (Expression arg : args) {
+                    if (arg instanceof J.MethodInvocation || arg instanceof J.NewClass) {
+                        anyComplex = true;
+                        break;
```
timtebeek: I don't think we'd need the above variable, nor the below `if`, when we return if deemed complex, right?
```suggestion
                        return md;
```

src/main/java/org/openrewrite/java/migrate/lang/ExtractExplicitConstructorInvocationArguments.java
```diff
@@ -146,7 +154,7 @@ public J.ClassDeclaration visitClassDeclaration(J.ClassDeclaration classDecl, Ex
                     public J.MethodInvocation visitMethodInvocation(J.MethodInvocation mi, ExecutionContext ctx2) {
                         mi = super.visitMethodInvocation(mi, ctx2);
                         if (isExplicitConstructorInvocation(mi)) {
-                            return JavaTemplate.builder(argumentList.toString()).contextSensitive().build()
+                            return JavaTemplate.builder(argumentList.toString()).build()
                                     .apply(getCursor(), mi.getCoordinates().replaceArguments(), inlineArgs.toArray());
```
timtebeek:
```suggestion
                            return JavaTemplate.apply(argumentList.toString(), getCursor(), mi.getCoordinates().replaceArguments(), inlineArgs.toArray());
```

src/main/java/org/openrewrite/java/migrate/lang/ExtractExplicitConstructorInvocationArguments.java
knutwannheden: Good call on dropping the flag + extra `if`. I went with a single `noneMatch` guard rather than the literal suggestion: the recipe *acts* when an argument is complex and bails when none are, so returning `md` on the first complex arg would invert the guard and make it a no-op (the new regression tests catch this). Pushed:

```java
if (args.stream().noneMatch(arg -> arg instanceof J.MethodInvocation || arg instanceof J.NewClass)) {
    return md;
}
```

PR #1147, opened 2026-06-25
Don't rewrite non-JDK `Optional` in OptionalNot{Present,Empty} recipes

### LINKED ISSUE #1146: OptionalNotPresentToIsEmpty rewrites `!isPresent()` on a non-`java.util.Optional` type, producing uncompilable `isEmpty()`

- Gradle plugin `org.openrewrite:plugin` v7.35.0
- rewrite-migrate-java v3.38.0
- Active recipe: `org.openrewrite.java.migrate.UpgradeToJava21` (which pulls in `org.openrewrite.java.migrate.util.OptionalNotPresentToIsEmpty`)

Gradle plugin, applied via an init script over a large multi-module project:

```groovy
rootProject {
    plugins.apply(org.openrewrite.gradle.RewritePlugin)
    dependencies {
        rewrite("org.openrewrite.recipe:rewrite-migrate-java:3.38.0")
    }
    rewrite {
        activeRecipe("org.openrewrite.java.migrate.UpgradeToJava21")
        setExportDatatables(true)
    }
}
```

The affected module depends on Jolt (`com.bazaarvoice.jolt:jolt-core:0.1.8`), which ships its own
`com.bazaarvoice.jolt.common.Optional<T>`. That class exposes `isPresent()` and `get()` but **has no
`isEmpty()` method** (its full public API is `empty()`, `of(T)`, `get()`, `isPresent()`, `equals`, `toString`).

The project is private, but the third-party type is public and small; the relevant snippet is below.

The receiver is a Jolt `Optional`, not a JDK `Optional`:

```java
import com.bazaarvoice.jolt.common.Optional; // NOT java.util.Optional

class ConvertMarkdownToHtmlRTL {
    Optional<Object> applySingle(Object input) {
        Optional<Object> baseHtmlOpt = convert(input);
        if (!baseHtmlOpt.isPresent()) {   // original, compiles fine
            return Optional.empty();
        }
        // ...
    }
}
```

The code left unchanged. `baseHtmlOpt` is a `com.bazaarvoice.jolt.common.Optional`, not a
`java.util.Optional`, so the recipe should not apply. The recipe's `MethodMatcher` is in fact scoped to the
JDK type — `new MethodMatcher("java.util.Optional isPresent()")` with template
`#{any(java.util.Optional)}.isEmpty()`
(`OptionalNotPresentToIsEmpty.java`, lines 46 and 56 on `v3.38.0`) — so the intent is clearly JDK-only.

## What did you see instead?

The recipe rewrote it to call `isEmpty()` on the Jolt type, which does not have that method, so the module
no longer compiles:

```java
import com.bazaarvoice.jolt.common.Optional;

class ConvertMarkdownToHtmlRTL {
    Optional<Object> applySingle(Object input) {
        Optional<Object> baseHtmlOpt = convert(input);
        if (baseHtmlOpt.isEmpty()) {   // does not compile: no isEmpty() on com.bazaarvoice.jolt.common.Optional
            return Optional.empty();
        }
        // ...
    }
}
```

Because the JDK-scoped `MethodMatcher` nevertheless matched a `com.bazaarvoice.jolt.common.Optional`
receiver, this looks like the matcher firing when the receiver's type is unresolved/incompletely attributed
(the third-party `Optional` not being on the recipe's typed classpath), falling through to a name-only match
on `isPresent()`. A type-scoped migration recipe should not rewrite when it cannot positively confirm the
receiver is `java.util.Optional` — otherwise it silently emits uncompilable code with no warning or error from
the run itself.

- Note this is a distinct failure mode from #417/#418 (a `ClassCastException` thrown by the sibling
`OptionalNotEmptyToIsPresent`); here there is no crash — the recipe completes "successfully" and the breakage
only surfaces at the next `compileJava`.

No stack trace — the recipe run completes without error. The failure is a downstream Java compile error:

```
error: cannot find symbol
        if (baseHtmlOpt.isEmpty()) {
                       ^
  symbol:   method isEmpty()
  location: variable baseHtmlOpt of type com.bazaarvoice.jolt.common.Optional<Object>
```

## Are you interested in contributing a fix to OpenRewrite?

Open to it if you can point me at the preferred guard (e.g. requiring resolved receiver type before applying). Happy to add a recipe test that reproduces the mis-fire on a non-JDK `Optional`.

timtebeek:
## What's changed?

`OptionalNotPresentToIsEmpty` and `OptionalNotEmptyToIsPresent` now skip a source file when the simple name `Optional` is bound to an explicitly imported type other than `java.util.Optional`.

## Why?

- Fixes #1146.

When a third-party `Optional` (e.g. `com.bazaarvoice.jolt.common.Optional`, which has `isPresent()` but no `isEmpty()`) is off the recipe's typed classpath, javac can misattribute `bar.isPresent()` to `java.util.Optional` — for instance when a `java.util.*` import is in scope. The `MethodMatcher` (`java.util.Optional isPresent()`) then matches and the recipe rewrites the call to `isEmpty()`, which does not exist on the third-party type, so the module no longer compiles.

Because the type information is uniformly (mis)attributed to `java.util.Optional` in this scenario, a type-based guard on the receiver can't distinguish the cases. The reliable signal is the source itself: if the file explicitly imports a non-JDK type named `Optional`, the bare `Optional` name cannot be `java.util.Optional`, so the recipe conservatively leaves the file untouched.

The sibling `OptionalNotEmptyToIsPresent` has the identical structure and the same failure mode, so it gets the same guard.

## Tests

Added `doNotChangeNonJdkOptionalMisattributedToJdk` to both recipe test classes. Both reproduce the misfire (they fail on `main` and pass with this change). A test covering a fully-resolved third-party `Optional` is also added.

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/util/OptionalNotEmptyToIsPresent.java
+++ b/src/main/java/org/openrewrite/java/migrate/util/OptionalNotEmptyToIsPresent.java
@@ -49,6 +49,21 @@ public class OptionalNotEmptyToIsPresent extends Recipe {
                 new UsesMethod<>(optionalIsPresentMatcher));
         return Preconditions.check(check, new JavaVisitor<ExecutionContext>() {
             @Override
+            public J visitCompilationUnit(J.CompilationUnit cu, ExecutionContext ctx) {
+                // A type whose method invocations are misattributed to `java.util.Optional` (e.g. when a
+                // third-party `Optional` is off the recipe classpath) can fool the `MethodMatcher`, producing
+                // an uncompilable `isPresent()` call. Skip the file when the simple name `Optional` is bound to
+                // some other explicitly imported type.
+                for (J.Import anImport : cu.getImports()) {
+                    if (!anImport.isStatic() && "Optional".equals(anImport.getClassName()) &&
+                            !"java.util.Optional".equals(anImport.getTypeName())) {
+                        return cu;
+                    }
+                }
+                return super.visitCompilationUnit(cu, ctx);
+            }
+
+            @Override
             public J visitStatement(Statement s, ExecutionContext ctx) {
                 if (s instanceof J.Unary) {
                     J.Unary unary = (J.Unary) s;
--- a/src/main/java/org/openrewrite/java/migrate/util/OptionalNotPresentToIsEmpty.java
+++ b/src/main/java/org/openrewrite/java/migrate/util/OptionalNotPresentToIsEmpty.java
@@ -49,6 +49,21 @@ public class OptionalNotPresentToIsEmpty extends Recipe {
                 new UsesMethod<>(optionalIsPresentMatcher));
         return Preconditions.check(check, new JavaVisitor<ExecutionContext>() {
             @Override
+            public J visitCompilationUnit(J.CompilationUnit cu, ExecutionContext ctx) {
+                // A type whose method invocations are misattributed to `java.util.Optional` (e.g. when a
+                // third-party `Optional` is off the recipe classpath) can fool the `MethodMatcher`, producing
+                // an uncompilable `isEmpty()` call. Skip the file when the simple name `Optional` is bound to
+                // some other explicitly imported type.
+                for (J.Import anImport : cu.getImports()) {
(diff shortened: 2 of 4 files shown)


timtebeek:
This should be fixed in the type information upstream.

PR #1153, opened 2026-06-26
Add test case to ensure jakarta.xml.ws-api is not duplicated 

evie-lau:
- Followup to #1151, which duplicates `jakarta.xml.ws-api`.

If starting dependencies (pre-EE10) contained both `jakarta.jws-api` and `jakarta.xml.ws-api`, then the `ReplaceJakartaJwsWithJakartaXmlWs` would migrate and end up with two instances of `jakarta.xml.ws-api`


timtebeek:
Hi thanks! Instead of compensating for a duplication here, I propose we look to avoid the problem upstream if we can
- https://github.com/openrewrite/rewrite/pull/8129

evie-lau:
That sounds good, I can remove the recipe change here and keep the test case

PR #1155, opened 2026-06-26
Upgrade maven-ejb-plugin alongside jakarta.ejb-api

### LINKED ISSUE #1154: Upgrade `maven-ejb-plugin` when upgrading `jakarta.ejb-api` in EE9
In EE9 migration, when `javax.ejb-api` gets upgraded to `jakarta.ejb-api:4.0.x`, a corresponding `maven-ejb-plugin` plugin that uses it needs to be upgraded as well.

`maven-ejb-plugin` requires `3.2.1` or higher for EE9 (current release `3.3.0`)

Additionally, it's `<ejbVersion>` configuration needs to be updated to 4.0.

```xml
<plugin>
	<groupId>org.apache.maven.plugins</groupId>
	<artifactId>maven-ejb-plugin</artifactId>
	<version>3.2.1</version>
	<configuration>
		<ejbVersion>4.0</ejbVersion>
	</configuration>
</plugin>
```

evie-lau:
- Fixes #1154

Also handles decoupling old `javax.ejb-api` version from `maven-ejb-plugin`'s `<ejbVersion>` configuration


src/main/resources/META-INF/rewrite/jakarta-ee-9.yml
```diff
@@ -364,6 +364,16 @@ recipeList:
       version: 4.0.x
       onlyIfUsing: javax.ejb..*
       acceptTransitive: true
+  - org.openrewrite.maven.UpgradePluginVersion:
+      groupId: org.apache.maven.plugins
+      artifactId: maven-ejb-plugin
+      newVersion: 3.2.x
```
evie-lau, line 370: I think 3.3.0 should work as well, but 3.2.1 is the minimum required.

Thoughts?

src/main/resources/META-INF/rewrite/jakarta-ee-9.yml
```diff
@@ -364,6 +364,16 @@ recipeList:
...
+  # After https://github.com/openrewrite/rewrite/issues/3810 is resolved, can use ChangePluginConfiguration to target specific config without replacing entire block
+  - org.openrewrite.xml.ChangeTagValue:
+      elementName: /project/build/plugins/plugin/configuration/ejbVersion
```
timtebeek: Would it help to use something like the following to pick up plugins configured in profiles or pluginManagement? This is from memory, so don't accept blindly. 😅 
```suggestion
      elementName: //plugins/plugin/configuration/ejbVersion
```

src/main/resources/META-INF/rewrite/jakarta-ee-9.yml
timtebeek: Fine to use this; should resolve to whatever version is available in 3.2 range; if you're confident 3.x would work for the foreseeable future we could use that as well to cover both, although it could lead to some unnecessary changes.

src/main/resources/META-INF/rewrite/jakarta-ee-9.yml
evie-lau: Good catch, I was worried about the specific XPath

evie-lau:
Added a case where some versions need decoupling. This may have been more common in older apps using `javax.ejb-api`. There's a version `3.2` that could be used as a shared variable for both `javax.ejb-api` version and `maven-ejb-plugin`'s `<ejbVersion>` configuration. 

After the jakarta version releases, the version syntax no longer matches. 
ie.
- `4.0.1` in `jakarta.ejb-api`
- `4.0` in `<ejbVersion>` in `maven-ejb-plugin`

Issue #1148, opened 2026-06-25
UseMapOf rewrites the initializer of a HashMap-typed variable to Map.of(), leaving uncompilable code


protocol7:
## What version of OpenRewrite are you using?

* org.openrewrite:rewrite-core:8.85.4
* org.openrewrite.recipe:rewrite-migrate-java:3.37.0

## What is the smallest, simplest way to reproduce the problem?

`UseMapOf` rewrites the initializer of a variable to `Map.of(..)` but does not touch the variable's declared type. When the variable is declared with the concrete type `java.util.HashMap` (rather than the `Map` interface), the result no longer compiles: `Map.of(..)` returns an immutable `java.util.Map`, which is not assignable to a `HashMap` variable.

This only reproduces for a field — as a local variable the value tends to be inlined, and the declared type disappears.

```java
@Test
void useMapOfLeavesIncompatibleHashMapDeclaredType() {
  rewriteRun(
      spec -> spec.recipe(new UseMapOf()),
      // language=java
      java(
          """
          package com.helloworld;

          import java.util.HashMap;

          class Main {
            private static final HashMap<String, String> VALUES =
                new HashMap<String, String>() {
                  {
                    put("key", "value");
                  }
                };
          }
          """,
          spec -> spec.markers(javaVersion(25))));
}
```

## What did you see instead?

The recipe rewrites only the initializer and leaves the declared type as `HashMap`:

```diff
 import java.util.HashMap;
+import java.util.Map;

 class Main {
-  private static final HashMap<String, String> VALUES =
-      new HashMap<String, String>() {
-        {
-          put("key", "value");
-        }
-      };
+  private static final HashMap<String, String> VALUES = Map.of("key", "value");
 }
```

The rewritten source does not compile:

```
incompatible types: java.util.Map<java.lang.String,java.lang.String>
  cannot be converted to java.util.HashMap<java.lang.String,java.lang.String>
```

We hit this on a real test class during a batch run of the conventions recipes, and it had to be repaired by hand.

## What did you expect to see?

Either:

1. The recipe also widens the declared type to `Map<String, String>`, producing compilable code:

   ```java
   private static final Map<String, String> VALUES = Map.of("key", "value");
   ```

   ...or

2. The recipe skips the rewrite when the declared type is a concrete subtype (`HashMap`, `LinkedHashMap`, ...) that would not accept the `Map.of(..)` result.

timtebeek:
[...]

PR #1161, opened 2026-07-01
Skip UseEnumSetOf empty conversion for static fields to avoid circular class-init (#1157)

### LINKED ISSUE #1157: UseEnumSetOf introduces circular class-init ClassCastException
Here's a fun edge case, feel free to ignore :)

- org.openrewrite.recipe:rewrite-migrate-java:3.38.0
- org.openrewrite:rewrite-core:8.85.7
- org.openrewrite:rewrite-java:8.85.7

`UseEnumSetOf` rewrites `Set.of()` to `EnumSet.noneOf(E.class)` without checking whether `E`'s own constants have a circular class-initialization dependency back on the class that declares the field being rewritten. When that's the case, the resulting code compiles fine but can throw `ClassCastException` at runtime depending on which class happens to get touched first.

```java
@Test
void useEnumSetOfIntroducesCircularStaticInitClassCastException() {
  rewriteRun(
      spec -> spec.recipe(new UseEnumSetOf(true)),
      // language=java
      java(
          """
          package com.helloworld;

          import java.util.Set;

          public enum ConfigType {
            ENUM,
            BOOLEAN,
            INTEGER;

            public static final Set<ConfigProperty> SUBSCRIBE_TO_ALL = Set.<ConfigProperty>of();
          }
          """,
          spec -> spec.markers(javaVersion(25))),
      // language=java
      java(
          """
          package com.helloworld;

          import static com.helloworld.ConfigType.BOOLEAN;
          import static com.helloworld.ConfigType.ENUM;
          import static com.helloworld.ConfigType.INTEGER;

          public enum ConfigProperty {
            TRIAL_TYPE(ENUM),
            IS_TRIAL_ENABLED(BOOLEAN),
            MIN_ACCOUNT_AGE_MINUTES(INTEGER);

            ConfigProperty(final ConfigType type) {}
          }
          """,
          spec -> spec.markers(javaVersion(25))));
}
```

`ConfigProperty`'s constants are constructed with arguments (`ENUM`, `BOOLEAN`, `INTEGER`) that are static fields of `ConfigType`, and `ConfigType` separately declares a field of type `Set<ConfigProperty>`.

## What did you see instead?

The recipe rewrites `ConfigType`'s field:

```diff
 package com.helloworld;

+import java.util.EnumSet;
 import java.util.Set;

 public enum ConfigType {
   ENUM,
   BOOLEAN,
   INTEGER;

-  public static final Set<ConfigProperty> SUBSCRIBE_TO_ALL = Set.<ConfigProperty>of();
+  public static final Set<ConfigProperty> SUBSCRIBE_TO_ALL = EnumSet.noneOf(ConfigProperty.class);
 }
```

This compiles cleanly. But if anything touches `ConfigProperty` before `ConfigType` (e.g. a caller that only needs `ConfigProperty.TRIAL_TYPE`, or class-loading order in a larger app), it throws at runtime (reproduced by hand-compiling the rewritten output with `javac`/`java` 17.0.19, outside the recipe test harness):

```
Exception in thread "main" java.lang.ExceptionInInitializerError
	at com.helloworld.ConfigProperty.<clinit>(ConfigProperty.java:8)
	at com.helloworld.Main.main(Main.java:5)
Caused by: java.lang.ClassCastException: class com.helloworld.ConfigProperty not an enum
	at java.base/java.util.EnumSet.noneOf(EnumSet.java:113)
	at com.helloworld.ConfigType.<clinit>(ConfigType.java:12)
	... 2 more
```

`ConfigProperty` genuinely is an enum — the message is misleading. What happens: `ConfigProperty.<clinit>` starts constructing `TRIAL_TYPE`, which needs `ConfigType.ENUM`, which triggers `ConfigType.<clinit>` (nested, same thread). `ConfigType.<clinit>` finishes its own constants and reaches `SUBSCRIBE_TO_ALL = EnumSet.noneOf(ConfigProperty.class)`, which reflectively invokes `ConfigProperty.values()` via `Class.getEnumConstantsShared()`. But `ConfigProperty`'s own `$VALUES` array hasn't been assigned yet — its `<clinit>` is still stuck constructing the very first constant, several frames up the stack. `values()` throws `NullPointerException`; `getEnumConstantsShared()` swallows `InvocationTargetException`/etc. and returns `null`; `EnumSet.noneOf` sees `null` and throws `ClassCastException("... not an enum")` instead of propagating the real cause.

We hit this for real in an internal enum pair analogous to the above (one enum's constants carrying a "type" from a sibling enum that also holds an `EnumSet.noneOf(...)` field) during a batch run of migration recipes, and had to revert the change by hand.

Either:

1. `UseEnumSetOf` detects that `E`'s own class-initialization can re-enter the class declaring the field (directly, or via a static import used in one of `E`'s constant constructor arguments) and skips the rewrite, leaving `Set.of()`/`Collections.emptySet()` in place, or
2. At minimum, the recipe should not apply when `E`'s constants reference static state on another enum that isn't yet guaranteed to be initialized — this is inherently unsafe for any pair of enums with a two-way static dependency, regardless of which one happens to load first in a given run.

timtebeek:
- Fixes #1157

`UseEnumSetOf` rewrites an empty `Set.of()` to `EnumSet.noneOf(E.class)`. In a **static field initializer**, `EnumSet.noneOf` reflectively reads `E`'s constants during class initialization. When two enums have a circular static dependency (one enum's constants carry a value from a sibling enum that also holds an `EnumSet.noneOf(...)` static field), this can re-enter a not-yet-initialized enum and throw a misleading `ClassCastException("... not an enum")` at runtime — even though the code compiles cleanly.

## Fix

Skip the empty `Set.of()` → `EnumSet.noneOf(E.class)` conversion when the assignment target is a **static field**. Static field initializers run during `<clinit>`, which is the only place the re-entrant class-init hazard exists; local variables and instance fields are unaffected and still convert (e.g. the existing `dontHaveArgs` test). The non-empty `Set.of(E.X, ...)` case is untouched since it does not use the reflective `noneOf` path.

- This mirrors the approach of #1148 (skip when the assignment target is a concrete `HashMap`).

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/util/UseEnumSetOf.java
+++ b/src/main/java/org/openrewrite/java/migrate/util/UseEnumSetOf.java
@@ -78,6 +78,14 @@ public class UseEnumSetOf extends Recipe {
                             }
 
                             if (args.get(0) instanceof J.Empty) {
+                                // Skip the empty `Set.of()` -> `EnumSet.noneOf(E.class)` conversion for static
+                                // field initializers: `EnumSet.noneOf` reflectively reads the enum's constants
+                                // during class initialization, which can re-enter a not-yet-initialized enum and
+                                // throw a misleading `ClassCastException` when two enums have a circular static
+                                // dependency (issue #1157).
+                                if (isStaticField(parent)) {
+                                    return mi;
+                                }
                                 JavaType firstTypeParameter = ((JavaType.Parameterized) type).getTypeParameters().get(0);
                                 JavaType.ShallowClass shallowClass = JavaType.ShallowClass.build(firstTypeParameter.toString());
                                 return JavaTemplate.builder("EnumSet.noneOf(" + shallowClass.getClassName() + ".class)")
@@ -118,6 +126,11 @@ public class UseEnumSetOf extends Recipe {
                     }
                 }
                 return false;
+            }
+
+            private boolean isStaticField(Cursor parent) {
+                return parent.getValue() instanceof J.VariableDeclarations &&
+                        ((J.VariableDeclarations) parent.getValue()).hasModifier(J.Modifier.Type.Static);
             }
 
             private boolean isArrayParameter(final List<Expression> args) {
--- a/src/main/java/org/openrewrite/java/migrate/util/UseEnumSetOf.java
+++ b/src/main/java/org/openrewrite/java/migrate/util/UseEnumSetOf.java
@@ -78,11 +78,6 @@ public class UseEnumSetOf extends Recipe {
                             }
 
                             if (args.get(0) instanceof J.Empty) {
-                                // Skip the empty `Set.of()` -> `EnumSet.noneOf(E.class)` conversion for static
-                                // field initializers: `EnumSet.noneOf` reflectively reads the enum's constants
-                                // during class initialization, which can re-enter a not-yet-initialized enum and
-                                // throw a misleading `ClassCastException` when two enums have a circular static
-                                // dependency (issue #1157).
                                 if (isStaticField(parent)) {
                                     return mi;
                                 }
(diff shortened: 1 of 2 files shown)


protocol7:
Quick as usual, thanks!

PR #1162, opened 2026-07-03
feat(migrate): support return statements in IfElseIfConstructToSwitch recipe

### LINKED ISSUE #776: Enhance IfElseIfConstructToSwitch to support "return" statements
Currently the `IfElseIfConstructToSwitch` recipe does not convert to a switch if any of the parts would return. 
However, if all branches would return, we could convert it to and it could be a potential candidate. Only the intermediate phase of having `return`s in the case is not valid, hence would not be picked up by: 
```java
        if (obj instanceof String s) {
            return s;
            return String.valueOf(l);
        } else {
            throw new IllegalStateException("Unexpected value: " + obj);
        }
    }
```

could become

```java
        return switch (obj) {
            case String s -> s;
            case Long l -> String.valueOf(l);
            default -> throw new IllegalStateException("Unexpected value: " + obj);
        };
    }
```

arnabnandy7:
## What's changed?
Allows converting if-else-if chains containing return statements to switch statements.
* Removed the blanket `returns(...)` guard in `validatePotentialCandidate()` within `IfElseIfConstructToSwitch.java`.
* Updated `buildSwitchTemplate()` to check if the branch statement is a bare `J.Return`. If so, it wraps it in block braces at the template-compilation level (`case X -> {#{any()};}`).
* Updated the existing return-related tests in `IfElseIfConstructToSwitchTest.java` to expect successful switch conversions, and added new tests to cover mixed branches, braceless return chains, and nested returns.

## What's your motivation?
- Closes #776

* The template compilation logic using `{#{any()};}` to ensure the mock code parsed in `JavaTemplate` is valid while avoiding double semicolons during replacement.
* The test cases covering nested returns (e.g. returns inside lambda loops) to ensure they are preserved correctly.

An alternative would have been to manually synthesize `J.Block` AST nodes in the visitor before applying `JavaTemplate`. However, wrapping the raw return in explicit braces at the template-string level is much cleaner and less error-prone.

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/lang/IfElseIfConstructToSwitch.java
+++ b/src/main/java/org/openrewrite/java/migrate/lang/IfElseIfConstructToSwitch.java
@@ -37,7 +37,6 @@ import org.openrewrite.staticanalysis.groovy.GroovyFileChecker;
 import org.openrewrite.staticanalysis.kotlin.KotlinFileChecker;
 
 import java.util.*;
-import java.util.concurrent.atomic.AtomicBoolean;
 
 import static java.util.Collections.emptyList;
 import static java.util.Collections.singletonList;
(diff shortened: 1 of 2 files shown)


timtebeek:
I'm finding it hard to fit in a review here @arnabnandy7 ; I've unassigned myself and hope colleagues can pick this up. I'll be out for about a month. :/

PR #1165, opened 2026-07-06
Skip `UseMapOf` for `LinkedHashMap`/`TreeMap` in the prose `put()` form (#1163)

### LINKED ISSUE #1163: UseMapOf still converts `LinkedHashMap` to `Map.of` for the statement/`put()` form
**What version of OpenRewrite are you using?**

- rewrite-migrate-java: 3.39.0
- rewrite-core: 8.86.1
- rewrite-java: 8.86.1

**What is the smallest, simplest way to reproduce the problem?**

```java
import org.openrewrite.java.migrate.util.UseMapOf;
import org.openrewrite.test.RewriteTest;

import static org.openrewrite.java.Assertions.java;

class UseMapOfTest implements RewriteTest {

    @Test
    void doesNotConvertLinkedHashMapBuiltWithPutStatements() {
        rewriteRun(
            spec -> spec.recipe(new UseMapOf()),
            //language=java
            java(
                """
                import java.util.LinkedHashMap;
                import java.util.Map;

                class Main {
                    static Map<String, String> ordered() {
                        Map<String, String> m = new LinkedHashMap<>();
                        m.put("a", "1");
                        m.put("b", "2");
                        m.put("c", "3");
                        return m;
                    }
                }
                """
            )
        );
    }
}
```

**What did you expect to see?**

- No change — for the same reason as #1112. `LinkedHashMap` guarantees insertion-order iteration; `Map.of(...)` makes no ordering guarantee (its iteration order is unspecified and deliberately randomized per JVM instance). Converting `LinkedHashMap` (or `TreeMap`) to `Map.of` silently drops the ordering contract the original code relied on.

- #1113 already made `UseMapOf` skip `LinkedHashMap`/`TreeMap` — but only for the anonymous-class (double-brace) initializer form from #1112. The equivalent map built with separate `put(...)` statements should be skipped for the identical reason and is not.

**What did you see instead?**

The statement form is still rewritten, discarding the insertion-order guarantee (and swapping the `LinkedHashMap` import for `HashMap`):

```java
import java.util.HashMap;
import java.util.Map;

class Main {
  static Map<String, String> ordered() {
    Map<String, String> m =
        new HashMap<>(
            Map.of(
                "a", "1",
                "b", "2",
                "c", "3"));
    return m;
  }
}
```

timtebeek:
- Fixes #1163

- `UseMapOf` recognizes two input shapes: the anonymous-class (double-brace) initializer and a `new HashMap<>()` declaration followed by a chain of `put(..)` statements. #1113 taught the anonymous-class path to skip `HashMap` subclasses like `LinkedHashMap`/`TreeMap` — since `Map.of(..)` makes no iteration-order guarantee, converting an order-sensitive map silently drops its ordering contract.

The prose/statement path missed the same guard. `NEW_HASH_MAP` is built with `matchOverrides=true`, so `NEW_HASH_MAP.matches(new LinkedHashMap<>())` is `true`, and the chain was still collapsed into `new HashMap<>(Map.of(..))` — dropping insertion order and swapping the `LinkedHashMap` import for `HashMap`.

`matchingTargetName` now applies the same exact-type check used by the anonymous-class path (`TypeUtils.isOfClassType(..., "java.util.HashMap")`), so only an exact `HashMap` declaration is absorbed.

Added a regression test mirroring the issue's reproduction.

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/util/UseMapOf.java
+++ b/src/main/java/org/openrewrite/java/migrate/util/UseMapOf.java
@@ -302,6 +302,12 @@ public class UseMapOf extends Recipe {
                         if (!NEW_HASH_MAP.matches(nc)) {
                             return null;
                         }
+                        // Skip `HashMap` subclasses like `LinkedHashMap`/`TreeMap`: `Map.of(..)` makes no
+                        // iteration-order guarantee, so absorbing their `put(..)` chain would silently drop
+                        // the ordering contract the original code relied on (issue #1163, matching #1113).
+                        if (!TypeUtils.isOfClassType(nc.getClazz() != null ? nc.getClazz().getType() : null, "java.util.HashMap")) {
+                            return null;
+                        }
                         if (nc.getBody() != null) {
                             return null;
                         }
--- a/src/test/java/org/openrewrite/java/migrate/util/UseMapOfTest.java
+++ b/src/test/java/org/openrewrite/java/migrate/util/UseMapOfTest.java
@@ -290,6 +290,30 @@ class UseMapOfTest implements RewriteTest {
         );
     }
 
+    @Issue("https://github.com/openrewrite/rewrite-migrate-java/issues/1163")
+    @Test
+    void doNotChangeLinkedHashMapBuiltWithPutStatements() {
+        //language=java
+        rewriteRun(
+          java(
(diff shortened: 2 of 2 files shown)


protocol7:
Thanks!

PR #1167, opened 2026-07-07
Add RedundantUtf8Charset recipe to drop redundant UTF-8 from Files calls

timtebeek:
The character-based `java.nio.file.Files` methods always default to UTF-8, so passing `StandardCharsets.UTF_8` explicitly is redundant; this recipe removes it and cleans up the now-unused import. The `MethodMatcher`s are scoped to the exact `Charset` overloads (e.g. `readString(Path, Charset)`) so non-charset overloads are never touched, and `isUtf8` still guards against non-UTF-8 charsets like `ISO_8859_1`. It is wired into `java-version-11.yml`, the lowest version where every targeted no-charset overload compiles (`readString`/`writeString` are `@since 11`; the other five are `@since 8`). Display name and description use the `@Getter final String` convention, and `recipes.csv` is regenerated accordingly.


src/main/java/org/openrewrite/java/migrate/nio/file/RedundantUtf8Charset.java
```diff
@@ -0,0 +1,111 @@
+/*
+ * Copyright 2025 the original author or authors.
```
steve-aom-elliott: 2026

src/main/java/org/openrewrite/java/migrate/nio/file/RedundantUtf8Charset.java
```diff
@@ -0,0 +1,111 @@
...
+                "java.nio.file.Files write(java.nio.file.Path, java.lang.Iterable, java.nio.charset.Charset, ..)",
+                "java.nio.file.Files lines(java.nio.file.Path, java.nio.charset.Charset)",
+                "java.nio.file.Files newBufferedReader(java.nio.file.Path, java.nio.charset.Charset)",
+                "java.nio.file.Files newBufferedWriter(java.nio.file.Path, java.nio.charset.Charset, ..)"}) {
+            MATCHERS.add(new MethodMatcher(signature));
+        }
+    }
+
+    @Getter
+    final String displayName = "Remove redundant `StandardCharsets.UTF_8` from `java.nio.file.Files` calls";
```
steve-aom-elliott: Might suggest to rephrase slightly to end with `method calls` rather than just `calls`.

src/main/java/org/openrewrite/java/migrate/nio/file/RedundantUtf8Charset.java
```diff
@@ -0,0 +1,111 @@
...
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return Preconditions.check(new UsesType<>("java.nio.charset.StandardCharsets", false), new JavaIsoVisitor<ExecutionContext>() {
+            @Override
+            public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, ExecutionContext ctx) {
+                J.MethodInvocation mi = super.visitMethodInvocation(method, ctx);
+                if (MATCHERS.stream().noneMatch(m -> m.matches(mi))) {
+                    return mi;
+                }
+                List<Expression> arguments = mi.getArguments();
+                for (int i = 0; i < arguments.size(); i++) {
```
steve-aom-elliott, line 68: I suppose to be fair, you _could probably_ start at index `1` for the arguments given everyone of the method signatures has `java.nio.file.Path` as the first argument, but likely an over-optimization

src/main/java/org/openrewrite/java/migrate/nio/file/RedundantUtf8Charset.java
```diff
@@ -0,0 +1,111 @@
...
+                        maybeRemoveImport("java.nio.charset.StandardCharsets");
+
+                        JavaType.Method methodType = mi.getMethodType();
+                        if (methodType != null && i < methodType.getParameterTypes().size()) {
+                            List<JavaType> parameterTypes = new ArrayList<>(methodType.getParameterTypes());
+                            List<String> parameterNames = new ArrayList<>(methodType.getParameterNames());
+                            parameterTypes.remove(i);
+                            if (i < parameterNames.size()) {
+                                parameterNames.remove(i);
+                            }
```
steve-aom-elliott: Shouldn't `parameterTypes` and `parameterNames` be the same length as each other? Just a bit confused by the need for the guard.

src/test/java/org/openrewrite/java/migrate/nio/file/RedundantUtf8CharsetTest.java
```diff
@@ -0,0 +1,264 @@
+/*
+ * Copyright 2025 the original author or authors.
```
steve-aom-elliott: 2026

src/main/java/org/openrewrite/java/migrate/nio/file/RedundantUtf8Charset.java
timtebeek: Fixed, bumped to 2026.

src/main/java/org/openrewrite/java/migrate/nio/file/RedundantUtf8Charset.java
timtebeek: Done, display name now ends with `method calls` (recipes.csv regenerated).

src/main/java/org/openrewrite/java/migrate/nio/file/RedundantUtf8Charset.java
timtebeek: Left the loop starting at 0 as you suspected — keeping it general so it doesn't rely on Path always being arg 0, and the cost is negligible.

src/main/java/org/openrewrite/java/migrate/nio/file/RedundantUtf8Charset.java
timtebeek: Good catch — `getParameterNames()` always matches `getParameterTypes()` in length, so the guard was redundant. Removed it.

src/test/java/org/openrewrite/java/migrate/nio/file/RedundantUtf8CharsetTest.java
timtebeek: Fixed, bumped to 2026.

steve-aom-elliott (approved):

PR #1170, opened 2026-07-16
Move Java EE 8 servlet update into separate recipe

rlsanders4:
## What's changed?

This PR moves the Java EE 8 Servlet update to 3.x into a separate recipe. This new recipe is then referenced by the `org.openrewrite.java.migrate.javaee8` recipe list. Also, the order of the recipes in this list was adjusted to ensure that  `org.openrewrite.java.migrate.javaee7` recipe is called first.

## What's your motivation?

Splitting this recipe out allows for users to call it individually, without invoking the full Java EE 8 recipe list. Moving the Java EE 7 recipes to the top of the Java EE 8 list ensures that updates are made in the correct order for users migrating across multiple Java EE levels.

@timtebeek 

### Checklist


timtebeek (approved):
Thanks once again! No idea why these were ever in a different order; this indeed makes more sense.

PR #1179, opened 2026-07-30
Add mapping for CheckForNull to JSpecify annotation

zbynek:
## What's changed?
When migrating to JSpecify. `javax.annotation.CheckForNull` is now replaced by `org.jspecify.annotations.Nullable`.
Semantically `CheckForNull` is closer to `org.jspecify.annotations.Nullable` than `javax.annotation.Nullable` is (see [StackOverflow](https://stackoverflow.com/a/12301937/32987888)), so this is a safe change.

## What's your motivation?
* The recipe should remove all null-related annotations that belong to `javax.annotation` to avoid manual steps.
* Currently the recipe changes `@CheckForNull Object[]` to `Object @CheckForNull[]`, but then fails to replace it, resulting in code that does not compile (the old annotation does not have `TYPE_USE` target).

🤷 

🤷

Make a custom .yml recipe.

## Any additional context
Projects using SpotBugs for null analysis likely prefer `@CheckForNull` over `javax.annotation.Nullable`, so they might benefit from this change.

### Checklist

### DIFF
--- a/src/main/resources/META-INF/rewrite/jspecify.yml
+++ b/src/main/resources/META-INF/rewrite/jspecify.yml
@@ -72,6 +72,10 @@ recipeList:
       newFullyQualifiedTypeName: org.jspecify.annotations.Nullable
       ignoreDefinition: true
   - org.openrewrite.java.ChangeType:
+      oldFullyQualifiedTypeName: javax.annotation.CheckForNull
+      newFullyQualifiedTypeName: org.jspecify.annotations.Nullable
+      ignoreDefinition: true
+  - org.openrewrite.java.ChangeType:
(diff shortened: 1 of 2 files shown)


zbynek:
```
./gradlew test --tests "*JSpecifyBestPracticesTest"
```
succeeds but
```
./gradlew test
```
fails locally -- most of the failed tests seem to be related to https://github.com/openrewrite/rewrite/pull/8355  -- @timtebeek looped.

timtebeek (approved):
Thanks a lot, also for the helpful link explaining the nuances. I've rebased and will merge when CI succeeds

PR #1190, opened 2026-08-10
Add jakarta.validation-api dependency when migrating com.sun.istack.NotNull

steve-aom-elliott:
The `JavaxValidationMigrationToJakartaValidation` aggregator rewrites `com.sun.istack.NotNull` to `jakarta.validation.constraints.NotNull` via `ChangeType`, but the accompanying `AddDependency` for `jakarta.validation:jakarta.validation-api` was guarded by `onlyIfUsing: javax.validation..*`. Sources that used only `com.sun.istack.NotNull` had their imports swapped with no corresponding dependency added, producing an uncompilable result.

This adds a second `AddDependency` entry guarded by `onlyIfUsing: com.sun.istack.NotNull` so the dependency is added in that case as well. Test coverage extended to assert the POM change.


steve-aom-elliott:
For context, I think
- ~https://github.com/openrewrite/rewrite-migrate-java/pull/973 was potentially an incorrect move, and I'm looking into how this should have been dealt with instead, as `com.sun.istack.NotNull` is not a validation constraint, but rather more of an indicator. Need to dig into this further.~

timtebeek (approved):

PR #1213, opened 2026-08-20
Upgrade `ARG` default values used in Docker `FROM` instructions

timtebeek:
- Follow up to #1212.

## What's changed

- #1212 made `UpgradeDockerImageVersion` skip any `FROM` whose image name or tag contains a `$`, since the value can not be determined statically. That is the right call for a bare `ARG IMAGE_TAG`, but a great many Dockerfiles pin the Java version in a global `ARG` with a default value:

```dockerfile
ARG java_version=17
FROM eclipse-temurin:${java_version}-jre
```

Here the value *is* statically known, and the version can be bumped by rewriting the `ARG` default rather than the `FROM`.

The visitor now collects the literal defaults of the global `ARG`s (those before the first `FROM`, which are the only ones a `FROM` can reference) and resolves variable image references against them. When the resolved reference is an upgradable Java image, it writes the upgrade back to whichever of the two carries the value:

| Before | After |
| --- | --- |
| `ARG java_version=17`<br>`FROM eclipse-temurin:${java_version}` | `ARG java_version=25`<br>`FROM eclipse-temurin:${java_version}` |
| `ARG JAVA_VERSION=11`<br>`FROM eclipse-temurin:${JAVA_VERSION}-jre` | `ARG JAVA_VERSION=25`<br>`FROM eclipse-temurin:${JAVA_VERSION}-jre` |
| `ARG IMAGE_TAG=11-jre-alpine`<br>`FROM eclipse-temurin:${IMAGE_TAG}` | `ARG IMAGE_TAG=25-jre-alpine`<br>`FROM eclipse-temurin:${IMAGE_TAG}` |
| `ARG BASE_IMAGE=openjdk:11-jre`<br>`FROM ${BASE_IMAGE}` | `ARG BASE_IMAGE=eclipse-temurin:25-jre`<br>`FROM ${BASE_IMAGE}` |
| `ARG BASE_IMAGE=openjdk`<br>`FROM ${BASE_IMAGE}:11-jre` | `ARG BASE_IMAGE=eclipse-temurin`<br>`FROM ${BASE_IMAGE}:25-jre` |

Note the last two: an `openjdk` base has no tag beyond 17, so the deprecated image name has to move to `eclipse-temurin` along with the tag, whether that name sits in the `FROM` or in the `ARG`.

Everything that can not be resolved is still left untouched: an `ARG` without a default, an `ARG` whose default is itself built from another variable, a variable that only contributes part of the image name (`FROM ${REGISTRY}/eclipse-temurin:11-jre`), a variable that does not supply the leading version (`FROM eclipse-temurin:11${SUFFIX}`), and any `ARG` declared after the first `FROM`. `ARG`s that are not used in a `FROM` at all are never rewritten. As before, a digest pin is dropped when the tag is upgraded.

The fully literal `FROM` path is unchanged.

## Tests

Added parameterized cases for each row of the table above, plus multi-stage `ARG` reuse, digest pin removal through an `ARG`, and a set of negative cases covering the unresolvable and unrelated shapes. The existing tests are unchanged and still pass.


src/main/java/org/openrewrite/java/migrate/UpgradeDockerImageVersion.java
```diff
@@ -50,48 +61,284 @@ public class UpgradeDockerImageVersion extends Recipe {
...
+    }
+
+    /**
+     * The parser keeps any quotes around an {@code ARG} default value as part of the literal text, so they have to be
+     * taken off before matching a version, and put back on when writing the upgraded value.
+     */
+    @Value
+    private static class QuotedText {
+        String quote;
+        String text;
```
timtebeek: Not yet happy with this handling, as I believe it's inconsistent with other usages. Investigating upstream first.

timtebeek:
Currently exploring a registry method in the DockerImageReference trait, to make it easier to match & change.
- https://github.com/openrewrite/rewrite/pull/8599

PR #1226, opened 2026-08-25
Drop the `rewrite-analysis` dependency from `URLConstructorToURICreate`

timtebeek:
`URLConstructorToURICreate` was the only consumer of `org.openrewrite.meta:rewrite-analysis`, using `ConstantFold` to resolve a `String` identifier back to the literal it was initialized with. That lookup is now a walk over the enclosing blocks for the matching variable declaration, which drops the `RecipeRunException` catch that existed solely because `ConstantFold` throws on a lambda parameter.

To keep the previous behaviour, the walk matches on `JavaType.Variable` rather than on the name — the owner distinguishes a local shadowing a field from the field itself — and refuses to fold a variable that is assigned to anywhere in the file, including through `this`, which is the condition `ConstantFold` expressed as a single assigned value.

Tests cover the local-variable case that was previously untested, plus the reassigned, shadowed-field, and reassigned-through-`this` cases; all 14 pass.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -72,7 +72,6 @@ dependencies {
     implementation("org.openrewrite.recipe:rewrite-joda:$rewriteVersion")
     implementation("org.openrewrite.recipe:rewrite-jenkins:$rewriteVersion")
     implementation("org.openrewrite:rewrite-templating:$rewriteVersion")
-    implementation("org.openrewrite.meta:rewrite-analysis:$rewriteVersion")
 
     runtimeOnly("org.openrewrite:rewrite-java-8")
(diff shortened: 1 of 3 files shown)


MBoegers (approved):
This will not work for external constants Assume:
```java
// src/main/Constants.java
public class Constants {
   public static final String requestURL = "..";
   //...
}
```
where `Constants.requestURL` is used in different locations.
Not common, as it breaks configurability and multiple coding best practices, but still possible to encounter in small tools.

Okay to remove the overhead but want to flag the new gap.

timtebeek:
Indeed a known gap, much as we saw on the rewrite-spring module as well:
- https://github.com/openrewrite/rewrite-spring/pull/1139

PR #1229, opened 2026-08-31
Leave a Groovy `def` declaration alone instead of splicing `var` into its variable name.

motlin:
## What's changed?

`DeclarationCheck.isSingleVariableDefinition` now refuses declarations that expose no type expression for `var` to replace. The single guard covers `UseVarForConstructors`, `UseVarForObject`, `UseVarForGenericsConstructors`, and `UseVarForGenericMethodInvocations`.

## What's your motivation?

The `var` recipes corrupt Groovy `def` declarations, non-idempotently.

```groovy
// before
def newFile = new File(file.parentFile, newName)
assert file.renameTo(newFile) : "Failed to rename ${file} to ${newFile}"

// after one run
def varnewFile = new File(file.parentFile, newName)
assert file.renameTo(newFile) : "Failed to rename ${file} to ${newFile}"

// after a second run on the same file
def varvarnewFile = new File(file.parentFile, newName)
```

Originally hit on `rewrite-migrate-java` 3.42.1 with `rewrite-maven-plugin` 6.46.1, against a real
Maven archetype post-generate script. Still reproduces on `main` at `v3.43.0`.

Applying an `org.openrewrite.FindSourceFiles` precondition of `**/*.java` to the `var`
recipes.

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/lang/var/DeclarationCheck.java
+++ b/src/main/java/org/openrewrite/java/migrate/lang/var/DeclarationCheck.java
@@ -74,7 +74,9 @@ final class DeclarationCheck {
      * @return true if single variable definition with initialization and without var
      */
     private static boolean isSingleVariableDefinition(J.VariableDeclarations vd) {
-        TypeTree typeExpression = vd.getTypeExpression();
+        if (!replaceableTypeExpression(vd.getTypeExpression())) {
+            return false;
+        }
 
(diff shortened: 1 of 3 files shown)


motlin:
- See also https://github.com/liftwizard/liftwizard/pull/4125

timtebeek (approved):
Thanks for providing the fix here as well! Looks good to me based on a reading of the code. Any final thoughts @greg-at-moderne ?

PR #1230, opened 2026-09-01
Improve test coverage for LombokValToFinalVar

sullis:
Adds a test covering `lombok.val` inside a nested lambda, where the lambda parameter shadows a field of the same name.

The recipe correctly rewrites `val key = ...` to `final var key = ...` and removes the now-unused `lombok.val` import, leaving the shadowing lambda parameter untouched.

# Recipe

https://docs.openrewrite.org/recipes/java/migrate/lombok/lombokvaltofinalvar


sullis:
CI is green. Ready for review / feedback  @greg-at-moderne

greg-at-moderne (approved):
Thanks for coming up with this one.

PR #1232, opened 2026-09-02
Do not add `final` to a try-with-resources resource

sullis:
`LombokValToFinalVar` rewrote `val` to `final var` unconditionally. Resources declared in a try-with-resources statement are already implicitly final, so the added `final` modifier is redundant there.

This checks whether the variable declaration's parent is a `J.Try.Resource` and, if so, emits plain `var` instead of `final var`.

```java
// before
try (val writer = new StringWriter()) {
    writer.append(aaa);
}

// after
try (var writer = new StringWriter()) {
    writer.append(aaa);
}
```

Adds a test covering the try-with-resources case.


sullis:
Ready for review @greg-at-moderne

sullis:
CI build is green.

greg-at-moderne (approved):
Another great contribution. Thank you!

PR #1233, opened 2026-09-02
Preserve casts when converting `lombok.val` to `final var`

sullis:
## What's changed?

`LombokValToFinalVar` built the `final var` declaration with a `JavaTemplate`, which re-prints the initializer expression. Re-printing can lose type information: a cast to a type variable such as `(T) o` came out mangled, because the template parse has no knowledge of the enclosing class's type parameters.

This replaces the template with a direct LST transformation:

- Build the `var` `J.Identifier` (as the existing no-initializer branch already did) and swap it in for the `val` type expression, so both code paths now share that construction.
- Prepend a `final` modifier only when one isn't already present, moving the original type expression's prefix onto the modifier so formatting is preserved.

The initializer is carried over untouched, so casts — including generic ones — survive the conversion. Behavior for declarations without an initializer is unchanged (still a plain `var`, per openrewrite/rewrite#5637).

Also included is a small unrelated test fix: `UpgradeToJava25Test.upgradeToJava25` pinned `maven-compiler-plugin` to `3.15.x` in an assertion pattern, which breaks as soon as a newer minor is released. The pattern now accepts any `3.15+` version.

The prefix/modifier juggling is the part worth a close look — the existing tests cover annotated, already-`final`, multi-variable, and comment-carrying declarations, and two new tests cover the casts: `preserveCast` (`(String) "foo"`) and `preserveGenericCast` (`(T) o`).

I can't run the build locally — dependency resolution needs credentials for `artifacts.codegenomeproject.org` — so I'm relying on CI for verification.

Keeping `JavaTemplate` and making it type-aware. Sidestepping the re-print entirely is simpler, and it also removes the divergence between the initializer and no-initializer code paths.

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/lombok/LombokValToFinalVar.java
+++ b/src/main/java/org/openrewrite/java/migrate/lombok/LombokValToFinalVar.java
@@ -19,17 +19,19 @@ import lombok.Getter;
 import org.openrewrite.*;
 import org.openrewrite.internal.ListUtils;
 import org.openrewrite.java.JavaIsoVisitor;
-import org.openrewrite.java.JavaTemplate;
 import org.openrewrite.java.search.MaybeUsesImport;
 import org.openrewrite.java.search.UsesType;
 import org.openrewrite.java.service.AnnotationService;
 import org.openrewrite.java.tree.J;
+import org.openrewrite.java.tree.Space;
 import org.openrewrite.java.tree.TypeTree;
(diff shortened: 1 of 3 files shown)


sullis:
- Ready for review @greg-at-moderne

sullis:
- CI build is green

greg-at-moderne (approved):

greg-at-moderne:
Thanks for the PR. Looks well designed and I appreciate the tests.

PR #1236, opened 2026-09-09
Add lombok.config recipes

sullis:
Adds two `lombok.config` recipes, built on a shared `LombokConfig` helper that parses a config file following the line grammar of Lombok's own `ConfigurationParser`. Both act on the `lombok.config` at the root of the repository, and read any `*.config` file, since an `import` names a file by path rather than by name.

Neither recipe is part of `LombokBestPractices`; both are registered in `recipes.csv` and are run on their own.

## `ConsolidateLombokConfig`

Merges the directives of every nested `lombok.config` into the root `lombok.config` and deletes the nested files.

- Directives are appended, so whatever the root already contains is left exactly as written. Directives it already declares, itself or through an `import`, are dropped rather than repeated; whitespace and key case are normalized for comparison only.
- Nested files are merged shallowest first, the order Lombok reads them in.
- A nested file stays in place when hoisting it would change what Lombok does: it declares `config.stopBubbling` or sits under a directory that does, it has an `import`, `clear` or `-=`, a `lombok.config` between it and the root speaks about one of the same directives, another `lombok.config` imports it, no Java source sits at or below it, or it holds nothing but comments and unparseable lines.
- No changes are made at all when there is no root `lombok.config`, when two files assign conflicting values to the same key, when the root `clear`s or `-=`s a key a nested file speaks about, or when the root's imports cannot be resolved.

Note that hoisting widens a directive's scope from one directory to the whole project, which the recipe description calls out.

## `AddStopBubblingToLombokConfig`

Appends `config.stopBubbling = true` to the root `lombok.config` unless it already declares that key. Without it, Lombok keeps walking up past the project and a `lombok.config` in a parent directory of the checkout takes part in the build.

- Nothing is added when the key is already declared, whatever value it is assigned and whether the root declares or imports it, or when the root's imports cannot be resolved.
- Only the root file is changed.

## Tests

79 tests covering merging and merge order, deduplication, conflict detection, each non-hoistable case, `stopBubbling` ancestors, import resolution and cycles, comment handling, case-insensitive keys, Java-source layout, and CRLF and missing-trailing-newline root files.


sullis:
Ready for review @greg-at-moderne  @timtebeek

greg-at-moderne (commented):
The change itself looks line to me. I appreciate the defensive approach.

The only thing I'd kindly request is getting rid of the verbosity and excessive comments/texts.

sullis:
> The change itself looks line to me. I appreciate the defensive approach.
> 
> The only thing I'd kindly request is getting rid of the verbosity and excessive comments/texts.

I pushed an update.

greg-at-moderne (approved):

PR #1237, opened 2026-09-10
Fix annotation handling in `LombokValToFinalVar`

sullis:
When building the replacement `var` identifier, the recipe used `AnnotationService.getAllAnnotations(getCursor())`. At that point the cursor is on the `J.VariableDeclarations`, so this returns the annotations of the whole declaration (e.g. a `@SuppressWarnings("unchecked")` on the statement), not the annotations attached to the type expression. Those declaration-level annotations were then attached to the new `var` identifier as well, leaving them printed twice.

This takes the annotations from the type expression itself instead, which is what the new type identifier should carry over. Declaration-level annotations stay where they already are on the `J.VariableDeclarations`.

Added a test covering `@SuppressWarnings("unchecked")` on a `val` declaration.

### DIFF
--- a/src/main/java/org/openrewrite/java/migrate/lombok/LombokValToFinalVar.java
+++ b/src/main/java/org/openrewrite/java/migrate/lombok/LombokValToFinalVar.java
@@ -21,7 +21,6 @@ import org.openrewrite.internal.ListUtils;
 import org.openrewrite.java.JavaIsoVisitor;
 import org.openrewrite.java.search.MaybeUsesImport;
 import org.openrewrite.java.search.UsesType;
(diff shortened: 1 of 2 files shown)


sullis:
cc: @greg-at-moderne  @timtebeek

greg-at-moderne (approved):

greg-at-moderne:
The CI failure seems unrelated. Yet I fail to understand why it would be happening in this branch, but not happening in `main`.

greg-at-moderne:
OK. I think I get it. The branch is in the fork (understandably) and we don't want to share Code Genome secrets.

Issue #467, opened 2024-04-29
`URLThreeArgumentConstructor` does not introduce newly thrown exception


timtebeek:
## What version of OpenRewrite are you using?
- OpenRewrite v8.24.0
- rewrite-migrate-java v2.12.0

## How are you running OpenRewrite?
I'm running `org.openrewrite.java.migrate.net.URLConstructorsToURI.URLThreeArgumentConstructorRecipe`, via:
https://github.com/openrewrite/rewrite-migrate-java/blob/f399d72046b68fd0820a6b326badfd5ff9482105/src/main/java/org/openrewrite/java/migrate/net/URLConstructorsToURI.java#L44-L58

## What is the smallest, simplest way to reproduce the problem?
```java
class A {
    URL urlConstructor(String protocol, String host, String file) throws IOException {
        return new URL(protocol, host, file);
    }
}
```

## What did you expect to see?
```java
class A {
    URL urlConstructor(String protocol, String host, String file) throws URISyntaxException, IOException {
            return new URI(protocol, null, host, -1, file, null, null).toURL();
        }
}
```

## What did you see instead?
```java
class A {
    URL urlConstructor(String protocol, String host, String file) throws IOException {
            return new URI(protocol, null, host, -1, file, null, null).toURL();
        }
}
```
Note the lacking `throws URISyntaxException`.

## What is the full stack trace of any errors you encountered?
Compiler error, as `URISyntaxException` does not extend `IOException`, in contrast to `MalformedURLException` thrown before. As reported by @IanDarwin

timtebeek:
This recipe is implemented as a [Refaster recipe](https://docs.openrewrite.org/authoring-recipes/refaster-recipes), and as such it's well possible the fix should land in https://github.com/openrewrite/rewrite-templating to introduce the new exception there from what's thrown in the after template method. /cc @knutwannheden

Riyazul555:
Hey @timtebeek I have a suggestion on this issue 
It seems like in your URLThreeArgumentConstructor class, you are converting new URL(String, String, String) constructors to new URI(...).toURL() without adding the throws URISyntaxException declaration in the AfterTemplate method. To fix this issue, you should include throws URISyntaxException in your AfterTemplate method like this:

```
public static class URLThreeArgumentConstructor {
        @BeforeTemplate
        URL urlConstructor(String protocol, String host, String file) throws Exception {
            return new URL(protocol, host, file);
        }

        @AfterTemplate
[...]

PR #1241, opened 2026-09-16
Convert 5-arg URL constructor with URLStreamHandler to URL.of

### LINKED ISSUE #849: Forbidden method invocation: `java.net.URL#<init>(java.lang.String,java.lang.String,int,java.lang.String,java.net.URLStreamHandler)` [Deprecated in Java 21]
- https://github.com/checkstyle/checkstyle/pull/17789

I am using

- OpenRewrite v1.2.3
- Maven/Gradle plugin v1.2.3
- rewrite-_module_ v1.2.3

I am using the Maven plugin, and my project is a single module project.
```xml
<rewrite-migrate-java.version>3.17.0</rewrite-migrate-java.version>
```

```java

        final URL url = new URL("test", null, 0, "", new URLStreamHandler() {
            @Override
            protected URLConnection openConnection(URL u) {
                return urlConnection;
            }
        });
```

```java

        final URL url = new URL("test", null, 0, "", new URLStreamHandler() {
            @Override
            protected URLConnection openConnection(URL u) {
                return urlConnection;
            }
        });
```

## What did you see instead?

```java

        final URL url = URL.of(new URI("test:///"), new URLStreamHandler() {
            @Override
            protected URLConnection openConnection(URL u) {
                return urlConnection;
            }
        });
```

```
stacktrace output here
```

harrisleesh:
## What's changed?

`URLConstructorsToNewURI` now also matches the deprecated 5-arg constructor `new URL(String protocol, String host, int port, String file, URLStreamHandler handler)` and converts it to `URL.of(new URI(protocol, null, host, port, file, null, null), handler)`, preserving the custom stream handler. The URI construction follows the same shape already used for the 3- and 4-arg variants.

## What's your motivation?

- Fixes https://github.com/openrewrite/rewrite-migrate-java/issues/849 — this variant was left out and @timtebeek welcomed a contribution to fill the gap.

- `URL.of(URI, URLStreamHandler)` requires Java 20+, while the 3/4-arg replacements only need `URI.toURL()`. I kept everything in the same recipe since it is listed under the Java 21 migration, but happy to gate the 5-arg case behind a version precondition if you prefer.
- Same caveat as #467 applies: the URI constructor throws `URISyntaxException` where the URL constructor threw `MalformedURLException`.

## Any additional context

Heads-up on validation: I could not resolve `org.openrewrite` artifacts from the Code Genome repository locally yet, so I verified the change in a standalone project against rewrite 8.90.4 — the new test fails without the recipe change and both tests pass with it. Relying on CI for the exact HEAD build; if anything fails I'll follow up promptly.

### Checklist
- [ ] I've run `./gradlew build` locally, and committed any resulting changes to `recipes.csv` (blocked on Code Genome credentials — no new recipe class was added, so `recipes.csv` should be unaffected)
- [x] I've [formatted the lines I changed](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#code-style-and-formatting), without reformatting code I didn't touch


greg-at-moderne (approved):
Makes sense.

Thanks for the contribution!

greg-at-moderne:
> URL.of(URI, URLStreamHandler) requires Java 20+, while the 3/4-arg replacements only need URI.toURL(). I kept everything in the same recipe since it is listed under the Java 21 migration, but happy to gate the 5-arg case behind a version precondition if you prefer.

Thanks for calling this out. I think it's safe. Do you agree @timtebeek?

timtebeek:
Thanks yes! Seems fine for now, as we've held off on adding the recipe by default until we get the exception handling in place:
https://github.com/openrewrite/rewrite-migrate-java/blob/4a9098111fae4b7c9fe0c3f2ea22a30e818697e1/src/main/resources/META-INF/rewrite/java-version-21.yml#L35-L36

